PR #12, opened 2019-12-27
Add openSUSE instructions

jubalh:


sharkwouter:
Oh wow, thanks a lot! I'll see if I can test this in a VM tomorrow

jubalh:
I will also push it to official repositories. But it will take some time until all the reviews and tests are through. Especially during the holidays.
Once it's in there I will update the README to reflect this.

sharkwouter:
I've tested it now. It seems Leap 15.1 doesn't have the required version of the python3-gobject package. Minigalaxy requires version 3.30. I also tried installing in Tumbleweed, but it doesn't find the package.

jubalh:
Which version of python3-gobject is required? I should add this to the spec file.

> I also tried installing in Tumbleweed, but it doesn't find the package.

Did you run the commands from this PR to add the repo?

sharkwouter:
Yes I did, it told me there was no minigalaxy package on Tumbleweed. On Leap it installed fine, but it doesn't start because python3-gobject version 3.30 or newer is required for it to start.

jubalh:
> Yes I did, it told me there was no minigalaxy package on Tumbleweed. 

The commands in this PR actually add the repo. the `ref` then refreshes the caches, and then you can search/install it.

I can't believe that it doesn't work for you since it works here ;)

Can you paste the output of the commands. Or if you already ran them you can use `zypper lr -v` to show me your repo list.

sharkwouter:
Ah, I'll test again. Maybe I forgot that part

jubalh:
That's what I meant in the previous comment with `Did you run the commands from this PR to add the repo?` ;)

sharkwouter:
I've tested it again. This time on the KDE version of Tumbleweed. The installation worked, but I needed to install some additional package before I was able to launch Minigalaxy.

These are the packages which seem to be required for Minigalaxy to work on OpenSUSE:
- typelib-1_0-Gtk-3_0
- typelib-1_0-WebKit2-4_0
- python3-gobject-Gdk

I think we should also add sudo in front of the zypper commands in the README. I can do this or you could update the pull request. Let me know what you'd prefer.

One more thing, yesterday I released a new version which fixes most crashes and freezes. You can find it here: https://github.com/sharkwouter/minigalaxy/releases/tag/0.9.1

jubalh:
> I've tested it again. This time on the KDE version of Tumbleweed. The installation worked, but I needed to install some additional package before I was able to launch Minigalaxy.

Thanks for testing!

> These are the packages which seem to be required for Minigalaxy to work on OpenSUSE:
    typelib-1_0-Gtk-3_0
    typelib-1_0-WebKit2-4_0
    python3-gobject-Gdk

Ok, good! I added them now :-)

> One more thing, yesterday I released a new version which fixes most crashes and freezes. You can find it here: https://github.com/sharkwouter/minigalaxy/releases/tag/0.9.1

I added your repo to one of my scripts which notifies my every time a new release is happening. So thanks for hilighting it here, I'll also get notified from my script in the future upon each relesase :-)

BTW minigalaxy is already on it's way to the official repositories: https://build.opensuse.org/request/show/759773

jubalh:
> I think we should also add sudo in front of the zypper commands in the README.

I don't think so. Usually Linux users know that commands that change the systems state (like installing new software) need to be run as root.

So it could do that some users run a `su -` before calling zypper because they do general systems maintenance. Others might not and would add a `sudo` in any case I think they know about it and only want to read the actual commands for adding the repository and installing the program.

But of course if you like it better you can add a sudo.

sharkwouter:
> I added your repo to one of my scripts which notifies my every time a new release is happening. So thanks for hilighting it here, I'll also get notified from my script in the future upon each relesase :-)

Oh nice, that's really useful.

> BTW minigalaxy is already on it's way to the official repositories: https://build.opensuse.org/request/show/759773

Awesome! I should start trying to get it into Debian.

> I don't think so. Usually Linux users know that commands that change the systems state (like installing new software) need to be run as root.

You're probably right OpenSUSE has a different default sudo setup.

jubalh:
I rebased the commt on current master. So that you can just merge.

sharkwouter:
Thanks! Merged!

sharkwouter:
You might want to add libgtk-3-0 as a dependency as well actually. Not sure if openSUSE ships any variants without it.

jubalh:
I think that's pulled in already.

PR #29, opened 2020-01-01
This commit allow to open a dialog box and choose quickly your install folder

Odelpasso:
Hello,

I did this pull request to add a new button in "preference".
Instead of to insert manually the path for your installation folder, the new button opens a new dialog box which allow to choose quickly the folder you want.

Screenshot

I have the basic with python (and i love it) and very new with PyOBject :D

### DIFF
--- a/data/ui/preferences.ui
+++ b/data/ui/preferences.ui
@@ -18,18 +18,7 @@
             <property name="can_focus">False</property>
             <property name="layout_style">end</property>
             <child>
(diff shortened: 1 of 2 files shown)


sharkwouter:
Thanks for the effort! I've tested it and it does work. I'm a bit reluctant to merge it, though. The placement of the button would make it harder to add more preferences to the menu later.

Would it be possible to change the entry field into a file chooser button instead of adding an additional button? That would solve the problem.

PR #32, opened 2020-01-02
Use FileChooserButton instead of GetEntry to set your installation folder

Odelpasso:
It's still better to choose quickly your installation folder instead of to write it manually :P


minigalaxy/window/preferences.py
```diff
@@ -69,8 +69,13 @@ def __save_language_choice(self) -> None:
             lang, _ = model[lang_choice][:2]
             self.__config.set("lang", lang)
 
+    @Gtk.Template.Callback("on_button_chooser_file_set")
```
sharkwouter: This function isn't needed

PR #34, opened 2020-01-03
Keep install file and avoid to download it again even if the file exists

Odelpasso:
Hello,

What exactly does the Pull Request do :
**1- Revert the line of codes which delete the "_Name_of_my_Game.sh" in download folder.**

Why i did that : I pray for people who have a low bandwidth.
Wasteland 2 has a download size of 15Gb. If these people change/lost their PC, either they saved the install game's folder/install game's file or they lost everything. 

In my case, i have all my _Name_of_my_Game.sh in a External Hard Drive.

**2- Avoid to download again the Install file if it already exists.**
After the 1-, we keep the _Name_of_my_Game.sh_ for each game. 
Unfortunately, if the user needs to re-install the game he wants and click on "Download", the file is downloaded and replace the existing file --> Poor low bandwidth users.

Solution : 
MiniGalaxy test if the file exist and if it's the case, install only the game without to download again the .sh

**3- Change the button label on the Main window**
Initially after to change 1- and 2-, the button in the main window show "Download"

Solution :  
MiniGalaxy test if the file exist and if it's the case, the button label "Download" is replaced by "Install"

### DIFF
--- a/data/ui/preferences.ui
+++ b/data/ui/preferences.ui
@@ -56,6 +56,7 @@
           <object class="GtkGrid">
             <property name="visible">True</property>
             <property name="can_focus">False</property>
+            <property name="row_homogeneous">True</property>
             <child>
               <object class="GtkLabel">
                 <property name="visible">True</property>
@@ -81,7 +82,7 @@
               <object class="GtkLabel">
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
-                <property name="label" translatable="yes" context="install_path" comments="The place directory in which games are installed. Ends with &quot;: &quot;.">Installation path: </property>
+                <property name="label" translatable="yes" context="install_path" comments="The place directory in which games are installed. Ends with &quot;: &quot;.">Install path: </property>
               </object>
               <packing>
                 <property name="left_attach">0</property>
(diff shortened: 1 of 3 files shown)


minigalaxy/window/gametile.py
```diff
@@ -135,7 +143,12 @@ def __install_game(self) -> None:
         os.makedirs(temp_dir)
 
         # Extract the installer
-        with zipfile.ZipFile(self.download_path) as archive:
+        if os.path.exists(self.download_path):
+            install_path = self.download_path
```
sharkwouter: Can you rename this to installer_path?

data/ui/preferences.ui
```diff
@@ -100,6 +101,28 @@
                 <property name="top_attach">1</property>
               </packing>
             </child>
+            <child>
+              <object class="GtkLabel">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="tooltip_text" translatable="yes">Enable this option if you want to keep the installer after the installation</property>
+                <property name="label" translatable="yes">Installer Files</property>
```
sharkwouter: Could you change the text of the label to ``Keep installers: ``?

PR #36, opened 2020-01-04
Delete unnecessary code

Odelpasso:
I think we can delete these lines of code since the install path is set by FileChooserButton and not by GetEntry text bar.

### DIFF
--- a/minigalaxy/window/preferences.py
+++ b/minigalaxy/window/preferences.py
@@ -108,21 +108,10 @@ class Preferences(Gtk.Dialog):
     def save_pressed(self, button):
         self.__save_language_choice()
         self.__set_keep_installer()
-        if self.__save_install_dir_choice():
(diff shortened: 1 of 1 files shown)


sharkwouter:
Perhaps this line can go since you can't select a directory which doesn't exist: https://github.com/sharkwouter/minigalaxy/blob/41f3d256c02fa5ade3f38224bf575979dfae113d/minigalaxy/window/preferences.py#L82

But the code removed in this PR is important. In this part it checks if it can write to the directory by trying to write a file to it and returning False if there is an error: https://github.com/sharkwouter/minigalaxy/blob/41f3d256c02fa5ade3f38224bf575979dfae113d/minigalaxy/window/preferences.py#L88

Try setting the directory to ``/var`` for instance, that's not a user writable directory.

Issue #22, opened 2019-12-31
Preferences - Stay Logged In


SwampRabbit:
I would like to suggest an addition to Preferences, with a selection option to Stay Logged In.

Currently the default is to keep users logged in unless they select Logout from the menu.

This would allow users choose if they want have the client logout after closing or to stay logged in.

Thank you for your consideration.

sharkwouter:
That's a pretty good idea actually. It won't be high priority for now, though, but I'll keep it open and work on it later

SwampRabbit:
Thank you, no rush at all.  Keep up the great work, it is an amazing application.

sharkwouter:
This has been implemented in #37, thanks @Odelpasso! The feature will be available in the next release.

SwampRabbit:
@sharkwouter and @Odelpasso thank you for implementing this so quickly.  Keep up the great work, it is really appreciated!

PR #37, opened 2020-01-04
Implement "stay logged" possibility

Odelpasso:
Hello,

I implemented the possibility for an user to quit MiniGalaxy (with Quit button) stay logged or disconnected. For that, the user must to check or uncheck the check button in Preferences. 

In my case, it works at 90% because i'm quite frustrated. If the user close the windows with the X button at the top right, the process still exist + stay logged if he launch MiniGalaxy again.

If the code is correct, i think for the 1st time i fix an issue : 
https://github.com/sharkwouter/minigalaxy/issues/22

### DIFF
--- a/data/ui/preferences.ui
+++ b/data/ui/preferences.ui
@@ -135,6 +135,29 @@
                 <property name="top_attach">2</property>
               </packing>
             </child>
+            <child>
+              <object class="GtkLabel">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="halign">start</property>
+                <property name="label" translatable="yes">Stay logged in:</property>
+              </object>
+              <packing>
+                <property name="left_attach">0</property>
+                <property name="top_attach">3</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkSwitch" id="button_stay_logged_in">
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="active">True</property>
+              </object>
+              <packing>
+                <property name="left_attach">1</property>
+                <property name="top_attach">3</property>
+              </packing>
+            </child>
           </object>
           <packing>
             <property name="expand">False</property>
--- a/minigalaxy/window/preferences.py
+++ b/minigalaxy/window/preferences.py
@@ -35,6 +35,7 @@ class Preferences(Gtk.Dialog):
     button_cancel = Gtk.Template.Child()
     button_file_chooser = Gtk.Template.Child()
     button_save = Gtk.Template.Child()
+    button_stay_logged_in = Gtk.Template.Child()
     combobox_language = Gtk.Template.Child()
     switch_install = Gtk.Template.Child()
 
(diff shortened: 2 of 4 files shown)


sharkwouter:
Thanks for the pulI request! It looks pretty good, but I do have some remarks and I might be able to help you out with the issue you have.

I think you'll want to remove this line: https://github.com/sharkwouter/minigalaxy/blob/14587c09ae51bcfc86dc47035cb7232426ce4af5/bin/minigalaxy#L61

Then make an exit function in ``window.py`` and use ``self.connect`` in the init to connect a destroy event to that fucntion. Although I'm not 100% sure if that would work.

You should probably also check if the setting is set on startup and show the login screen if not. You can probably just unset the refresh token and username in the init of ``windows.py`` if it is not set.

Could you rename the ``stay_log`` to ``stay_logged_in`` everywhere? That would make it clearer a bit clearer, sorry :/

Odelpasso:
I modified my commit to check if the user set the option to stay logged.
It was easiest for me to do that instead of the destroy event + `self.connect`

But if you have the possibility to code that and show me how it's implemented  (on Discord), it will be great for my knowledge.

minigalaxy/window/preferences.py
```diff
@@ -45,6 +46,13 @@ def __init__(self, parent, config):
         self.__set_language_list()
         self.button_file_chooser.set_filename(config.get("install_dir"))
         self.switch_install.set_active(self.__config.get("keep_installers"))
+        self.button_stay_log.set_active(self.__config.get("stay_log"))
```
sharkwouter: Here it still says stay_log

Odelpasso:
I think is good now.

Odelpasso:
Typo Error. It's good now

minigalaxy/window/window.py
```diff
@@ -44,6 +44,11 @@ def __init__(self, name):
         if not os.path.exists(THUMBNAIL_DIR):
             os.makedirs(THUMBNAIL_DIR)
 
+        # Check if the user unset to stay logged in Preference and refresh token
+        if not self.api.config.get("stay_logged_in"):
```
sharkwouter: This part causes a problem for old installations. An old configuration file will return ``None`` here, but it should be ``True``. Try this code:
```
stay_logged_in = self.api.config.get("stay_logged_in")
        if stay_logged_in is not None and not stay_logged_in:
            self.config.unset("username")
            self.config.unset("refresh_token")
        else:
            self.config.set("stay_logged_in", True)
```

Odelpasso:
Change done. 
I replaced the GtkCheckButton to a GtkSwitch

data/ui/preferences.ui
```diff
@@ -135,6 +135,29 @@
                 <property name="top_attach">2</property>
               </packing>
             </child>
+            <child>
+              <object class="GtkLabel">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="halign">start</property>
+                <property name="label" translatable="yes">Stay Logged</property>
```
sharkwouter: Can you change the text on the label to ``Stay logged in: ``

sharkwouter:
Looks good! Thanks a lot for the work!

PR #43, opened 2020-01-07
Add Support Button

BlindJerobine:
Added a support button in game-specific menu to open the Support Webpage.


sharkwouter:
Thanks for the pull request! This is not a bad idea, I'll see if I can take a look at it tonight.

sharkwouter:
I've tested and reviewed this now. It works well, but I do have few minor gripes with it:

1. Clicking the support button results in an exception if there is no internet connection. This should probably give you an error message.

2. The get_info and get_download_info use the same API endpoint. Could you make the get_info function just return the result of the request  and make the get_download_info function use that?

3. The get_info function takes a string instead of a game object.

BlindJerobine:
I'll edit the code :+1:

sharkwouter:
Could you change the error message to be a bit more descriptive? Maybe something like:

Couldn't open support page
Please check your internet connection

sharkwouter:
Thanks for the work!

PR #48, opened 2020-01-08
Add pt_BR translation

EsdrasTarsis:
Add Brazilian Portuguese translation, I had to do some fix typos, but I think now the translation is complete.


sharkwouter:
It looks good and works well, thanks for contributing!

PR #55, opened 2020-01-12
Implement option to enable/disable FPS counter for Nvidia and AMD GPU users

Odelpasso:
Hello,

The title explains this commit :D

### DIFF
--- a/data/ui/preferences.ui
+++ b/data/ui/preferences.ui
@@ -165,6 +165,31 @@
                 <property name="top_attach">3</property>
               </packing>
             </child>
+            <child>
+              <object class="GtkSwitch" id="switch_show_fps">
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="halign">start</property>
+                <property name="valign">center</property>
+              </object>
+              <packing>
+                <property name="left_attach">1</property>
+                <property name="top_attach">4</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkLabel">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="halign">start</property>
+                <property name="valign">center</property>
+                <property name="label" translatable="yes">Show FPS:</property>
+              </object>
+              <packing>
+                <property name="left_attach">0</property>
+                <property name="top_attach">4</property>
+              </packing>
+            </child>
           </object>
           <packing>
             <property name="expand">False</property>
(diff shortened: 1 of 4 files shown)


Odelpasso:
After reflection, i think that it's not necessary to check if nvidia or amdgpu is loaded to enable the FPS counter. 

FPS env variable are enabled at the same time if the option is enabled in Preference.

sharkwouter:
I'm getting the following error when trying to open the preferences window with this change:
```
  File "/home/wouter/minigalaxy/minigalaxy/window/window.py", line 89, in show_preferences
    preferences_window = Preferences(self, self.config)
  File "/home/wouter/minigalaxy/minigalaxy/window/preferences.py", line 52, in __init__
    self.switch_enable_fps.set_active(self.__config.get("enable_fps"))
```

Odelpasso:
Updated boss :D

sharkwouter:
I'm still seeing the same issue :(

sharkwouter:
I've added my commit with the configuration fix. It does work now.

sharkwouter:
I've also tested it on Intel now. With Intel GPUs it doesn't seem to cause any issues.

minigalaxy/window/gametile.py
```diff
@@ -298,6 +298,13 @@ def __start_game(self) -> subprocess:
     def __get_execute_command(self) -> list:
         files = os.listdir(self.__get_install_dir())
 
+        # Enable FPS Counter for Nvidia or AMD (Mesa) users
+        if self.api.config.get("enable_fps") is None:
+            self.apt.config.set("enable_fps", False)
```
sharkwouter: Could you rename it to show_fps?

data/ui/preferences.ui
```diff
@@ -165,6 +165,31 @@
...
+                <property name="top_attach">4</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkLabel">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="halign">start</property>
+                <property name="valign">center</property>
+                <property name="label" translatable="yes">Enable FPS:</property>
```
sharkwouter: Could you change it to Show FPS:

Odelpasso:
Boss, there is a small bug and i don't know how to fix that.
How to reproduce : 
- Delete config.json
- Launch Minigalaxy and launch a game --> FPS is not shown/disabled by default
- Go to Preference, enable "Show FPS" and launch a game --> FPS is shown
- Return to Preference, disable "Show FPS" and launch a game --> FPS is still shown and if you check in the config.json, "show fps" is set to False.

sharkwouter:
That bug can be solved by unsetting the environment variables when show fps is disabled in the preferences window.

Odelpasso:
I think is good now :D

PR #57, opened 2020-01-13
Update openSUSE instructions

jubalh:
minigalaxy landed now in the official repos of openSUSE Tumbleweed.
Future stable releases will thus contain it too.
However let's still mention the devel project (from where it goes into
official repos) and how to add it.


sharkwouter:
Oh wow, that's awesome! Thanks for the work!

PR #59, opened 2020-01-15
Added translation

thomasb22:
Hi, here's a first translation in French.


sharkwouter:
Thanks! It looks good!

PR #66, opened 2020-01-30
Partial implementation of wine

Odelpasso:
Hello,

Partial implementation of Windows games. Maybe the code can be code can be improved or simplified.

Currently, only Windows games with only one installer .exe file works (old games or indies games). 

@sharkwouter wants to update code to move downloading code to an other .py file. I hope it will not be difficult to update my code. :D 

About Windows games installation, i prefered to use Wine instead of innoextract. It was not complicated but there was twice the line of code. 
Currently, i have 4 games on GoG and these 4 games are extracted into 4 different folders ... I don't want to imagine the number of folder for all games available on GoG (~ 3000)

### DIFF
--- a/data/ui/gametile.ui
+++ b/data/ui/gametile.ui
@@ -10,6 +10,20 @@
         <property name="can_focus">False</property>
         <property name="orientation">vertical</property>
         <child>
+          <object class="GtkModelButton" id="menu_button_support">
+            <property name="visible">True</property>
+            <property name="can_focus">True</property>
+            <property name="receives_default">True</property>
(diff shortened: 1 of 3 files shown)


Odelpasso:
@sharkwouter What do you think about my patch v2 ?

sharkwouter:
I'm currently working on giving users the option to download windows games and currently it can download all parts of a Windows games. I haven't quite figured what to do with it afterwards, though, I'll look into that later. If you want to check it out, take a look at this branch. Let me know if you'd like to help out and what you'd need.

sharkwouter:
@Odelpasso and I have decided this pull request can be closed, since it's based on old code and won't be updated directly.

PR #68, opened 2020-02-10
Update README.md

ArturWroblewski:
Added a YouTube video link showing the minigalaxy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,9 @@ A simple GOG client for Linux that lets you download and play your GOG Linux gam
 
 ![screenshot](screenshot.jpg?raw=true)
 
+See how easy it is
+https://youtu.be/6ey3mO1QhZ4
+
 ## Features
 
 The most important features of Minigalaxy:


sharkwouter:
Sorry for waiting so long with replying to this one. I think it would be cool to include the video, but I don't like just having it as a link. Do you know if it is possible to show the youtube web player inside the README and on the website?

ArturWroblewski:
I have to prepare the animated gif manually. You can't convert from video so it will take some time.

sharkwouter:
That's a shame, I was not aware of that. Then maybe a short gif might be better.

Let me come back to you on this after the release. I think this is not required to be ready before then. Feel free to join our [Discord](https://discord.gg/RC4cXVD) if you'd like to have a chat about it.

sharkwouter:
I've thought about this for a while longer. In the end, I don't really think a video or gif adds much to the existing screenshot.

Sorry for taking so long for this conclusion. So much has happened since February. That's not a good excuse, though, I shouldn't have let you wait :(

PR #79, opened 2020-03-03
Updated french translation

thomasb22:
Updated french translation. But I have a bug, the preference window don't work, with this error message:

```
  File "/home/se7h/Logiciels/minigalaxy-fork/minigalaxy/ui/window.py", line 61, in show_preferences
    preferences_window = Preferences(self)
  File "/home/se7h/Logiciels/minigalaxy-fork/minigalaxy/ui/preferences.py", line 31, in __init__
    self.switch_show_fps.set_active(Config.get("show_fps"))
```


sharkwouter:
I'll look into it tonight. Thanks for reporting and updating so fast!

sharkwouter:
Actually, could you post the contents of ``~/.config/minigalaxy/config.json``?

sharkwouter:
Anyway, I've merged the updated translation, since that is unrelated. I'll do a test of all translations before release, just in case something doesn't work.

thomasb22:
You're right, I renamed the directory `~/.config/minigalaxy`, and I logged back in, and the preferences window works.
You'd have to edit the file if there are missing settings in this version.

sharkwouter:
What I think it happening is that the configuration file is being reloaded before the new fields have been added to it. So most likely, restarting Minigalaxy would have worked as well. I will need to test this, though.

PR #84, opened 2020-03-11
Adding Norwegian Nynorsk translation.

LordPilum:
The title says it all, really.
.po and .mo files created with poedit based on existing nb_NO.po file.

Please let me know if I need to update anything.


sharkwouter:
I've tested it and it works well. Thanks!

PR #87, opened 2020-03-11
Update pt_BR translation

EsdrasTarsis:
Sorry to send this translation too late, I hope this update will be released in the next release.

You can notify me by email (on my GitHub profile) to update the translation before a release.


sharkwouter:
That's a shame, I'll email you next time then. I think the next release will still take a while to come out, though, but I'll include it.

My only complaint here is the inclusion of the file ``data/po/pt_BR.mo``. If you can remove it, that would be great, since it isn't needed.

EsdrasTarsis:
Done, thanks for the feedback.

sharkwouter:
Thanks! I've merged it

PR #88, opened 2020-03-12
Fix icon size to conform guidelines and pass linter check

tim77:
Need for https://github.com/flathub/flathub/pull/1402#pullrequestreview-373598742


sharkwouter:
I've replaced the image, since it wasn't quite the same one. Would it be possible to offer both the 192x192 and 128x128 version without breaking the flatpak build?

tim77:
> Would it be possible to offer both the 192x192 and 128x128 version without breaking the flatpak build?

I guess so. If we look on [appstream-glib](https://github.com/hughsie/appstream-glib/tree/master/libappstream-glib) sources there is no 192px format here, so that's why AppStream check failed (probably).

sharkwouter:
I've included both now and renamed the images to follow the flatpak id naming.

Issue #82, opened 2020-03-10
Installation instructions look weird on the website


sharkwouter:
The Minigalaxy website uses github pages to build from the README. The code blocks which contains the command line instructions for installation and links look weird on it, though.

Website: https://sharkwouter.github.io/minigalaxy/
![Bildschirmfoto 2020-03-10 um 10 53 42](https://user-images.githubusercontent.com/5042659/76300683-6e793680-62bd-11ea-945b-7c4ee9dcc3b6.png)

sharkwouter:
This has been fixed thanks to @giancluciano

PR #90, opened 2020-03-14
Fixed installation instructions links in README.md

giancluciano:
Installation instructions look weird on the website #82
Hi @sharkwouter 
I couldn't test in the website, but I changed to HTML tags and kept working in Github.


sharkwouter:
That was the solution! Thanks a lot!

PR #94, opened 2020-03-19
Windows games

Odelpasso:
Add functions to use real name for games install files.
Add Wine implementation.

### DIFF
--- a/data/ui/gametile.ui
+++ b/data/ui/gametile.ui
@@ -10,20 +10,6 @@
         <property name="can_focus">False</property>
         <property name="orientation">vertical</property>
         <child>
-          <object class="GtkModelButton" id="menu_button_support">
-            <property name="visible">True</property>
-            <property name="can_focus">True</property>
-            <property name="receives_default">True</property>
-            <property name="text" translatable="yes" context="support" comments="Opens the official support webpage">Support</property>
-            <signal name="clicked" handler="on_menu_button_support_clicked" swapped="no"/>
-          </object>
-          <packing>
-            <property name="expand">False</property>
-            <property name="fill">True</property>
-            <property name="position">0</property>
-          </packing>
-        </child>
-        <child>
           <object class="GtkModelButton" id="menu_button_open">
             <property name="visible">True</property>
             <property name="can_focus">True</property>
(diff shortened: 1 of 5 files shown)


minigalaxy/ui/gametile.py
```diff
@@ -47,11 +48,11 @@ def __init__(self, parent, game, api):
 
         # Set folder for download installer
         self.download_dir = os.path.join(CACHE_DIR, "download")
-        self.download_path = os.path.join(self.download_dir, self.game.name)
+        self.download_path = os.path.join(self.download_dir, self.get_real_name())
 
         # Set folder if user wants to keep installer (disabled by default)
         self.keep_dir = os.path.join(Config.get("install_dir"), "installer")
-        self.keep_path = os.path.join(self.keep_dir, "{}.sh".format(self.game.name))
+        self.keep_path = os.path.join(self.keep_dir, self.get_real_name())
```
sharkwouter: This way of setting this cannot work. Using get_real_name here and on line 51 adds 4 api requests on the main thread for each game you own. For me that means I will have to wait for more than 1000 requests to the API have been made before the application starts responding again. We shouldn't need that many requests at all.

data/ui/gametile.ui
```diff
@@ -54,6 +54,46 @@
...
+      <object class="GtkBox">
+        <property name="visible">True</property>
+        <property name="can_focus">False</property>
+        <property name="orientation">vertical</property>
+        <child>
+          <object class="GtkModelButton" id="menu_win_winecfg">
+            <property name="visible">True</property>
+            <property name="can_focus">True</property>
+            <property name="receives_default">True</property>
+            <property name="text" translatable="yes" context="support" comments="Opens the official support webpage">Winecfg</property>
```
sharkwouter: Do change the translation information in glade for these buttons

sharkwouter: I also think the button should say something less specific which makes some sense to people not familiar with wine. Using "Settings" or "Configuration" would probably work better.

data/ui/gametile.ui
```diff
@@ -54,6 +54,46 @@
...
+            <signal name="clicked" handler="on_menu_button_winecfg_clicked" swapped="no"/>
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">True</property>
+            <property name="position">0</property>
+          </packing>
+        </child>
+        <child>
+          <object class="GtkModelButton" id="menu_win_regedit">
```
sharkwouter: Do we really need a regedit button? I think it's a bit much, it would probably be better without.

minigalaxy/installer.py
```diff
@@ -3,6 +3,7 @@
 import zipfile
 import subprocess
 import gi
+from os import listdir
```
sharkwouter: You can also use os.listdir in the code instead of importing listdir separately

minigalaxy/launcher.py
```diff
@@ -88,9 +102,25 @@ def __get_execute_command(game) -> list:
             return ["scummvm", "-c", scummvm_config]
 
     # Wine
-    if "prefix" in files and shutil.which("wine"):
```
sharkwouter: This code was for GOG games for Linux which use wine. It should not be deleted.

minigalaxy/launcher.py
```diff
@@ -4,12 +4,26 @@
...
 
 
+def config_game(game, options):
+
+    prefix_dir = os.path.join(Config.get("install_dir"), "prefix")
+    prefix = os.path.join(prefix_dir, game.name)
+
+    os.environ["WINEPREFIX"] = prefix
+
+    if options == "winecfg":
```
sharkwouter: just winecfg is better here, then you won't need a magic string either

minigalaxy/launcher.py
```diff
@@ -88,9 +102,25 @@ def __get_execute_command(game) -> list:
             return ["scummvm", "-c", scummvm_config]
 
     # Wine
-    if "prefix" in files and shutil.which("wine"):
-        # This still needs to be implemented
-        return [os.path.join(game.install_dir, "start.sh")]
+    if shutil.which("wine"):
```
sharkwouter: This check needs to change to something filename related or it should move somewhere else, either at the bottom of the method or in a different method

data/ui/gametile.ui
```diff
@@ -131,6 +171,33 @@
             <property name="index">1</property>
           </packing>
         </child>
+        <child type="overlay">
+          <object class="GtkMenuButton" id="menu_win">
```
sharkwouter: Could a menu item be added to the current settings button instead?

PR #95, opened 2020-03-20
Choose platform

Odelpasso:
This new PR is the previous PR (Windows games) + the new code.
It's just a try, i think something better can be done. But at this state it works for all my games.

I can choose for Xenonauts if i want to install Windows or Linux Version.

Negative point : The dialog box to choose the platform version. It's not the good box and i don't know how to have a correct box (with 3 buttons -> Cancel, Linux, Windows).
For the current box, Cancel button install Windows version and OK button install Linux version.

Like i said, it's just an Alpha code :P

### DIFF
--- a/data/ui/gametile.ui
+++ b/data/ui/gametile.ui
@@ -10,20 +10,6 @@
         <property name="can_focus">False</property>
         <property name="orientation">vertical</property>
         <child>
-          <object class="GtkModelButton" id="menu_button_support">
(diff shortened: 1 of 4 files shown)


sharkwouter:
I don't really like the idea of doing this with a dialog message. To be honest, I'm not even sure if I like giving people this option. I don't really see how we could do this without introducing clutter. I'll close it for now.

PR #100, opened 2020-03-21
Add Wine icon for Windows games

Odelpasso:


minigalaxy/paths.py
```diff
@@ -18,6 +18,8 @@
 if not os.path.exists(LOGO_IMAGE_PATH):
     LOGO_IMAGE_PATH = os.path.abspath(os.path.join(LAUNCH_DIR, "../share/icons/hicolor/192x192/apps/com.github.sharkwouter.minigalaxy.png"))
 
+LOGO_ICON_PATH = os.path.abspath(os.path.join(LAUNCH_DIR, "../data/icons/20x20/winehq_logo_glass.png"))
```
sharkwouter: This should work, but only if you're not installing system wide. That's why there is a check for LOCALE_DIR and LOGO_IMAGE_PATH. This should also be less generic. ICON_WINE_PATH would be better.

Odelpasso:
I updated the commit but still does not work. When i launch Minigalaxy, i have this error : 
"AttributeError: 'Pixbuf' object has no attribute 'show'"

PR #102, opened 2020-03-21
Enable FPS counter for Vulkan games and AMD users

Odelpasso:
Add variable environment for Mesa Vulkan FPS counter

### DIFF
--- a/minigalaxy/launcher.py
+++ b/minigalaxy/launcher.py
@@ -79,6 +79,14 @@ def __get_execute_command(game) -> list:
         prefix = os.path.join(game.install_dir, "prefix")
         os.environ["WINEPREFIX"] = prefix
 
+        # Enable ACO by default for AMD users.
+        # Mesa 19.x and > are needed.
+        # ACO allow to compile more quickly than with LLVM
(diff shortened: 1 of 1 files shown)


minigalaxy/launcher.py
```diff
@@ -79,6 +79,14 @@ def __get_execute_command(game) -> list:
         prefix = os.path.join(game.install_dir, "prefix")
         os.environ["WINEPREFIX"] = prefix
 
+        # Enable ACO by default for AMD users.
+        # Mesa 19.x and > are needed.
+        # ACO allow to compile more quickly than with LLVM
+        # and avoid stuttering in Vulkan games (only).
+        # This option is not enabled by default
+        # We'll can remove this variable when it'll be done
+        os.environ["RADV_PERFTEST"] = "aco"
```
sharkwouter: Is this needed for this?

Odelpasso: No, it's not needed to display FPS Vulkan but i did not want to do an other PR only for that ... :X
It's an other option no related to FPS display

sharkwouter: Ah okay. Could you remove it from this PR then?

PR #103, opened 2020-03-21
Add environment variable for Mesa ACO.

Odelpasso:
Useful for AMD GPU users and Vulkan games played by Wine.
Allow to reduce stuttering in game, shaders are compiled by ACO
and not LLVM.

Currently, this option is disabled by default. We'll can remove
this option once it will enabled.

### DIFF
--- a/minigalaxy/launcher.py
+++ b/minigalaxy/launcher.py
@@ -79,6 +79,14 @@ def __get_execute_command(game) -> list:
         prefix = os.path.join(game.install_dir, "prefix")
         os.environ["WINEPREFIX"] = prefix
 
+        # Enable ACO by default for AMD users.
+        # Mesa 19.x and > are needed.
+        # ACO allow to compile more quickly than with LLVM
+        # and avoid stuttering in Vulkan games (only).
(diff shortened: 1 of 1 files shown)


sharkwouter:
This pull request is a bit hard for me to judge. I can't really test it, since I don't own an affected graphics card.

I have read that using ACO can cause issues in some games, though, so it might be better to not include it and make the driver/OS figure it out. What do you think?

Odelpasso:
ACO could cause issues in some games but it was in the past, at the beginning during the development. Now it's implemented in Mesa so it's mature and very beneficial for Linux players.
And it will be enabled by default in a few months now.

What do you mean by : 
`so it might be better to not include it and make the driver/OS figure it out.`

sharkwouter:
Sorry about that, I'll explain better.

If Mesa is soon making ACO the default, I would say it would be better to keep this change out of the code base of Minigalaxy. Users can enable ACO themselves as well. I wouldn't want it to break anything for users with older versions of Mesa.

Thanks for your work, though! I really do appreciate it!

PR #105, opened 2020-03-22
Update pt_BR translation 

EsdrasTarsis:
pt_BR translation for the Wine feature in Minigalaxy, the zh_TW file has changed because I probably compiled it together with the pt_BR file using the compile-translations script.


sharkwouter:
Looks good to me. Thanks!

PR #107, opened 2020-03-25
Add initial OARS rating

tim77:
Need for https://github.com/flathub/flathub/pull/1402


sharkwouter:
Thanks! Looks good to me

Issue #110, opened 2020-03-31
No icons for already installed games


berarma:
After working around issue #109 my already installed games were found but the icon disappeared. Before they were found installed they had their icon.

I'm using the deb package version 0.9.3 in Debian 10.

sharkwouter:
That's odd, it should be downloading them. Did you start Minigalaxy with a working internet connection? How fast is your connection? Not that I expect that to be the reason, but I want to make sure to rule it out.

berarma:
It's weird because all icons were showing before I configured the game path and minigalaxy found the already installed games. When it found them it stopped showing their icons.

sharkwouter:
I was able to reproduce this on the latest code in git. I'll look into it.

sharkwouter:
This also affects showing the store page or not. I'm not creating a new issues for this, since the fix will be the same.

sharkwouter:
This should be fixed in master now.

wnhre2ur8cxx8:
This is still an issue. just installed 1.0 via AUR and the installed games show up but they are the only ones without icon.

sharkwouter:
Yeah, it isn't in the version in the AUR yet. It will be in the next release.

PR #111, opened 2020-03-31
Add link to a game's store page

larslindq:
Just a tiny addition that I had in my local repository.
I made some changes to how the tiles are populated but I'm not a Python expert so I might have broken something.

### DIFF
--- a/data/ui/gametile.ui
+++ b/data/ui/gametile.ui
@@ -52,6 +52,20 @@
           </packing>
         </child>
         <child>
+          <object class="GtkModelButton" id="menu_button_store">
+            <property name="visible">True</property>
+            <property name="can_focus">True</property>
+            <property name="receives_default">True</property>
+            <property name="text" translatable="yes" context="store" comments="Open store page">Store page</property>
+            <signal name="clicked" handler="on_menu_button_store_clicked" swapped="no"/>
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">True</property>
+            <property name="position">3</property>
+          </packing>
+        </child>
+        <child>
           <object class="GtkModelButton" id="menu_button_support">
             <property name="visible">True</property>
             <property name="can_focus">True</property>
@@ -62,7 +76,7 @@
           <packing>
             <property name="expand">False</property>
             <property name="fill">True</property>
-            <property name="position">3</property>
+            <property name="position">4</property>
           </packing>
         </child>
       </object>
--- a/data/ui/gametile.ui
+++ b/data/ui/gametile.ui
@@ -10,59 +10,17 @@
         <property name="can_focus">False</property>
         <property name="orientation">vertical</property>
         <child>
-          <object class="GtkModelButton" id="menu_button_open">
-            <property name="visible">True</property>
-            <property name="can_focus">True</property>
-            <property name="receives_default">True</property>
-            <property name="text" translatable="yes" context="open_files" comments="Opens the directory the game is installed in">Open Files</property>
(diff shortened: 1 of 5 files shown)


sharkwouter:
Oh wow, thanks for the pull request!

I will try it ou before the end of the week. I did look at the code for a bit and I did find a line I find odd in the library, I'll add a comment.

This is not something I've though about before honestly. Is it a feature you were missing?

minigalaxy/ui/library.py
```diff
@@ -45,7 +45,6 @@ def __update_library(self):
         GLib.idle_add(self.__load_tile_states)
         # Get already installed games first
         self.games = self.__get_installed_games()
-        self.__create_gametiles()
```
sharkwouter: I don't think this line should be removed

larslindq:
Yeah, I have some games with DLCs and it's just nice to have a quick way to go to the store to view which ones you have and whatnot.

minigalaxy/ui/library.py
larslindq: Yeah, I removed it because the store page URL is not available at the time where the tiles for installed games are created, and I noticed that create_gametiles is called later on anyway after self.games have been fully populated. The URL is read from the API in add_games_from_api but if the tile for a game already exists then it doesn't matter.

If you prefer to keep it as is I could take a second look. I think would need to find a way to update the tiles if they already exist in create_gametiles. The current behaviour is to skip the game if a tile for it already exists. The fixed behaviour would be to update the tile.

sharkwouter: This might be related to #110 then.

sharkwouter:
It does work well. The only issue I see is that the store page is shown when you go from offline mode to online, but I think the fix for that might be worse than the issue. I'll look into the one comment I made some more today. I'll merge it this afternoon.

sharkwouter:
Also I think the order of items should be:

- Store
- Support Page
- Open Files
- Uninstall

Should I make that change or would you like to?

data/ui/gametile.ui
```diff
@@ -51,6 +51,20 @@
             <property name="position">2</property>
           </packing>
         </child>
+        <child>
+          <object class="GtkModelButton" id="menu_button_store">
+            <property name="visible">True</property>
+            <property name="can_focus">True</property>
+            <property name="receives_default">True</property>
+            <property name="text" translatable="yes" context="store" comments="Open store page">Store page</property>
```
sharkwouter, line 59: Should be Store Page with capital letters

minigalaxy/ui/library.py
sharkwouter: Oh wow, I see it now. It is being called twice. Good catch!

sharkwouter: I just remembered the reason why I implemented it like this. If your internet connection is extremely slow, the tiles for the games you have installed will already show while. This is what is causing #110, but I think we should fix this is a different way. Can you revert this change?

larslindq: Ah I see. I reverted the removal.

larslindq:
I can re-order them. Will fix the text at the same time.

sharkwouter:
Yeah, that makes sense. Thanks!

larslindq:
> Yeah, that makes sense. Thanks!

Done!

sharkwouter:
Thanks! Merged!

I do see an issue with it not appearing for already installed games, but this is the same issue as the icons not showing, which I need to fix anyway.

sharkwouter:
I've solved the issue with it by replacing tiles which were made in offline mode by ones from the API when possible.

PR #112, opened 2020-04-05
Fix RDNS name according to guidelines

tim77:
https://docs.flatpak.org/en/latest/conventions.html#application-ids

https://github.com/flathub/flathub/pull/1402#issuecomment-606765971


sharkwouter:
Can you make sure setup.py still installs all files?

sharkwouter:
Thanks for updating! I'll try it out

sharkwouter:
Seems to work, thanks!

PR #118, opened 2020-04-06
[pt_BR] Update translation

EsdrasTarsis:


sharkwouter:
Thanks for the quick update!

PR #123, opened 2020-04-07
updated turkish translation

fuzunspm:


sharkwouter:
Thanks for the quick update! It looks good

PR #128, opened 2020-04-09
Update Polish translation

ArturWroblewski:
4 translations added

### DIFF
--- a/data/po/pl.po
+++ b/data/po/pl.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2020-04-06 21:15+0200\n"
-"PO-Revision-Date: 2020-03-10 09:31+0100\n"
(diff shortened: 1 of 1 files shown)


data/po/pl.po
```diff
@@ -243,10 +243,9 @@ msgstr "Pokaż FPS w grach:"
 
 #. Has to end with ": "
 #: data/ui/preferences.ui:201
-#, fuzzy
 msgctxt "show_windows_games"
 msgid "Show Windows games: "
-msgstr "Pokaż FPS w grach:"
+msgstr "Pokaż FPS w grach: "
```
sharkwouter, line 248: I think this should be changed. It still contains FPS

sharkwouter:
Thanks for the PR! It looks good, but I did find one small issue

data/po/pl.po
ArturWroblewski: You're right . I don't know how it could get there :)

ArturWroblewski:
> Thanks for the PR! It looks good, but I did find one small issue

Fixed

sharkwouter:
Thanks! I've merged it

Issue #98, opened 2020-03-21
Don't remove spaces from directory names when installing


sharkwouter:
Currently directories which are created contain the game name, but with the spaces removed. This is not needed. The filter function which cleans up the name and removes the spaces from it which is currently used should not be changed, though, instead a new one could be added.

This might require a lot of changes, so make sure to check all files for usages. The main user must be gametile.py at the moment.

larslindq:
I started looking at this and was thinking if it's possible to just change get_install_dir in gametile.py.

Instead of:
` return os.path.join(Config.get("install_dir"), self.game.get_stripped_name())`
Do this:
`return os.path.join(Config.get("install_dir"), self.game.name)`

What do you think? Just tried it locally and it seems to work fine. Will it break anything?

sharkwouter:
That might work, but to really test if it doesn't break anything you'd need to start Minigalaxy without internet connection and check if it still finds your installed games.

larslindq:
Tested without network connection and it detected my installed games fine.
` __get_installed_games` is quite robust in that it doesn't seem to care about the name of the game directory itself, only if there is a gameinfo file in it or not.

sharkwouter:
This has been resolved by @larslindq, thanks for that!

PR #129, opened 2020-04-11
#98 Don't remove spaces from directory names when installing

larslindq:
I have only discovered one instance to change, so I'm opening a PR.

I have tested locally with two already installed games (native and Wine) with their old directories as well as having installed two new games with their new directories. Both scenarios have worked.


sharkwouter:
Thanks for testing it! Looks good to me

PR #135, opened 2020-04-20
Quick fixes and follow GNOME translation guidelines

tim77:


sharkwouter:
Thanks! I can't really judge the quality of the translation, but I don't see anything weird or breaking, so I'll assume it's an improvement.

PR #140, opened 2020-04-24
Add repology badge to README.md

not-chicken:
I believe it would be useful for users to see the packaging status on the readme. I am not sure where would be a good spot on readme for it though.
I have choosen to right-align it here to avoid wasting space.

repology.

There are also a few smaller badge variants https://repology.org/project/minigalaxy/badges if you'd like something smaller.


sharkwouter:
Thanks! It looks quite okay on the website as well, really useful

PR #144, opened 2020-04-27
Create a new config when config.json is empty

SvdB-nonp:
Got an issue where the config.json file is empty, which prevents the program from being launched as the json.decode will fail on this file. Added a print statement to log it, but you may replace this with actual logging how you guys see fit.


sharkwouter:
This is a really good idea, thanks a lot! I can't test it right now, so I'll do so tomorrow evening

sharkwouter:
It works well, thanks for the work!

PR #147, opened 2020-04-29
Add disk space checking before download.

SvdB-nonp:
Currently minigalaxy will simply download until the disk is full, which may slow down the pc while giving no benefit (as download cannot be completed anyway). This mainly concerns the .cache folder, where the files reside before finishing.
To prevent this, this code will check the disk space before initialising the download. For Linux it should prevent the whole download, as it is one file, while for Windows it will only stop downloading the last file that doesn't fit.

Edit: Currently also analysing whether to add disk space validation to extracting installer, saving installer to install dir, and copying to final install destination.

Also removed the obsolete save_file.close() command, as the code already uses the 'with' context manager to close save_file upon any exit of the block.

Edit:
147: PR for disk space checking
- issue 155

### DIFF
--- a/minigalaxy/download_manager.py
+++ b/minigalaxy/download_manager.py
@@ -37,10 +37,7 @@ class __DownloadManger:
         diskspace_available = self.get_availablediskspace(location)
         # Reserved space can be used to mimic a full disk or prevent filling a disk completely.
         reserved_space = 0
-        if diskspace_available < (download_remaining + reserved_space):
-            return False
-        else:
-            return True
+        return False if diskspace_available < (download_remaining + reserved_space) else True
 
     def download(self, download):
         if isinstance(download, Download):
@@ -124,11 +121,6 @@ class __DownloadManger:
             download.cancel()
             self.__current_download = None
             print("There is insufficient disk space to cache the download.")
-            # self.window.show_error(
-            #     text="There is insufficient disk space to cache the download.",
-            #     secondary_text=(f"Cache location is '{CACHE_DIR}' ({self.get_availablediskspace(download.save_location)})"
-            #                     f" and current download is {file_size}."),
-            # )
         else:
             if downloaded_size < file_size:
                 with open(download.save_location, download_mode) as save_file:
--- a/minigalaxy/download_manager.py
+++ b/minigalaxy/download_manager.py
@@ -8,6 +8,17 @@ from minigalaxy.download import Download
 
 
 class __DownloadManger:
+    #TODO: Move static method to ui.gametile?
+    @staticmethod
+    def get_availablediskspace(location):
+        """Check disk space available to the user. This method uses the absolute path so
+        symlinks to disks with sufficient space are correctly measured. Note this is
+        a linux-specific command."""
+        absolute_location = os.path.realpath(location)
+        disk_status = os.statvfs(os.path.dirname(absolute_location))
+        available_diskspace = disk_status.f_frsize * disk_status.f_bavail
+        return available_diskspace
+
     def __init__(self):
         self.__queue = queue.Queue()
         self.__current_download = None
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -22,6 +22,7 @@ def install_game(game, installer, main_window=None) -> None:
 
         # Make a temporary empty directory for extracting the installer
         temp_dir = os.path.join(CACHE_DIR, "extract/{}".format(game.id))
+        # TODO: Add disk space checking for extracting installer?
         if os.path.exists(temp_dir):
             shutil.rmtree(temp_dir, ignore_errors=True)
         os.makedirs(temp_dir)
@@ -38,6 +39,7 @@ def install_game(game, installer, main_window=None) -> None:
             os.makedirs(library_dir)
 
         # Copy the game files into the correct directory
+        # TODO: Add disk space checking for final install directory?
         shutil.move(os.path.join(temp_dir, "data/noarch"), game.install_dir)
 
         # Remove the temporary directory
@@ -65,6 +67,7 @@ def install_game(game, installer, main_window=None) -> None:
         download_dir = os.path.join(CACHE_DIR, "download")
         if not os.path.exists(keep_dir):
             os.makedirs(keep_dir)
+        # TODO:Add disk space checking for keep_dir location?
         try:
             # It's needed for multiple files
             for file in os.listdir(download_dir):
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -198,6 +198,7 @@ class GameTile(Gtk.Box):
     def __install(self):
         GLib.idle_add(self.update_to_state, self.state.INSTALLING)
         self.game.install_dir = self.__get_install_dir()
+        # TODO: Add disk space checking for installing the game?
         try:
             if os.path.exists(self.keep_path):
                 install_game(self.game, self.keep_path, main_window=self.parent.parent)


sharkwouter:
Sorry for the late reply. I think this is a great idea, thanks! I would like to ask for one adjustment, though. Could you make the disk space check into a separate method? That way it would be a bit cleaner

SvdB-nonp:
> Sorry for the late reply. I think this is a great idea, thanks! I would like to ask for one adjustment, though. Could you make the disk space check into a separate method? That way it would be a bit cleaner

That is more pythonic, so I agree. I actually will take a look at symlinking as well, just to confirm it will not falsely raise an error while the location is symlinked to a disk with enough space.

SvdB-nonp:
Updated this with relevant code moved to separate functions. My tests seem all fine (including app restart continuing downloads), but maybe ytou can verify that the last two if-statements indeed need to be skipped when this error occurs. If not, they need to be outside the Else block.

sharkwouter:
I've looked at the code again. I see the problem you ran into now. I'll be making a small change which would allow you to actually show an error instead of raising an exception. Raising an exception makes the application unusable, which is not really what we want.

sharkwouter:
I know this make it a little bit more complex, but what you could do is something like this:
```
diff --git a/minigalaxy/download_manager.py b/minigalaxy/download_manager.py
index adc4684..b0dbe85 100644
--- a/minigalaxy/download_manager.py
+++ b/minigalaxy/download_manager.py
@@ -13,6 +13,7 @@ class __DownloadManger:
         self.__current_download = None
         self.__cancel = False
         self.__paused = False
+        self.window = None
 
         download_thread = threading.Thread(target=self.__download_thread)
         download_thread.daemon = True
diff --git a/minigalaxy/ui/window.py b/minigalaxy/ui/window.py
index 2931353..4256bac 100644
--- a/minigalaxy/ui/window.py
+++ b/minigalaxy/ui/window.py
@@ -9,6 +9,7 @@ from minigalaxy.api import Api
 from minigalaxy.config import Config
 from minigalaxy.paths import UI_DIR, LOGO_IMAGE_PATH, THUMBNAIL_DIR
 from minigalaxy.ui.library import Library
+from minigalaxy.download_manager import DownloadManager
 
 
 @Gtk.Template.from_file(os.path.join(UI_DIR, "application.ui"))
@@ -25,13 +26,16 @@ class Window(Gtk.ApplicationWindow):
     menu_logout = Gtk.Template.Child()
     window_library = Gtk.Template.Child()
 
-    def __init__(self, name):
+    def __init__(self, name="Minigalaxy"):
         Gtk.ApplicationWindow.__init__(self, title=name)
         self.api = Api()
         self.show_installed_only = False
         self.search_string = ""
         self.offline = False
 
+        # Make the DownloadManager able to output errors
+        DownloadManager.window = self
+
         # Set library
         self.library = Library(self, self.api)
         self.window_library.add(self.library)

```
In that way you can use window as a parent for an error message. I've added a show_error function to the Window class in ``minigalaxy/ui/window.py``, you can use that to show an error. Then you just cancel the download without raising an exception. The error message should contain which file failed to download.

I would make the function which checks if there is enough disk space except a file size and return true or false.

SvdB-nonp:
Updated based on your latest comments. 
1. Added appropriate cancel methods. Now it doesn't even prevent following downloads in the queue from downloading. Tested this by putting 3 downloads in the queue, where only the 2nd is too large. 1 and 3 both succesfully downloaded respectively before and after the 2nd cancelled due to this error.
2. Changed it so it returns true or false instead of an error.
3. Added the necessary code to call window.showerror.
However, I couldn't get it to work due to this error:
`Gdk-Message: 15:52:06.128: minigalaxy: Fatal IO error 2 (No such file or directory) on X server :0.`
I have no clue what this is, it lacks any tracktrace as it is not a Python error, but I kept the code in (commented) so anyone else can test it on a different linux installation (I used default Ubuntu 20.04).

I will now work on dots and colons preventing a game to install, but for clarity I think it is best to keep this in a different pull request.

sharkwouter:
Thanks for updating. I'll see if I can take a look tonight. The error you mention is a bit odd, I'll see if I can reproduce it.

The fix for the dots and colons preventing game installations should be a different PR indeed. I've found out the hard way that pull requests with many features in it are very hard to review well.

sharkwouter:
I didn't find time to go over this yesterday. I'll try again on Sunday.

SvdB-nonp:
No worries. Lack of time is also a problem on my end (hence the large gap of inactivity).
When I do actually have time I can always try to debug it some more myself. Need to learn more about threading in python anyway.

minigalaxy/download_manager.py
```diff
@@ -5,19 +5,43 @@
...
         download_thread.start()
 
+    def check_diskspace(self, file_size, downloaded_size, location):
+        """This method will return True when the disk space available is sufficient
+        for the Download. If not sufficient, it returns False."""
+        download_remaining = file_size - downloaded_size
+        diskspace_available = self.get_availablediskspace(location)
+        # Reserved space can be used to mimic a full disk or prevent filling a disk completely.
+        reserved_space = 0
+        if diskspace_available < (download_remaining + reserved_space):
```
sharkwouter, line 40: This could be simplified. Something like ``return not diskspace_available < (download_remaining + reserved_space)`` should work.

minigalaxy/download_manager.py
```diff
@@ -96,29 +120,38 @@ def __download_file(self, download):
...
-        if download.number == download.out_of_amount:
-            finish_thread = threading.Thread(target=download.finish)
-            finish_thread.start()
-        if self.__queue.empty():
-            Config.unset("current_download")
+        if not self.check_diskspace(file_size, downloaded_size, download.save_location):
+            download.cancel()
+            self.__current_download = None
+            print("There is insufficient disk space to cache the download.")
+            # self.window.show_error(
```
sharkwouter, line 127: The commented out code should be removed when this PR is finished.

minigalaxy/ui/window.py
```diff
@@ -32,6 +33,9 @@ def __init__(self, name="Minigalaxy"):
         self.search_string = ""
         self.offline = False
 
+        # Make the DownloadManager able to output errors
+        DownloadManager.window = self
```
sharkwouter, line 37: I feel like we should be able to come up with a better solution for this, but I'll need to think about it. I am thinking of making the Window class a singleton, so you can just call ``self.window = Window()`` in the constructor of the DownloadManager without causing circulair dependencies. Alternatively we could find a way to pass the window to the downloadmanager through the constructor, but for that the downloadmanager is not created in the right place at the moment.

sharkwouter: Sorry about this, I know this solution was my idea. This one wouldn't stop me from merging, it's more of a note to myself.

SvdB-nonp: No worries. I would actually propose to delay this PR until that is finished, since what it fixes already rarely occurs and we then can solve it cleanly.

Also, I can then take a look at disk checking elsewhere where it is relevant (e.g. extracting and final download location come to mind).

Making this PR a draft for the time being.

sharkwouter:
@SvdB-nonp I'm looking into merging this now, but there is a little bit of an issue, which is that I don't know how to test this. Would it be possible for you to write a test case for it which confirms that it works?

SvdB-nonp:
Do you mean an automated test case or just how to test it maually?
I simply used the reserved space variable to test it. So my root disk is 40Gb here and then I set reserved space to 30Gb and it will fail on every game that is larger than 10Gb. If you have a large disk, you may have to set it pretty high.

The number is very high, so have to find out what I used to have it reserve 30Gb. Maybe I will do a small commit to make that easier. E.g.:
`reserved space = number_in_gigabytes * conversion_factor`

makson96:
This probably needs to be rebase with current master, as there are changes affecting it:
- no keep_dir
- changes in download_manager.py

minigalaxy/download_manager.py
```diff
@@ -8,6 +8,17 @@
 
 
 class __DownloadManger:
+    #TODO: Move static method to ui.gametile?
```
makson96, line 11: It is ok where it is. I would keep only UI stuffs in ui.gametile.
If you don't like this static method you can move it out of __DownloadManager() class as function not related to this class.

sharkwouter:
Good point, I still think it would be good to have this

SvdB-nonp:
I mainly wanted to move it, so it can be used in other parts of the program as well. And since most things are spawned from gametile, it seemed most practical (although I agree it should be UI-only).

I like the suggestion to move it outside of the class, so at least it is clear it can be moved without affecting other parts of the class.

sharkwouter:
Good idea. It might make sense add the function to see if there is enough disk space for a file of a specific size in ``install.py``

makson96:
Hi,
I put some more thoughts on this one and would like to propose a little improvement.
You put some comments in the code where you see that disk space check should take place. I am not sure that is a good idea. With such an approach we might end up in the middle of installation process with info "Well we won't have enough space for it after all". And what to do now?
I propose that disk space check should be performed only once - right after clicking download by user. We should collect some experimental data and estimate what is the size ratio between installer and installed game and use that info to estimate if we have enough space for whole installation process to take place.

What is more, from what I see, if we have not enough space, we only print info about it to stdout. I think that this is not enough and user should also receive proper error window. Ideally it would be something like this:
"
It seems that you don't have enough free disk space for this installation. Do you want to continue anyway?
OK                    Cancel
"

SvdB-nonp:
Yes, I agree with it all. I am not sure whether there are size differences between extract/install/download, hence checking it multiple times. But I agree that if we can do it at once, it is infinitely better.

The error window was in at some point, but there was a gtk error which I was unable to solve. Note the removed comments in the last commit, the draft code for it is there in case that helps.

sharkwouter:
I'm closing this in favor of #291 which continues on this work. Thank for the contribution!

PR #154, opened 2020-05-14
Added Simplified Chinese translation

dummyx:


sharkwouter:
Thanks for the contribution. I don't know Chinese, but I don't see any formatting errors, so I'll merge it.

PR #162, opened 2020-07-05
Update desktop and appstream files

protheory8:


sharkwouter:
I think the 0.9.0 release should stay in there. The rest is looking pretty good.

PR #163, opened 2020-07-24
Installer name filter dots colons

SvdB-nonp:
This PR is to fix the S.T.A.L.K.E.R. series (and probably others) installing through minigalaxy due to weird characters in installer file names. Install directory name was already stripped, and I use the same function to strip the installer names.

- When only a single .bin file is present, don't append a number.
_Shadow of Chernobyl installer only recognises the .bin when it is exactly the same name as the .exe. I made it generic in case other (old) gog titles have the same limitation._
- Installer file names are now stripped, just like the prefix directories.
_The dots prevent the GOG installers from recognising the bin files. I have no other games with dots in the name, so it can be the same for other games as well. This has the additional benefit that the gog installers have the stripped dirname as default (same as prefix) when launched._
- Keep dir name is also stripped so reading/writing are all based on the stripped name.
_Without this, the program will always redownload installers, as it does not recognize the stripped names of the installers.

Edit:
163: PR for installer name trimming
- Issue 150 (mentions silent install... needed?)
- Issue 148


minigalaxy/ui/gametile.py
```diff
@@ -178,17 +178,20 @@ def __download_file(self) -> None:
         self.download = []
         download_path = self.download_path
         finish_func = self.__install
+        number_of_files = len(download_info['files'])
         for key, file_info in enumerate(download_info['files']):
-            if key > 0:
+            if key > 0 and number_of_files > 2:
```
sharkwouter, line 183: This breaks the installation for games with only one bin file.

sharkwouter (changes requested):

minigalaxy/ui/gametile.py
sharkwouter: An example is prince of persia

PR #164, opened 2020-07-24
Remove cached download on cancel

SvdB-nonp:
This fixes cached downloads not being deleted when cancelled. When canceling a lot, this can make the cache very large and it is never cleared without finishing these downloads.

The UI should not delete files itself, however the downloadmanager only knows about 1 single file without the game name (or install dir name) to remove the others (completed or not).

Hence to keep the code change as minimal as possible, I added it in gametile.py as all information necessary to delete all related files is here. 
The only downside is that it will delete also cached files for next iterations of a game (Cancelling 'someshooter' will also remove cached files of 'someshooter 2'). However, normally there should only be one game cached at a time (at least, after this PR is merged).

Edit:
164: PR for cache removal
- Issue 119 (also mentions cache/extracts)

### DIFF
--- a/data/io.github.sharkwouter.Minigalaxy.metainfo.xml
+++ b/data/io.github.sharkwouter.Minigalaxy.metainfo.xml
@@ -8,14 +8,17 @@
     <summary>A simple GOG client for Linux</summary>
 
     <description>
-        <p>Minigalaxy lets you download, install and play your GOG Linux games without getting in your way.</p>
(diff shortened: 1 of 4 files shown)


SvdB-nonp:
Removed all changes not specific to this pull request.

Additional note: If we wanted a function to pause downloads, that is sorta what this code did until this change is pulled.

sharkwouter:
Thanks, this works well! Sorry for not looking into this sooner

Issue #158, opened 2020-06-02
"workingDir" from goggame.info is not used


Zauberfisch:
The `goggame-\d+.info` file of some games contains `"workingDir": "..."` to indicate the binary is to be executed in a certain directory.

An example game is `The Witcher 3: Wild Hunt - Game of the Year Edition`
```
{
    "clientId": "49292044087338605", 
    "gameId": "1495134320", 
    "language": "English", 
    "languages": [
        "en"
    ], 
    "name": "The Witcher 3: Wild Hunt - Game of the Year Edition", 
    "overlaySupported": true, 
    "playTasks": [
        {
            "category": "game", 
            "isPrimary": true, 
            "languages": [
                "*"
            ], 
            "name": "The Witcher 3: Wild Hunt - Game of the Year Edition", 
            "path": "bin//x64//witcher3.exe", 
            "type": "FileTask", 
            "workingDir": "bin//x64"
        }, 
        {
            "languages": [
                "*"
            ], 
            "link": "http://www.gog.com/en/support/the_witcher_3_wild_hunt", 
            "name": "Support", 
            "type": "URLTask"
        }
    ], 
    "rootGameId": "1495134320", 
    "version": 1
}
```

Starting the game:

* **ERROR** Starting through minigalaxy:
```
Failed to start The Witcher 3: Wild Hunt - Game of the Year Edition:
00c0:fixme:ntdll:NtQuerySystemInformation info_class SYSTEM_PERFORMANCE_INFORMATION
00c8:fixme:ver:GetCurrentPackageId (000000000363FDA0 0000000000000000): stub
00c0:fixme:msvcrt:__clean_type_info_names_internal (0x2411878) stub
00c0:fixme:msvcrt:__clean_type_info_names_internal (0x2325768) stub
```
* **ERROR** `cd "/home/$USER/GOG Games/Witcher3"; WINEPREFIX="/home/$USER/GOG Games/Witcher3/prefix" wine /home/$USER/GOG\ Games/Witcher3/bin/x64/witcher3.exe`
```
00c0:fixme:ntdll:NtQuerySystemInformation info_class SYSTEM_PERFORMANCE_INFORMATION
00c8:fixme:ver:GetCurrentPackageId (000000000332FDA0 0000000000000000): stub
00c0:fixme:msvcrt:__clean_type_info_names_internal (0x2411878) stub
00c0:fixme:msvcrt:__clean_type_info_names_internal (0x2325768) stub
```
[...]

PR #171, opened 2020-08-31
Checking for workingDir property and executing wine accordingly (Issue #158)

kibun1:
I've added two lines checking for the `workingDir` property in the `goggame-\d+.info` file and executing wine accordingly.
I'm not sure if this is also needed in non-wine program launches, at least I haven't found a game where that is the case, so for now I have left that out.


sharkwouter:
Thanks! This looks good to me and doesn't seem to break games which don't have a workdir, so I'll merge it and ask people to test if this fixes Witcher 3.

Issue #169, opened 2020-08-18
download option: gog's filename


arrowgent:
in addition to adding the option "download" with "install"
#161

it would be nice if the downloaded file was named the same as the file from gog's website
i dont know exactly how minigalaxy pulls the file from gog api

but an example is:

```
/home/$USER/.cache/minigalaxy/download/Thief™ Gold
```
should be
```
/home/$USER/.cache/minigalaxy/download/setup_thief_gold_1.26_nd_(21948).exe
```

similar happens for .sh files for linux versions

keep up the good work.

kibun1:
It seems to be possible to extract the filename using the download link (`api.get_real_download_link(file_info["downlink"])`)
For example, when I download rollercoaster tycoon (and it seems to be the same for other games), the real download link is

> https://content-system.gog.com/products/1207658945/users/USERID/namespaces/website/download?path=%2Fsecure%2Foffline%2Fwhatever%2F9133%2Fsetup_rollercoaster_tycoon_deluxe_1.20.015_%2817822%29.exe&token=TOKEN&access_token=ACCESS_TOKEN

A possible regex to extract the token using `re.search()` would be this: `'%2F(((?!%2F).)*)&token'`
I haven't sufficiently tested this yet, and at first glance changing the target filename of the download seems to be a relatively significant change, but if this is a desirable change then I could work on that.

sharkwouter:
@kibun1 I would say it is desirable that this will be changed, but it is somewhat complex. As we discussed in discord, the solution you mention here solves half the problem. In addition to that change, it would be good to put the installer in a directory with the same name as the installation directory in the installer directory if "keep installers" is enabled. That way, we can still check if the installer is there.

sharkwouter:
This should be fixed inn masterr now. Can you test it @arrowgent?

arrowgent:
> This should be fixed inn masterr now. Can you test it @arrowgent?

heyo.....
ok after reading this and looking at how to do this (also looking at the commits ~ sept 3 ~)

i got the master and launched ./minigalaxy

found the previous testcase: setup_thief_gold_1.26_nd_(21948).exe

appears to be created correctly in:
~/.cache/minigalaxy/download/Thief Gold/


mind you, i am using Ubuntu 18.04 LTS and have minigalaxy manually installed under /opt/minigalaxy
not that this matters for this issue.


everything looks good here

arrowgent:
additional:

[...]

PR #172, opened 2020-09-02
Retaining installation filenames

kibun1:
Downloaded installers will now retain their original filename, and kept installation files will be sorted in subdirectories with their game's name. `Installable` state is now checked by whether there is an executable file (+x flag or .exe/.sh file extension) in that game's directory.

Needs testing:
- Are all native linux game installers either .sh or marked for execution with the +x flag?
- Are all windows game installers .exe files?
- Does this fix broken installation for games that require the filenames to be as determined by gog?
- Is the direct download url for all games similar enough for the regex to extract the filename?

### DIFF
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -52,13 +52,12 @@ class GameTile(Gtk.Box):
         self.image.set_tooltip_text(self.game.name)
 
         # Set folder for download installer
-        self.download_dir = os.path.join(CACHE_DIR, "download")
+        self.download_dir = os.path.join(CACHE_DIR, "download", self.game.get_install_directory_name())
         self.download_path = os.path.join(self.download_dir, self.game.get_install_directory_name())
 
         # Set folder if user wants to keep installer (disabled by default)
         self.keep_dir = os.path.join(Config.get("install_dir"), "installer")
         self.keep_path = os.path.join(self.keep_dir, self.game.get_install_directory_name())
-
         if not os.path.exists(CACHE_DIR):
             os.makedirs(CACHE_DIR)
 
@@ -171,6 +170,13 @@ class GameTile(Gtk.Box):
             return True
         return False
 
+    def get_keep_executable_path(self):
+        if os.path.exists(self.keep_path):
+            for fil in os.scandir(self.keep_path):
+                if os.access(fil.path, os.X_OK) or os.path.splitext(fil)[-1] == ".exe" or os.path.splitext(fil)[-1] == ".sh":
+                    return fil.path
+        return ""
+
     def __download_file(self) -> None:
         Config.set("current_download", self.game.id)
         GLib.idle_add(self.update_to_state, self.state.QUEUED)
@@ -209,8 +215,9 @@ class GameTile(Gtk.Box):
         GLib.idle_add(self.update_to_state, self.state.INSTALLING)
         self.game.install_dir = self.__get_install_dir()
         try:
-            if os.path.exists(self.keep_path):
-                install_game(self.game, self.keep_path, main_window=self.parent.parent)
+            keep_executable_path = self.get_keep_executable_path()
+            if keep_executable_path:
+                install_game(self.game, keep_executable_path, main_window=self.parent.parent)
             else:
                 install_game(self.game, self.download_path, main_window=self.parent.parent)
         except (FileNotFoundError, BadZipFile):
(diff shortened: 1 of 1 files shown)


kibun1:
This should fix issue #169 as well as some other installation failures involving filenames of installfiles

minigalaxy/ui/gametile.py
```diff
@@ -180,10 +188,18 @@ def __download_file(self) -> None:
...
-            if key > 0:
-                download_path = "{}-{}.bin".format(self.download_path, key)
+            download_url = self.api.get_real_download_link(file_info["downlink"])
+            try:
+                # Extract the filename from the download url (filename is between %2F and &token)
+                download_path = os.path.join(self.download_dir, urllib.parse.unquote(re.search('%2F(((?!%2F).)*)&t', download_url).group(1)))
+                if key == 0:
+                    # Iff key = 0, denote the file as the executable's path
+                    self.download_path = download_path
+            except AttributeError:
```
sharkwouter, line 198: When would this exception trigger?

minigalaxy/ui/gametile.py
```diff
@@ -180,10 +188,18 @@ def __download_file(self) -> None:
...
         number_of_files = len(download_info['files'])
         for key, file_info in enumerate(download_info['files']):
-            if key > 0:
-                download_path = "{}-{}.bin".format(self.download_path, key)
+            download_url = self.api.get_real_download_link(file_info["downlink"])
+            try:
+                # Extract the filename from the download url (filename is between %2F and &token)
+                download_path = os.path.join(self.download_dir, urllib.parse.unquote(re.search('%2F(((?!%2F).)*)&t', download_url).group(1)))
+                if key == 0:
+                    # Iff key = 0, denote the file as the executable's path
```
sharkwouter, line 196: Typo in if

minigalaxy/ui/gametile.py
```diff
@@ -169,6 +170,13 @@ def __set_image(self):
             return True
         return False
 
+    def get_keep_executable_path(self):
+        if os.path.exists(self.keep_path):
+            for fil in os.scandir(self.keep_path):
```
sharkwouter, line 175: This causes an exception for me:
```
  File "/home/wouter/Sources/minigalaxy/minigalaxy/ui/library.py", line 108, in __create_gametiles
    self.__add_gametile(game)
  File "/home/wouter/Sources/minigalaxy/minigalaxy/ui/library.py", line 111, in __add_gametile
    self.flowbox.add(GameTile(self, game, self.api))
  File "/home/wouter/Sources/minigalaxy/minigalaxy/ui/gametile.py", line 64, in __init__
    self.reload_state()
  File "/home/wouter/Sources/minigalaxy/minigalaxy/ui/gametile.py", line 265, in reload_state
    elif self.get_keep_executable_path():
  File "/home/wouter/Sources/minigalaxy/minigalaxy/ui/gametile.py", line 175, in get_keep_executable_path
    for fil in os.scandir(self.keep_path):
NotADirectoryError: [Errno 20] Not a directory: '/home/wouter/Games/installer/METAL SLUG'
```

sharkwouter: To add to this, the rest of the gametiles are not being loaded because of this exception as well, so this is critical.

sharkwouter (changes requested):
Thanks! The code look good in general. I like the structure. It still has a critical issue which causes an exception for me, though. Can you look into that?

minigalaxy/ui/gametile.py
kibun1: Iff is short for if and only if (https://en.wikipedia.org/wiki/If_and_only_if) but I can change it to a normal if if you want, it doesn't really matter.

minigalaxy/ui/gametile.py
kibun1: It would trigger in case the regular expression can't find any matching string in the link, so `re.search()` returns None and `<None>.group(1)` triggers the exception

sharkwouter: Ah okay, that makes sense. Thanks!

minigalaxy/ui/gametile.py
kibun1: Ah, I see, that's an oversight by me, because I didn't have any installers saved before I made the change. Old installers have the same name as the directory where the new installers should be located, so that problem should only arise for old installer files.
I will add a quick directory check to fix that.

minigalaxy/ui/gametile.py
sharkwouter: I had no idea. It's probably clearer to change it

kibun1:
Newest commit should fix the "Not a directory" error and changed the comment

sharkwouter:
I just tested the new change. It seems to work fine. Will merge it

PR #176, opened 2020-09-13
Track downloads

sharkwouter:
I'm still not sure if this is the cleanest way to do this, but it is a working solution. The main downside is that the popover menu does not close automatically when the last download is finished. The button does hide when you then close it. Downloads are never deleted, just hidden. That could also be improved.

The overview shows the downloads per file. It shows a progress bar and a cancel button. The cancel button does not have an "are you sure?" warning yet. It also shows which downloads are queued and which one is being downloaded.

I think there are some things not happening which should be happening when a download is cancelled. I need to look into this more.

### DIFF
--- a/data/ui/application.ui
+++ b/data/ui/application.ui
@@ -140,6 +140,26 @@
           </packing>
         </child>
         <child>
+          <object class="GtkMenuButton" id="header_downloads">
+            <property name="can_focus">True</property>
+            <property name="receives_default">False</property>
+            <property name="no_show_all">True</property>
+            <property name="valign">center</property>
+            <property name="use_popover">False</property>
+            <child>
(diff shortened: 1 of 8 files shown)


sharkwouter:
Cancelling downloads doesn't work yet

makson96:
I think we should keep Gtk away from minigalaxy/download.py and minigalaxy/download-manager.py.
If new GUI file is needed it would be best to create one in minigalaxy/ui directory and import to it suitable non-gui modules.
Also this branch is way behind master and there were important changes in mentioned modules.

If you would like me to take a look at part with Python code it would be the best update this branch first.

sharkwouter:
I'm going to close this one. It's not implemented in the way it should.

PR #183, opened 2020-10-27
Dlc support

makson96:
OK. This pull request is to address issue #4 
I tested it on Stellaris and Neverwinter Nights and it seems to work fine for me:
neverwinter_dlc
There are few notes:
- in offline mode, there is no DLC button. Technically there is no reason why user cannot check installed DLCs in offline mode. If this is real use case, maybe we should create another issue for it.
- DLCs do not support update at the moment. What I saw from Stellaris DLCs installers names, they are connected to game version, so this might be a big deal. Please decide if it should be addressed in separate issue or in this pull request.
- Some DLCs (Stellaris: Infinite Frontiers eBook) points to game installer itself which results in game redownload and reinstall. Please decide if it should be addressed in separate issue or in this pull request.
- Some DLCs (Stellaris: Arachnoid Portrait Pack) cause following error in api.get_real_download_link:
"
  File "/home/makson/Publiczny/minigalaxy/GitHub/minigalaxy/build/lib/minigalaxy/ui/gametile.py", line 399, in <lambda>
    self.dlc_dict[title][0].connect("clicked", lambda x: self.__download_dlc(installer))
  File "/home/makson/Publiczny/minigalaxy/GitHub/minigalaxy/build/lib/minigalaxy/ui/gametile.py", line 324, in __download_dlc
    result = self.__download(download_info, finish_func, cancel_to_state)
  File "/home/makson/Publiczny/minigalaxy/GitHub/minigalaxy/build/lib/minigalaxy/ui/gametile.py", line 233, in __download
    download_url = self.api.get_real_download_link(file_info["downlink"])
  File "/home/makson/Publiczny/minigalaxy/GitHub/minigalaxy/build/lib/minigalaxy/api.py", line 156, in get_real_download_link
    return self.__request(url)['downlink']
  File "/home/makson/Publiczny/minigalaxy/GitHub/minigalaxy/build/lib/minigalaxy/api.py", line 193, in __request
    return response.json()
  File "/usr/lib/python3/dist-packages/requests/models.py", line 897, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
"
It is properly handled (ValueError) and proper message is returned to user, but I don't know what cause this error and how to fix it.
- If you find DLC button odd, could you try to fix it in Glade and propose solution.

Some others note from the work:
- Performance. reload_state for all tiles is our most expensive procedure, which can be noticed by user. I have tracked some cases, where it was executed two times in the row and I have deduplicated it. In a result I think, that aplication performance has improved, but please check if there are some corner cases, where tiles are not updated properly (I haven't seen any). Downloading DLCs icons is done in async now, so it should not hurt performance.
- Cleaning. I have deduplicated some code in gemetiles.ui. I hope it is now more clean and easier to maintain.

Waiting for the feadback.
Tomek

### DIFF
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -346,21 +346,17 @@ class GameTile(Gtk.Box):
 
     def __check_for_dlc(self, game_info):
         dlcs = game_info["expanded_dlcs"]
-        dlcs_present = False
         for dlc in dlcs:
             d_installer = dlc["downloads"]["installers"]
             if d_installer:
-                dlcs_present = True
                 d_icon = dlc["images"]["sidebarIcon"]
                 d_name = dlc["title"]
                 d_status = self.game.get_dlc_status(d_name)
                 self.dlc_box(d_icon, d_name, d_status, d_installer)
                 if dlc not in self.game.dlcs:
                     self.game.dlcs.append(dlc)
-        if dlcs_present:
+        if self.game.dlcs:
             self.menu_button_dlc.show()
-        else:
-            self.menu_button_dlc.hide()
 
     def dlc_box(self, icon, title, status, installer):
         if title not in self.dlc_dict:
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -1,13 +1,15 @@
 import shutil
 import gi
 gi.require_version('Gtk', '3.0')
-from gi.repository import Gtk, GLib, Gdk, GdkPixbuf
+from gi.repository import Gtk, GLib, Gdk, GdkPixbuf, Gio
 import os
 import webbrowser
 import threading
 import subprocess
 import re
 import urllib.parse
+import urllib.request
+from gi.repository.GdkPixbuf import Pixbuf
 from enum import Enum
 from zipfile import BadZipFile
 from minigalaxy.translation import _
@@ -37,6 +39,11 @@ class GameTile(Gtk.Box):
     update_icon = Gtk.Template.Child()
     menu_button_store = Gtk.Template.Child()
     menu_button_update = Gtk.Template.Child()
+    menu_button_dlc = Gtk.Template.Child()
+    dlc_image = Gtk.Template.Child()
+    dlc_name = Gtk.Template.Child()
+    dlc_button = Gtk.Template.Child()
+    dlc_button_icon = Gtk.Template.Child()
 
     state = Enum('state',
                  'DOWNLOADABLE INSTALLABLE UPDATABLE QUEUED DOWNLOADING INSTALLING INSTALLED NOTLAUNCHABLE UNINSTALLING'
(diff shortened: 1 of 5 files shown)


sharkwouter:
Thanks a lot for the work. I'll try to find time to formally review this on the weekend. In the meantime, I do see one thing which I would like to see implemented a little bit differently. Could you use json instead of pickle to save which dlc is installed? Pickle has some security issues associated with it.

makson96:
No problem. I have already changed that in latest commits.

minigalaxy/ui/gametile.py
```diff
@@ -233,83 +253,146 @@ def __download_file(self) -> None:
...
+                d_icon = dlc["images"]["sidebarIcon"]
+                d_name = dlc["title"]
+                d_status = self.game.get_dlc_status(d_name)
+                self.dlc_box(d_icon, d_name, d_status, d_installer)
+                if dlc not in self.game.dlcs:
+                    self.game.dlcs.append(dlc)
+        if self.game.dlcs:
+            self.menu_button_dlc.show()
+
+    def dlc_box(self, icon, title, status, installer):
```
sharkwouter, line 361: Please rename this function to describe what it does.

minigalaxy/ui/gametile.py
```diff
@@ -349,7 +432,7 @@ def reload_state(self):
             self.update_to_state(self.state.INSTALLABLE)
         else:
             self.update_to_state(self.state.DOWNLOADABLE)
-        check_upd_thread = threading.Thread(target=self.__check_for_update())
+        check_upd_thread = threading.Thread(target=self.__check_for_update_dlc())
```
sharkwouter, line 435: Not really related to this PR, but renaming this thread to check_update_thread would be better. Feel free to ignore this one, though.

minigalaxy/ui/library.py
```diff
@@ -89,23 +88,18 @@ def __create_gametiles(self) -> None:
         games_with_removed_tiles = []
         for child in self.flowbox.get_children():
             tile = child.get_children()[0]
-            if tile.current_state == tile.state.INSTALLED:
-                if not tile.game.image_url:
-                    games_with_removed_tiles.append(copy.deepcopy(tile.game))
-                    self.flowbox.remove(tile)
```
sharkwouter, line 95: Could you explain why this is not longer needed? It has been a long time since I wrote this code.

minigalaxy/ui/library.py
```diff
@@ -89,23 +88,18 @@ def __create_gametiles(self) -> None:
...
                 games_with_tiles.append(tile.game)
 
         for game in self.games:
-            if game in games_with_tiles:
-                continue
             if game in games_with_removed_tiles:
                 for tile_game in games_with_removed_tiles:
                     if game == tile_game:
                         game.install_dir = tile_game.install_dir
-                        break
```
sharkwouter, line 107: Does this break not do anything?

sharkwouter (changes requested):
Thanks for this PR. It seems to work well. I did make a small change to make sure the DLC button doesn't show on games without DLC. Sorry if that was a bit invasive.

I made some small comments. The code looks quite good overall and it's very close to being mergable I would say. I don't really have much more to add.

One more thing, though. Could you rebase it on master?

minigalaxy/ui/library.py
makson96: OK. So now this 2 if statements are combined in one line (92). This reduces number of indentations and make code more readable. flowbox.remove() was executed here just to execute flowbox.add few lines later in __add_gametile(). Now games_with_removed_tiles (well, when I think about it now, it doesn't seems to be propitiate name any more) do not go to __add_gametile() any more to avoid just simple tile removal and addition. This is part of avoiding double game reinitialization.

minigalaxy/ui/library.py
makson96: Well, I am not sure at the moment, but I don't think there should be more than one hit between self.games and games_with_removed_tiles per ttile. If such a situation will be the case, maybe we should diagnose why is it so instead of hiding it with break statement after first hit.

makson96:
Hi,
I have addressed your review. What is more, inspired by your questions I have simplified __create_gametiles even further. I don't see any sides effects and now code is obvious for what it does.
I have also expanded DLC JSON file to store also info of dlc_installers. This is just ground for future work with DLC update (I can now compare if installer changed), but it is not compatible with previous JSON (please remove any previous JSON files). The reason why I do it right now, is that I don't want to break JSON compatibility after it got merged with master.

minigalaxy/ui/library.py
sharkwouter: I think this break was there because it speeds up the code execution. You don't have to go through the full list if you already found the game.

Issue #41, opened 2020-01-05
Change installed filter to include installing games


sharkwouter:
Currently it is very hard to track which games are installing. This could be a way to make them a little bit more visible.

Steam marks games which are being downloaded as installed as well in your library.

makson96:
Should we close this now?

sharkwouter:
True, good point

PR #184, opened 2020-10-29
Issue 41

makson96:
This is rather straight forward fix for issue #41


sharkwouter:
Thanks for this one! I fixed a really small bug with it and changed the style a little bit to make it a bit more readable, hope you don't mind.

Issue #119, opened 2020-04-06
Temporary extracted files and installers may stay in some conditions


tim77:
Hello. :) I have found one another pretty unpleasant bug. Story: i have changed installation path for installed games which located on HDD. Then i queued for downloading couple of games and was AFK. When i back after some time a have seen that there was not free space on HDD ("Installation path") and whole downloading processes goes wrong. Zero free space on HDD is not the biggest problem there IMO, the problem is that temporary files in `cache/minigalaxy/extract` and many installers in `cache/minigalaxy/download` folder not auto cleaned and ate **tons** of space (on SSD).

For users this could be not obvious and would be nice if Minigalgaxy can handle this somehow. Maybe at least we should have some nice button in GUI to clean such trashy files.

sharkwouter:
Odd, this is a bug actually. Those files should be deleted afterinstallation. Thanks for reporting!

sharkwouter:
This should be resolved in the next release which is coming up this weekend.

PR #185, opened 2020-10-30
Connection errors handle for download manager

makson96:
At the moment if there will be connection issues during game download (and there are, I have encountered them with big installers), download_manager.py will wait for server response forever in this line
```
for chunk in download_request.iter_content(chunk_size=DOWNLOAD_CHUNK_SIZE):
```
What is more, because it is stuck here, loop is not progressing, so neither cancel, nor pause can recover user from this hang (ouch).
I have added 30 second timeout for getting response from the server. If this timeout is reached, function will return "ConnectionError". This error is handled and reconnection try is performed. If 5 such a tries failed, download will cancel exactly like in case of pressing cancel button.

I have also portioned "__download_file" function to smaller ones to make it more readable.

What is more I have improved cleaning .cache directory to hopefully solve #119


sharkwouter:
Thanks for this PR. This was indeed very much needed. What I do wonder, though, is how I can test if this works like intended. What was your method for this?

makson96:
1. Start minigalaxy from the terminal.
2. Start game download.
3. Disconnect your internet.
4. Wait for connection error to printed on terminal.
5. Immediately connect internet back.

sharkwouter:
This is quite an improvement actually, it works. If you leave the connection down for long enough, it will still result in a ReadTimeout exception from requests which isn't caught, though. I think this would be good for another PR.

In case you'd be interesting in working on improving the robustness of the download manager further, the following things will make it crash:
- A game cannot be unzipped.
- A timeout is reached on the connection.

There is probably more, but that is what I've noticed.

PR #193, opened 2020-11-15
Update Russian translation

protheory8:
Would be nice if someone double-checked though.


sharkwouter:
Thanks! It looks good to me

PR #200, opened 2020-11-17
[pt_BR] Update translation

EsdrasTarsis:
I updated and revised the Brazilian Portuguese translation and apparently compiled the other translations when I ran `scripts/compile-translations.sh`


sharkwouter:
Thanks!

PR #205, opened 2020-11-25
Change two dots to ellipsis

TotalCaesar659:


sharkwouter:
Thanks for the PR. I was a bit sceptical, but now I've tried it I think this does look better than before. The only issue is that it has to be consistent with all the translations as well. If you wouldn't mind updating the translations in data/po as well, I would like to merge it, otherwise I'm afraid I'd rather leave it out. Let me know what you decide.

TotalCaesar659:
@sharkwouter I've fixed translations, how to resolve these conflicts?

sharkwouter:
I'll look into this tomorrow. Thanks!

sharkwouter:
Thanks! I've merged this now!

PR #211, opened 2020-11-29
Fix #209

### LINKED ISSUE #209: Show store page option doesn't appear anymore
Not sure what broke this, but this option is no longer showing up unfortunately. It won't stop the release, but it is a bit of a shame.

makson96:
So many bugs one day before the release. I will see what I can do.
This one will fix #209 and add unit test for checking that this won't be broken in the future.


sharkwouter:
Thanks! This works well

PR #212, opened 2020-11-29
Refactor installer together with some serious bug fixes

### LINKED ISSUE #208: Checksum failures crash the download manager
This leaves the user unable to continue.

makson96:
I wanted to fix #208 but together with it I ended up with whole installer.py module refactor. I also added some unit tests for it. I also fixed some bugs which I consider really serious:
- fix for minigalaxy crash on start if there is no "~/GOG Games" directory. This will probably affect all new users.
- fix that uninstalling game will end with checking game update and reporting that game update is available for just uninstalled game.
- offline mode seems to be broken. This will also fix this.

Please note that I have altered a little bit 'wine' games install process to be more in line with linux one, but I don't have wine to check this out. Please check wine games installation and report any error here before merge.

### DIFF
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -18,6 +18,7 @@ class Api:
         self.client_id = "46899977096215655"
         self.client_secret = "9d85c43b1482497dbbce61f6e4aa173a433796eeae2ca8c5f6129f2dc4de46d9"
         self.debug = os.environ.get("MG_DEBUG")
+        self.active_token = False
 
     # use a method to authenticate, based on the information we have
     # Returns an empty string if no information was entered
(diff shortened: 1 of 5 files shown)


sharkwouter:
Sorry for being so late with reviewing this, but it has to wait at least another day unfortunately. Just so you know what is up, even though I don't have the time today. Thanks a lot for this PR, though, from going over it briefly, I think it's a big improvement.

makson96:
Thanks for the early info. Waiting for the full review.

sharkwouter:
Thanks for the work, it's well made and a big improvement. I've finally managed to take a look at it and merge it.

PR #215, opened 2020-11-30
Update pl.po

ArturWroblewski:
Update PL


sharkwouter:
Thanks for the update. Unfortunately I've just merged a PR which was a bit older and causes a conflict. Could you resolve the conflict?

ArturWroblewski:
It should be okay by now. Please check.

sharkwouter:
Thanks! That looks good

Issue #38, opened 2020-01-05
Allow using the system's wine version instead of the bundled one


sharkwouter:
This may improve compatibility and performance, but it also has the potential to break games in unexpected ways, so it might be good to make it configurable.

Currently a gametile will compile a command to execute when starting a game based on which files are in the game's directory. It does already support native scummvm and dosbox installations. Wine games are already detected right now, but they aren't launched in a different way from other games.

The wine prefix is always in a directory called prefix from what I've seen. The location of the game's binary is probably best to retrieve by parsing the start.sh script.

Examples of wine packaged games being sold on GOG:

- Flatout
- Flatout 2

namandixit:
A better alternative (though a bit more complicated to implement) might be to add support for Steam Play's `compatibilitytools.d`. This way, not only Windows games would be playable with Proton/Wine but community-made tools such as [Boxtron](https://github.com/dreamer/boxtron), [luxtorpeda](https://github.com/dreamer/luxtorpeda), [Roberta](https://github.com/dreamer/roberta) could also be used for games that ship with outdated versions of DOSBox, SCUMM, etc.

sharkwouter:
It's true that proton is better than standard wine, but I wouldn't want the client to depend upon Steam. Using the system wide installation is also a bit easier and less likely to break with future releases of wine.

Dosbox and SCUMMVM games launched through Minigalaxy already use the versions installed on the system if available. This was added in 0.9.2.

namandixit:
> It's true that proton is better than standard wine, but I wouldn't want the client to depend upon Steam. 

It won't have to depend upon Steam, [see here](https://github.com/ValveSoftware/steam-for-linux/issues/6310). The software will have to parse the VDF files that describe the tool itself.

> Valve might change how proton is launched in the future, for wine the cli is stable

That is true. One would hope that since third party tools have cropped up that use the Steam Play interface, they will not maintain compatibility, but there are obviously no guarantees.

> Dosbox and SCUMMVM games launched through Minigalaxy already use the versions installed on the system if available. This was added in 0.9.2.

Right, but support for these had to be added manually, while supporting the Steam Play interface will mean that minigalaxy will automatically support any other similar toolsin the future (given they are compatible with Steam Play).

namandixit:
I am documenting what I have figured out in regards to this:

There are two files in a Steam Play compatible tool: `compatibilitytool.vdf` and `toolmanifest.vdf`. 

[...]

PR #220, opened 2020-12-07
Fix 219

### LINKED ISSUE #219: Stonekeep detected as windows game but it's a DOS game
In the Stonekeep folder there is a unins000.exe file so that it gets detected as windows program.

makson96:
This pull request is to fix #219
It also provides unit tests for failed scenario.

### DIFF
--- a/minigalaxy/launcher.py
+++ b/minigalaxy/launcher.py
@@ -51,12 +51,13 @@ def get_execute_command(game) -> list:
 
 def determine_launcher_type(files):
     launcher_type = "unknown"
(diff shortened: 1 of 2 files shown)


sharkwouter:
Thanks for the PR. I am afraid I won't be able to merge this one, though. Dosbox configurations for Windows are specific to Windows and will usually not work on Windows, so making this work will with all Dosbox games will require more work this this.

You may be able to take a look at Boxtron to see how they resolved this issue. The comments in #38 also have some good insights into how using proton/boxtron with Minigalaxy could be implemented.

PR #222, opened 2020-12-09
Add option to keep window maximized

TotalCaesar659:


sharkwouter:
Thanks for the PR, sorry that it took me a while to get back to you. I get why you would want this option, but I've been trying to think if there is a better way. It looks a little bit odd in the settings menu.

Would it be possible to make the window track if it is maximized and set the option based on that? Then a slider in the settings would not be necessary.

TotalCaesar659:
Tried to change but it doesn't save state when closing.

sharkwouter:
I've played around with this quite a bit now, seeing how I can change it to detect when the state of the window is changed, but I haven't been able to find it unfortunately :(

This may just not be possible in Gtk, let me know if you find another solution.

makson96:
Hi,
It can be done like this:
- In class constructor (line 34) you can connect maximization event like this:
```
self.connect('window-state-event', self.on_window_state_event)
```
- and then you can create on_window_state_event() method which will looks as follows:
```
    def on_window_state_event(self, window, event):
        if event.new_window_state & Gdk.WindowState.MAXIMIZED:
            Config.set("keep_window_maximized", True)
        else:
            Config.set("keep_window_maximized", False)
```

sharkwouter:
Great, that works. I'll connect the event itself with Glade then.

sharkwouter:
@TotalCaesar659 thanks for the initial work and the idea. I've been able to get it working and have merged it now.

TotalCaesar659:
@sharkwouter Great, thank you very much!

Issue #207, opened 2020-11-28
Some games will show there is an update even if the latest version is installed


sharkwouter:
An example for this is Bio Menace. I have attached the gameinfo file and server response for that game in this issue.

[server_response_biomenace.txt](https://github.com/sharkwouter/minigalaxy/files/5610410/server_response_biomenace.txt)
[gameinfo_biomenace.txt](https://github.com/sharkwouter/minigalaxy/files/5610407/gameinfo_biomenace.txt)

Other example are:
- Crystal Caves
- Akalabeth: World of Doom
- CAYNE (this one is free)
- Cannon Fodder

I'll update this issue as I find more.

Btw, if you launch minigalaxy with the environment variable MG_DEBUG set. you can see the API requests in the terminal.

makson96:
Yup, game version in gameinfo file is:
2.0.0.2
but API reports:
1.1 (gog-2)

Could you please provide result of:
grep -e gog-2 -r ~/GOG\ Games/Bio\ Menace/
so we could check if API game version is stored somewhere else in game files.

sharkwouter:
Unfortunately, I can't find gog-2 or 1.1 anywhere in the game files with grep, so we have to find a different solution.

I would propose we kill a couple of birds with one stone here and implement the following:

- When a user downloads a game a text file with some json properties is added to the directory the game is in. This could be called something like ``minigalaxy-info.json`` for instance. This would contain the game's name, game id, version, store page url and description as returned by the API when we request the download url.
- When creating tiles, this file is checked first. Meaning we always get all the information we need for a game even in offline mode.
- When updating we use the version in this file and update it.

sharkwouter:
I've added some games to the list.

PR #225, opened 2020-12-19
Fix 207

makson96:
This pull request is to address issue #207

The idea is as follows:
After game installation we will create minigalaxy-info.json file in game directory. This file will store game version (from Api) and DLC versions. It can also be used in future to store more data about Game.
If file is present minigalaxy will use it to check if update is available. If file is not present, minigalaxy will fall back to checking gameinfo file.
As it goes to DLCs. We keep compatibility with previous system. If DLC version cannot be determined from minigalaxy-info.json, we will also check minigalaxy-dlc.json. If DLC is present in both files, only minigalaxy-info.json is taken into account.
dlc_backord

This method is applied on all new games, dlcs installation, but backward compatibility is preserved.

I also hope, that code is now more clean and understandable.

Unit tests are adapted to this change and their range is expanded.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -9,6 +9,13 @@ sys.modules['minigalaxy.constants'] = m_constants
 sys.modules['minigalaxy.config'] = m_config
 from minigalaxy.api import Api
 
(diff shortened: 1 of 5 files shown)


tests/test_api.py
```diff
@@ -9,6 +9,13 @@
 sys.modules['minigalaxy.config'] = m_config
 from minigalaxy.api import Api
 
+API_GET_INFO_TOONSTRUCK = {'downloads': {'installers': [{'id': 'installer_windows_en', 'name': 'Toonstruck', 'os': 'windows', 'language': 'en', 'language_full': 'English', 'version': '1.0', 'total_size': 939524096, 'files': [{'id': 'en1installer0', 'size': 1048576, 'downlink': 'https://api.gog.com/products/1207666633/downlink/installer/en1installer0'}, {'id': 'en1installer1', 'size': 938475520, 'downlink': 'https://api.gog.com/products/1207666633/downlink/installer/en1installer1'}]},
```
sharkwouter, line 12: This is nice, makes the tests a lot cleaner

sharkwouter (approved):
I'm impressed, this is a pretty clean approach to this. Thanks for the contribution!

I even found an edge case which seems to not even cause any issues. Crystal Caves receives null as version from the API and it doesn't cause an issue.

Issue #126, opened 2020-04-08
Exclude mo files from git repository


sharkwouter:
Since they are compiled a bit of a pain to keep updated in the repo, I think excluding them could be a good idea. This means that modifications will need to make sure that the translations are build if the user clones the git repo and runs ``bin/minigalaxy``. A change like that would help translators, though.

For installations with ``setup.py`` the translations are already being compiled.

PR #229, opened 2020-12-22
Major Debian rework

stephanlachnit:
I'm interested in bringing this to Debian (depending on how fast we can merge this maybe even Debian 11), but there are some changes required before I'll do the chore.

The first commit is mostly to make it a bit easier to copy the packaging work for me, no big changes there, except that it targets Ubuntu 20.04 in terms of debhelper and Debian Policy (building on 18.04 may not work, idk).
Maybe noticable is that I removed the `tar-ignore = ".idea"` line, since I don't really get it's point. It really only makes sense for a release source tarball that - if you want generate it via dpkg  - should be done from a clean clone anyway.

In the second commit I moved dosbox and scummvm to Suggests since I feel it belongs there. You don't really need it for native titles, so the user should decide (as as with wine32).

The third commit is required to make this package actually reproducible. The files in `data/mo/` are binary files that get generated during the build and may change in that process, which is a big no-no for my workflow when packaging. Also the `minigalaxy.pot` template shouldn't be part of the repo if it's automatically generated.

Let me know what you think :)


makson96:
This is probably also related to those issues: #7 #126

sharkwouter:
First off, thanks for the work. It looks a bit better. My packaging process was a bit strange so far and it still has some issues.

Initially I wasn't quite sure what to think of just removing all the mo files and the pot file, but there is probably no clean way to get rid of them without having to update my documentation. I'll just merge this and update the documentation.

stephanlachnit:
Cool, do you have a rough ETA for the next release? The Debian 11 soft free on is the 12th February, at which point no new packages are accepted, so if you could a release until end of January that would be super nice :)

sharkwouter:
The plan is to do the next release in 2 weeks from now. That should have mainly bugfixes.

sharkwouter:
@stephanlachnit the 1.0.1 release is out now. Let me know if you need any help upstreaming it into Debian.

stephanlachnit:
Already packaged it and searched for a sponsor. I got someone interested, so I hope it will be uploaded soon.

sharkwouter:
Great, thanks for the update!

Issue #226, opened 2020-12-20
Tries to unzip .sh files (and fails)


Fluffkin:
West of Loathing for an example. Click "update". Update downloads then...

```
Executing integrity check for /home/zzzzz/.cache/minigalaxy/download/West of Loathing/west_of_loathing_1_11_11_11_11_41181.sh
Verifying archive integrity... MD5 checksums are OK. All good.
[/home/zzzzz/.cache/minigalaxy/download/West of Loathing/west_of_loathing_1_11_11_11_11_41181.sh]
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
note:  /home/zzzzz/.cache/minigalaxy/download/West of Loathing/west_of_loathing_1_11_11_11_11_41181.sh may be a plain executable, not an archive
unzip:  cannot find zipfile directory in one of /home/zzzzz/.cache/minigalaxy/download/West of Loathing/west_of_loathing_1_11_11_11_11_41181.sh or
        /home/zzzzz/.cache/minigalaxy/download/West of Loathing/west_of_loathing_1_11_11_11_11_41181.sh.zip, and cannot find /home/zzzzz/.cache/minigalaxy/download/West of Loathing/west_of_loathing_1_11_11_11_11_41181.sh.ZIP, period.
Exception in thread Thread-91:
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 954, in _bootstrap_inner
Failed to install West of Loathing: /home/zzzzz/.cache/minigalaxy/download/West of Loathing/west_of_loathing_1_11_11_11_11_41181.sh could not be unzipped.
    self.run()
  File "/usr/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.9/site-packages/minigalaxy/download.py", line 26, in finish
    self.__finish_func()
  File "/usr/lib/python3.9/site-packages/minigalaxy/ui/gametile.py", line 324, in __update
    install_success = self.__install(update=True)
  File "/usr/lib/python3.9/site-packages/minigalaxy/ui/gametile.py", line 290, in __install
    install_game(self.game, installer, main_window=self.parent.parent)
  File "/usr/lib/python3.9/site-packages/minigalaxy/installer.py", line 34, in install_game
[...]

PR #231, opened 2020-12-24
Choose OS Version with available before download

Odelpasso:
Hello,

I did this implementation to choose between Linux and Windows for a game when it's available.
Each Linux game has a Windows version and sometimes, the Windows version is better than the Linux version.

Example : 
- The Witcher 2 has a bad OpenGL port on Linux and works better with Wine/DXVK with Windows version.
- Graveyard Keeper has a mouse bug issue (never fixed) and not the Windows version. And the performance are better with DXVK too.

I'm not too familiar with Python so maybe there is a better approach. But it works at this moment.

Thanks for your reviews.

### DIFF
--- a/data/ui/os_version.ui
+++ b/data/ui/os_version.ui
@@ -1,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.38.2 -->
+<interface>
+  <requires lib="gtk+" version="3.24"/>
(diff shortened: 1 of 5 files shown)


Odelpasso:
Standby, MiniGalaxy has a bug which disallow to install Windows game, see : 
https://github.com/sharkwouter/minigalaxy/issues/226#issuecomment-751134926

Issue #232, opened 2020-12-25
[Bug]Install Windows game does not work


Odelpasso:
Hello,

Since the commit ffe09924a7b959514663d0e0d8d1b5d62041b315, there is an issue when a user wants to install a Windows games.

**Way to reproduce :** 

1- If the game is download for the 1st time, nothing happens when the command `command = ["WINEPREFIX={}".format(prefix_dir), "wine", installer, "/dir={}".format(temp_dir)]` is launched by the app. The button state pass to INSTALLED from DOWNLOADABLE but the game is never installed.

2- If i move the Windows game setup in "keep_dir", launch MiniGalaxy and click on "Install", i have this error message : 
```
barbiturik@debian-desktop:~/Bureau/minigalaxy$ ./bin/minigalaxy 
Exception in thread Thread-4:
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/home/barbiturik/Bureau/minigalaxy/minigalaxy/ui/gametile.py", line 269, in __install_game
    install_success = self.__install()
  File "/home/barbiturik/Bureau/minigalaxy/minigalaxy/ui/gametile.py", line 288, in __install
    err_msg = install_game(self.game, installer)
  File "/home/barbiturik/Bureau/minigalaxy/minigalaxy/installer.py", line 17, in install_game
    error_message = extract_installer(game, installer, tmp_dir)
  File "/home/barbiturik/Bureau/minigalaxy/minigalaxy/installer.py", line 73, in extract_installer
    process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
  File "/usr/lib/python3.9/subprocess.py", line 947, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.9/subprocess.py", line 1819, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] Aucun fichier ou dossier de ce type: 'WINEPREFIX=/home/barbiturik/GOG Games/Horizon Zero Dawn Complete Edition/prefix'
```

makson96:
This can be probably closed now.

sharkwouter:
Thanks, done!

PR #233, opened 2020-12-25
Fix bug #232

Odelpasso:
Fix error which disallow to install Windows games.
Implemented since commit #ffe0992
Fix bug #232


makson96:
Good catch. I don't have wine so I didn't test if wine games works fine with installer refactor.
Fix looks fine.

PR #234, opened 2020-12-25
Fix 226

makson96:
So, I did some research about #226 
It seams that if I manually corrupt installer it still reports:
```
Verifying archive integrity... MD5 checksums are OK. All good.
```
So it seems that this check is just for show?

Exception from issue is already addressed since:
https://github.com/sharkwouter/minigalaxy/commit/543bd9696f138f754faf540be4bfd8e069436eed
But we ended up in calling MessageDialog in non thread safe way, which crashes minigalaxy.
- This issue is addressed in this pull request
- I also altered installer to make sure, that damaged installer is removed after failed installation
- And finally we need to make sure that game.install_dir is always updated together with game.status_file_path to be sure that minigalaxy-info.json is in the same dir as game.

That said, I still don't know why corrupted installer is downloaded in #226 

In order to check this this pull request:
- Make sure that Beneath a Steel Sky is uninstalled
- Download Beneath a Steel Sky installer manually
- Corrupt installer
- Put it in "~/.cache/minigalaxy/download/Beneath a Steel Sky/"
- Try to install Beneath a Steel Sky

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -19,8 +19,10 @@ def install_game(game, installer):
         error_message = move_and_overwrite(game, tmp_dir, game.install_dir)
     if not error_message:
         error_message = copy_thumbnail(game)
-    error_message2 = remove_installer(installer)
-    error_message = error_message2 if not error_message else error_message
+    if not error_message:
+        error_message = remove_installer(installer)
+    else:
+        remove_installer(installer)
     if error_message:
         print(error_message)
     return error_message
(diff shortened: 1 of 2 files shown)


sharkwouter (changes requested):
Looks good. It's a cleaner solution than what was already there.

makson96:
I have addressed review. Please can you check it again now?

sharkwouter:
Thanks! Looks good

Issue #106, opened 2020-03-23
More game details


metafarion:
Another one for the pile of feature requests: Additional details on each game accessible via the game list interface.  Right now the UI is pretty sparse; each game has a title image and a Download/Play button and that's basically it.  It would be helpful to have actual text titles (as the images are not always super-legible), as well as game descriptions, ratings, etc.  Perhaps that much detail warrants a "Details" view for each game to keep the UI from getting too cluttered.  If this is a ton of work to implement, even a link to each game's page at GOG.com would be useful.  After all, we game hoarders don't necessarily remember what every game in our library is about.

sharkwouter:
Thanks for reporting. I think this is a good idea, but keeping the UI clean as well would be a priority here.

I see two ways this can be implemented:
- Adding a list view, pretty much like you describe. This would require creating a second library class in ``minigalaxy/ui/``, which should probably extend some library base class. That way we can later add more views. This would require some way to reuse the gametile code as well, it's quite the coding challenge.
- Adding a details/properties button to the game specific menu which currently allows you to uninstall, open files and open the support page. This should then open a window which shows additional information.

metafarion:
I like that there's are Store Page and Support links on each game now, but it'd be nicer if it were available on all games, not just the ones that are currently installed.

sharkwouter:
That's a good point, thanks!

makson96:
I see that I can now go to store page of uninstalled game:
![uninstall_store](https://user-images.githubusercontent.com/1833284/113819747-088a0780-977a-11eb-9e61-b7477183a4cd.png)
Is this issue done? If not, what is the objective now?

sharkwouter:
I would say it is not quite done yet. I see this is a request for extra information in the properties window.

sharkwouter:
The properties window has been updated on master to include more information

Issue #181, opened 2020-10-17
[Feature Request] Allow opening the store page of games which aren't installed


jianpan:
First of all, Minigalaxy is great! I really appreciate your hard work.

Just wondering if you can add a feature to open GOG store page for the selected game? It's quite useful to check game information at GOG. Thanks!

jianpan:
The release note mentioned this feature is included in 0.94. I'm using 0.94 from flathub but I couldn't find where is the Game Menu?

![image](https://user-images.githubusercontent.com/1950818/96349365-2739fc00-107d-11eb-9886-1473ad0e064a.png)

sharkwouter:
![store-page](https://user-images.githubusercontent.com/5042659/96363763-7b54e700-1136-11eb-9daa-eefd2837c496.png)
 Here is an example of where you can find the store page link. It only applies to installed games, though.

jianpan:
Got it. Thanks, sharkwouter. It would be nice to show this for uninstalled games as well. Sometimes I need to remind myself what the game is about before I install. :)

sharkwouter:
No problem. Your proposal makes sense. I'll keep this one open and see when I'll get to it. It probably won't make it in the next release, though. Thanks for the report.

breversa:
@sharkwouter : the "Store page" menu entry doesn't seem to appear reliably.

In addition to not showing on non-installed games (planned), it also doesn't appear for many of my games (e.g. Broforce, Hollow Knight, Wonderboy, etc.).

Or if it does, it's not always and I've been unable to establish a pattern so far.

sharkwouter:
Thanks for the report @breversa. Unfortunately a small oversight caused this to stop working in version 1.0.0. The next version will have a fix for it.

sharkwouter:
@breversa the issue you were having should've been resolved with version 1.0.1.

breversa:
Sadly it doesn't work as of 1.0.1 : the settings icon doesn't appear on non-installed games :

![image](https://user-images.githubusercontent.com/2071276/104453560-e6cb1b00-55a4-11eb-946c-d38a92ea271a.png)

sharkwouter:
My bad for not being clearer. This issue cannot be resolved yet, as we haven't worked on it yet, but you mentioned this additional issue:

> it also doesn't appear for many of my games (e.g. Broforce, Hollow Knight, Wonderboy, etc.).

I was wondering if that had been cleared up, since we did try to fix that one. Thanks!

[...]

PR #235, opened 2020-12-26
Allow to download Windows/Linux version if available (Linux games only)

Odelpasso:
Hello,

I did this implementation to choose between Linux and Windows for a game when it's available.
Each Linux game has a Windows version and sometimes, the Windows version is better than the Linux version.

Example :

- The Witcher 2 has a bad OpenGL port on Linux and works better with Wine/DXVK with Windows version.
- Graveyard Keeper has a mouse bug issue (never fixed) and not the Windows version. And the performance are better with DXVK too.

I'm not too familiar with Python so maybe there is a better approach. But it works at this moment.

Thanks for your reviews.

### DIFF
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -145,7 +145,7 @@ class Api:
                 possible_downloads.append(installer)
         if not possible_downloads:
             if operating_system == "linux":
-                return self.get_download_info(game, "windows")
+                return self.get_download_info(game)
             else:
                 raise NoDownloadLinkFound("Error: {} with id {} couldn't be installed".format(game.name, game.id))
 
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -132,7 +132,8 @@ class Api:
         return response
 
     # This returns a unique download url and a link to the checksum of the download
-    def get_download_info(self, game: Game, operating_system="linux", dlc_installers="") -> tuple:
+    def get_download_info(self, game: Game, dlc_installers="") -> tuple:
+        operating_system = Config.get("OS_Version")
         if dlc_installers:
             installers = dlc_installers
         else:
@@ -217,3 +218,11 @@ class Api:
             print("Response body: {}".format(response.text))
             print("")
         return response.json()
+
+    # Get os version version for which game
+    def get_os_game(self,game):
+        response = self.get_info(game)
+        installers = response["downloads"]["installers"]
+        for installer in installers:
+            game_os = installer["os"]
+        return game_os
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -31,7 +31,7 @@ def verify_installer_integrity(game, installer):
     if not os.path.exists(installer):
         error_message = _("{} failed to download.").format(installer)
     if not error_message:
-        if game.platform == "linux":
+        if game.platform == "linux" and Config.get("OS_Version") == "linux":
             try:
                 print("Executing integrity check for {}".format(installer))
                 os.chmod(installer, 0o744)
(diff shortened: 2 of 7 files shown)


Newbytee:
I agree that this is a good idea for a feature, but I'd like to point one thing out:

> The Witcher 2 has a bad OpenGL port on Linux and works better with Wine/DXVK with Windows version.

It's not even a bad port any more. They fixed it after release: https://youtu.be/Ejs9EzAU9DM It's a shame people still spread this misinformation considering that this is actually one of the better early AAA Linux ports. (no, I'm not including those that happened around the 2000s as that was a different era)

It is, however, still missing mod support.

Apologies for the off-topic comment.

minigalaxy/api.py
```diff
@@ -132,7 +132,8 @@ def get_info(self, game: Game) -> dict:
         return response
 
     # This returns a unique download url and a link to the checksum of the download
-    def get_download_info(self, game: Game, operating_system="linux", dlc_installers="") -> tuple:
+    def get_download_info(self, game: Game, dlc_installers="") -> tuple:
```
makson96, line 135: line 148 is:
```
return self.get_download_info(game, "windows")
```
This will break compatibility with that line.

minigalaxy/installer.py
```diff
@@ -69,7 +69,7 @@ def extract_installer(game, installer, temp_dir):
             os.makedirs(prefix_dir, mode=0o755)
 
         # It's possible to set install dir as argument before installation
-        command = ["WINEPREFIX={}".format(prefix_dir), "wine", installer, "/dir={}".format(temp_dir)]
+        command = ["env", "WINEPREFIX={}".format(prefix_dir), "wine", installer, "/dir={}".format(temp_dir)]
```
makson96, line 72: This is in pull request #233 and should be probably removed from this one.

makson96:
This breaks unit tests.
Please go to failing unit tests and change lines:
```
m_config.Config.get.return_value = "en"
```
to something like:
```
m_config.Config.get.side_effect = ["linux", "en"]
```
You can also change in those unit tests:
```
obs = api.get_download_info("Test Game")
```
to something like:
```
test_game = Game("Test Game")
obs = api.get_download_info(test_game)
```
(and add proper import at the top of the file)
```
from minigalaxy.game import Game
```

minigalaxy/installer.py
Odelpasso: God, i forgot to delete it before to push my commit :D

minigalaxy/api.py
Odelpasso: I tried all my Linux games (~ 7) and i never had an problem. Maybe we can just delete "windows" in the return function.

minigalaxy/api.py
```diff
@@ -145,7 +145,7 @@ def get_download_info(self, game: Game, dlc_installers="") -> tuple:
                 possible_downloads.append(installer)
         if not possible_downloads:
             if operating_system == "linux":
-                return self.get_download_info(game, "windows")
+                return self.get_download_info(game)
```
makson96, line 148: Unfortunately this won't do the trick. The point of this line is to get download_info for windows installer if linux one is not available. This change will put us into infinite loop. We can either:
- Resign from this feature if this is not a real use case (not recommended)
- get_download_info will require a little bit of redesign

makson96:
One more unit test to go.

Please also note, that I haven't run this code yet.

makson96:
@Odelpasso in PyCharm you can right click on "tests" directory and click "Run 'Unittests in tests'". To perform unit tests manually before sending changes on the server.
pycharm_unit_tests

sharkwouter:
Thanks for the PR. It would be cool to be able to set this before installing a game, but it should be a bit hidden and not set globally. I've discussed an alternative way with @Odelpasso which would integrate better into our system, but that does make this PR obsolete.

Our idea is the following:

- Add a "Properties" option to the existing per game menu. Picture to show which menu: Screenshot from 2020-12-27 17-22-51
- Show some additional information in this menu about the game
- Allow enabling downloading the Windows game instead.
- Also allow accessing this menu for games which aren't installed yet.

Going at it this way would also allow us to resolve #106 and possibly #181.

PR #237, opened 2020-12-27
Implement DXVK install for Windows games.

Odelpasso:
Hello,

I think it could be a good idea to install DXVK automatically after the installation of a Windows games.
It avoid to do that manually for each games that the user install.

The code works but i don't like it :
- Replace curl command to retrieve the last version of DXVK
- The time.sleep() function used to wait the download finishes.

Thanks for your reviews.

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -1,10 +1,12 @@
 import os
 import shutil
 import subprocess
+import time
 from minigalaxy.translation import _
 from minigalaxy.paths import CACHE_DIR, THUMBNAIL_DIR
 from minigalaxy.config import Config
-
+from minigalaxy.download import Download
+from minigalaxy.download_manager import DownloadManager
 
 def install_game(game, installer):
     error_message = ""
@@ -77,6 +79,11 @@ def extract_installer(game, installer, temp_dir):
     stdout, stderr = process.communicate()
     stdout = stdout.decode("utf-8")
     stderr = stderr.decode("utf-8")
+
+    # I don't find a better place :/
+    if Config.get("OS_Version") == "windows":
+        install_dxvk(game)
+
     if (process.returncode not in [0, 1]) or \
        (process.returncode in [1] and "(attempting to process anyway)" not in stderr):
         error_message = _("The installation of {} failed. Please try again.").format(installer)
@@ -141,3 +148,34 @@ def remove_installer(installer):
 
 def uninstall_game(game):
     shutil.rmtree(game.install_dir, ignore_errors=True)
+
+# For each installation of Windows game, download the latest version fo DXVK and install it.
+def install_dxvk(game):
+    ## Better approach possible with request and regex - Not so familiar :/
+    curl_command = 'curl -s https://github.com/doitsujin/dxvk/releases/latest 2>&1 | grep -Po [0-9]+\.[0-9]+\.[0-9]+'
+    version = subprocess.check_output(curl_command, shell=True)
+    version = version.decode("utf-8").strip()
+
+    prefix_dir = os.path.join(game.install_dir, "prefix")
+    dxvk_dir = os.path.join(CACHE_DIR, "DXVK")
+    dxvk_path = os.path.join(dxvk_dir, 'dxvk-{}.tar.gz'.format(version))
+
+    link = 'https://github.com/doitsujin/dxvk/releases/download/v{}/dxvk-{}.tar.gz'.format(version, version)
+    dxvk_script = os.path.join(dxvk_dir, "dxvk-{}".format(version))
+
+    if not os.path.exists(dxvk_dir):
+        os.makedirs(dxvk_dir, mode=0o755)
+
+    if not os.path.exists(dxvk_script):
+        download = Download(link, dxvk_path)
+        DownloadManager.download_now(download)
+
+        time.sleep(10) # I think there is better approach instead of this ugly thing
+
+        command = ["tar", "-xf", dxvk_path, "-C", dxvk_dir]
+        process = subprocess.Popen(command)
+        process.wait()
+
+    # Install
+    install = ["env", "WINEPREFIX={}".format(prefix_dir), "{}/setup_dxvk.sh".format(dxvk_script), "install"]
+    process = subprocess.Popen(install)


makson96:
I am not a wine user so I don't feel an expert on its configuration.
That said I got two general design question:
- Do we want to always install it together with wine games? I would rather make in an option. Maybe default option, but still an option.
- If we want to expand around wine installation/configuration, maybe we should do it in new separate module, instead of expanding installer?

minigalaxy/installer.py
```diff
@@ -77,6 +79,11 @@ def extract_installer(game, installer, temp_dir):
     stdout, stderr = process.communicate()
     stdout = stdout.decode("utf-8")
     stderr = stderr.decode("utf-8")
+
+    # I don't find a better place :/
+    if Config.get("OS_Version") == "windows":
+        install_dxvk(game)
```
makson96, line 85: Better place would be in install_game() function.

minigalaxy/installer.py
```diff
@@ -141,3 +148,34 @@ def remove_installer(installer):
 
 def uninstall_game(game):
     shutil.rmtree(game.install_dir, ignore_errors=True)
+
+# For each installation of Windows game, download the latest version fo DXVK and install it.
+def install_dxvk(game):
+    ## Better approach possible with request and regex - Not so familiar :/
+    curl_command = 'curl -s https://github.com/doitsujin/dxvk/releases/latest 2>&1 | grep -Po [0-9]+\.[0-9]+\.[0-9]+'
+    version = subprocess.check_output(curl_command, shell=True)
+    version = version.decode("utf-8").strip()
```
makson96, line 157: Don't call curl directly. Use request for that.

minigalaxy/installer.py
```diff
@@ -141,3 +148,34 @@ def remove_installer(installer):
...
+
+    prefix_dir = os.path.join(game.install_dir, "prefix")
+    dxvk_dir = os.path.join(CACHE_DIR, "DXVK")
+    dxvk_path = os.path.join(dxvk_dir, 'dxvk-{}.tar.gz'.format(version))
+
+    link = 'https://github.com/doitsujin/dxvk/releases/download/v{}/dxvk-{}.tar.gz'.format(version, version)
```
makson96, line 163: Format it rather that way:
```
link = 'https://github.com/doitsujin/dxvk/releases/download/v{0}/dxvk-{0}.tar.gz'.format(version)
```

minigalaxy/installer.py
```diff
@@ -141,3 +148,34 @@ def remove_installer(installer):
...
+    dxvk_script = os.path.join(dxvk_dir, "dxvk-{}".format(version))
+
+    if not os.path.exists(dxvk_dir):
+        os.makedirs(dxvk_dir, mode=0o755)
+
+    if not os.path.exists(dxvk_script):
+        download = Download(link, dxvk_path)
+        DownloadManager.download_now(download)
+
+        time.sleep(10) # I think there is better approach instead of this ugly thing
```
makson96, line 173: There should be no time.sleep() here.

minigalaxy/installer.py
```diff
@@ -141,3 +148,34 @@ def remove_installer(installer):
...
+    dxvk_path = os.path.join(dxvk_dir, 'dxvk-{}.tar.gz'.format(version))
+
+    link = 'https://github.com/doitsujin/dxvk/releases/download/v{}/dxvk-{}.tar.gz'.format(version, version)
+    dxvk_script = os.path.join(dxvk_dir, "dxvk-{}".format(version))
+
+    if not os.path.exists(dxvk_dir):
+        os.makedirs(dxvk_dir, mode=0o755)
+
+    if not os.path.exists(dxvk_script):
+        download = Download(link, dxvk_path)
```
makson96, line 170: Archive extraction and installation should  be a finish_func of Download object.

minigalaxy/installer.py
```diff
@@ -141,3 +148,34 @@ def remove_installer(installer):
...
+        DownloadManager.download_now(download)
+
+        time.sleep(10) # I think there is better approach instead of this ugly thing
+
+        command = ["tar", "-xf", dxvk_path, "-C", dxvk_dir]
```
makson96, line 175: Use Python tarfile module instead of calling system tar.

minigalaxy/installer.py
```diff
@@ -141,3 +148,34 @@ def remove_installer(installer):
...
+        process = subprocess.Popen(command)
+        process.wait()
+
+    # Install
+    install = ["env", "WINEPREFIX={}".format(prefix_dir), "{}/setup_dxvk.sh".format(dxvk_script), "install"]
+    process = subprocess.Popen(install)
```
makson96, line 181: We should wait for subprocess to finish.

sharkwouter:
Thanks for the PR. Do people not install dxvk system wide? Or can be not rely on that being the case? This PR will need some work, but I'm mainly concerned about downloading a binary directly, because it may not work on every distribution.

Odelpasso:
> Thanks for the PR. Do people not install dxvk system wide? Or can be not rely on that being the case? This PR will need some work, but I'm mainly concerned about downloading a binary directly, because it may not work on every distribution.

No DXVK is never installed in system wide, always in the wineprefix not in "/usr/lib". And these files work on all Linux system. The script replace d3d9/d3d10/d3d11 dlls located in "game_prefix/drive_c/windows/{syswow64, system32}

sharkwouter:
Oh that is interesting. Does it work with every wine version?

Odelpasso:
Yes, it's independent to Wine. In fact, Wine uses "wined3d" to translate d3d calls to opengl whereas DXVK translate d3d calls to Vulkan.

The script which is used replace d3d9/d3d1* like i said and set these dlls to native in winecfg. So you can use every version of Wine you want.

loathingKernel:
> No DXVK is never installed in system wide, always in the wineprefix not in "/usr/lib". And these files work on all Linux system. The script replace d3d9/d3d10/d3d11 dlls located in "game_prefix/drive_c/windows/{syswow64, system32}

That is not strictly true, DXVK can be installed system-wide in the sense that the DLLs can be made available under `/usr/share/dxvk` and then symlinked into each `WINEPREFIX`. The included `setup_dxvk` script provides that functionality, and many people, myself included use that approach. This approach saves both space and time as any updates to DXVK are made available to all prefixes automatically.

Issue #210, opened 2020-11-29
Loading the library can take a long time


sharkwouter:
For me when loading my library when starting Minigalaxy takes over 20 seconds. This is because python is bad at multithreading and the window freezes completely until it has checked if there are updates for all the games installed. This should be resolved by making a queue or running it on a different thread in the future. Multiprocess may work here.

strycore:
The concurrent.futures module is really good with that kind of stuff. I use that in Lutris to download icons concurrently. just be careful not to trigger UI changes from anything that runs in an outside thread or process. Get the results to the main thread and do UI updates from there.

sharkwouter:
Thanks for the tip! I'll look into it

makson96:
Do you think we can close this now?

sharkwouter:
Indeed, problem solved

PR #238, opened 2020-12-27
Fix_210

makson96:
Personally I was not affected by #210 ,but I noticed that __check_for_update_dlc() was not really called in separate thread.
Could you check if this simple fix can sole this issue?

### DIFF
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -438,7 +438,7 @@ class GameTile(Gtk.Box):
             return
         if self.game.is_installed():
             self.update_to_state(self.state.INSTALLED)
-            check_update_thread = threading.Thread(target=self.__check_for_update_dlc())
+            check_update_thread = threading.Thread(target=self.__check_for_update_dlc)
             check_update_thread.start()
         elif self.get_keep_executable_path():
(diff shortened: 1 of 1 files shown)


sharkwouter:
Good catch

sharkwouter (approved):

sharkwouter (changes requested):
I was a little bit too quick with approving and found 2 potential issues. There are 2 lines in the function being called which can cause a segmentation fault, since you cannot modify Gtk widgets in a separate thread. To get around this you can make the main thread execute the modification when it has time using the ``GLib.idle_add`` function. 

I see the issue in the following lines:

- https://github.com/sharkwouter/minigalaxy/blob/a15165a0cf273cae1417ff435722b49ca2b5e55c/minigalaxy/ui/gametile.py#L323
- https://github.com/sharkwouter/minigalaxy/blob/a15165a0cf273cae1417ff435722b49ca2b5e55c/minigalaxy/ui/gametile.py#L320

makson96:
Hi,
Thank you for the review. I have addressed this issue.
I also added:
"GLib.idle_add" to some calls in __check_for_dlc() as it is called inside of __check_for_update_dlc()
Could you please review it again.

sharkwouter:
That works, thanks!

PR #241, opened 2021-01-02
Add generic get_status() method for Game.

makson96:
This pull request adds generic get_status() method for Game object to use minigalaxy-info.json together with set_status().

### DIFF
--- a/minigalaxy/game.py
+++ b/minigalaxy/game.py
@@ -59,22 +59,13 @@ class Game:
 
     def is_update_available(self, version_from_api, dlc_title="") -> bool:
         update_available = False
-        installed_version = "0"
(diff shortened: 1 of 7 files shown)


sharkwouter:
Thanks for the PR. I think it is useful to have a function like this, but I'm not a big fan of the naming at the moment. Maybe calling it get_info would be better. Then have a get_dlc_info function which uses this function for dlc specific stuff. It's a bit odd for the generic function to have a dlc related option. If you could change that, I'll merge it.

makson96:
Hi,
I have addressed your review. Please check it again.
I have also moved __set_install_dir() from gametile to Game as this is not related to UI.
Best regards,

sharkwouter:
Thank! It looks good now

PR #243, opened 2021-01-04
Properties

Odelpasso:
Hello,

PR for the Properties project :D 
- Implement the Properties window available from menu button on Gametile.
- Move button Store/Settings/Support from Gametile.py to Properties.py.
- Button store available even if the game is not installed.
- New aesthetic for the menu on gametile.
- Implement variables and commands flags, available for Windows and Linux games.
- Properties switch/buttons disabled/enabled depending on the game's state (ex : Settings button will be disabled if the game is not installed or if the game is for Linux).
- Move Show_FPS switch to Preferences to Properties.

### DIFF
--- a/minigalaxy/launcher.py
+++ b/minigalaxy/launcher.py
@@ -70,11 +70,11 @@ def get_exe_cmd_with_var_command(game, exe_cmd):
     command_list = game.get_info("command").split()
     var_list = game.get_info("variable").split()
 
-    if not(var_list) or var_list[0] == "env":
-        exe_cmd = var_list + exe_cmd + command_list
-    else:
-        var_list.insert(0, "env")
-        exe_cmd = var_list + exe_cmd + command_list
+    if var_list:
+        if var_list[0] not in ["env"]:
+            var_list.insert(0, "env")
+
+    exe_cmd = var_list + exe_cmd + command_list
     return exe_cmd
 
 def get_windows_exe_cmd(game, files):
--- a/minigalaxy/launcher.py
+++ b/minigalaxy/launcher.py
@@ -5,7 +5,6 @@ import re
 import json
 import glob
 from minigalaxy.translation import _
-from minigalaxy.config import Config
 
 
 def config_game(game):
@@ -14,6 +13,11 @@ def config_game(game):
     os.environ["WINEPREFIX"] = prefix
     subprocess.Popen(['wine', 'winecfg'])
 
+def regedit_game(game):
+    prefix = os.path.join(game.install_dir, "prefix")
+
+    os.environ["WINEPREFIX"] = prefix
+    subprocess.Popen(['wine', 'regedit'])
 
 def start_game(game):
     error_message = ""
--- a/minigalaxy/launcher.py
+++ b/minigalaxy/launcher.py
@@ -70,6 +70,9 @@ def get_windows_exe_cmd(game, files):
     exe_cmd = [""]
     prefix = os.path.join(game.install_dir, "prefix")
     os.environ["WINEPREFIX"] = prefix
+    command = game.get_info("command").split(" ")
+    var = game.get_info("variable").split(" ")
+    i = 0
 
     # Find game executable file
     for file in files:
@@ -88,6 +91,18 @@ def get_windows_exe_cmd(game, files):
         executables.remove(os.path.join(game.install_dir, "unins000.exe"))
         filename = os.path.splitext(os.path.basename(executables[0]))[0] + '.exe'
         exe_cmd = ["wine", filename]
+
+    # Check if variable environment are set.
+    # Command can be empty
+    if var[0] != "":
+        while i < len(var):
+            if i < len(var):
+                exe_cmd.insert(i, var[i])
+                i = i + 1
+            else:
+                break
+
+    exe_cmd.extend(command)
     return exe_cmd
 
 
@@ -114,8 +129,24 @@ def get_scummvm_exe_cmd(game, files):
 
 
 def get_start_script_exe_cmd(game, files):
-    return [os.path.join(game.install_dir, "start.sh")]
+    command = game.get_info("command").split(" ")
+    var = game.get_info("variable").split(" ")
+    i = 0
 
+    exec_start = [os.path.join(game.install_dir, "start.sh")]
+
+    # Check if variable environment are set.
+    # Command can be empty
+    if var[0] != "":
+        while i < len(var):
+            if i < len(var):
+                exec_start.insert(i, var[i])
+                i = i + 1
+            else:
+                break
+
+    exec_start.extend(command)
+    return exec_start
 
 def get_final_resort_exe_cmd(game, files):
     # This is the final resort, applies to FTL
--- a/minigalaxy/launcher.py
+++ b/minigalaxy/launcher.py
@@ -65,14 +65,17 @@ def determine_launcher_type(files):
         launcher_type = "final_resort"
     return launcher_type
 
+def get_exe_cmd_with_var_command(game, exe_cmd):
+    command_list = game.get_info("command").split()
+    var_list = game.get_info("variable").split()
+    var_list.insert(0, "env") # env must be still in the 1st place
+    exe_cmd = var_list + exe_cmd + command_list
+    return exe_cmd
 
 def get_windows_exe_cmd(game, files):
     exe_cmd = [""]
     prefix = os.path.join(game.install_dir, "prefix")
     os.environ["WINEPREFIX"] = prefix
-    command = game.get_info("command").split(" ")
-    var = game.get_info("variable").split(" ")
-    i = 0
 
     # Find game executable file
     for file in files:
@@ -92,17 +95,7 @@ def get_windows_exe_cmd(game, files):
         filename = os.path.splitext(os.path.basename(executables[0]))[0] + '.exe'
         exe_cmd = ["wine", filename]
 
-    # Check if variable environment are set.
-    # Command can be empty
-    if var[0] != "":
-        while i < len(var):
-            if i < len(var):
-                exe_cmd.insert(i, var[i])
-                i = i + 1
-            else:
-                break
-
-    exe_cmd.extend(command)
+    exe_cmd = get_exe_cmd_with_var_command(game, exe_cmd)
     return exe_cmd
 
 
@@ -129,23 +122,8 @@ def get_scummvm_exe_cmd(game, files):
 
 
 def get_start_script_exe_cmd(game, files):
-    command = game.get_info("command").split(" ")
-    var = game.get_info("variable").split(" ")
-    i = 0
-
     exec_start = [os.path.join(game.install_dir, "start.sh")]
-
-    # Check if variable environment are set.
-    # Command can be empty
-    if var[0] != "":
-        while i < len(var):
-            if i < len(var):
-                exec_start.insert(i, var[i])
-                i = i + 1
-            else:
-                break
-
-    exec_start.extend(command)
+    exec_start = get_exe_cmd_with_var_command(game, exec_start)
     return exec_start
 
 def get_final_resort_exe_cmd(game, files):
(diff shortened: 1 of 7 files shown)


Odelpasso:
Last commit with a small change. Add game's name in the GtkDialog Title :-) 
Capture d’écran du 2021-01-04 17-02-18

makson96:
Hi,
Great work with this property button.
I would like to help with this one. I haven't run the code yet, but would like to give you some proper input tomorrow (so you don't have to wait too long for it). My early thoughts:
- Unit tests shows that we hit some regressions in launcher. I will hopefully show you how to fix it with proper review (tomorrow)
- Wouldn't "winecfg" be more proper name than "Settings"?
- Do we want to move "Show FPS in games" from global setting to per game setting?

sharkwouter:
I've ran it, but not looked into the code yet. It does work well and looks good.

I think renaming the settings button to winecfg would be good idea indeed.

Moving showing fps to per game is nice I think. Maybe renaming it to just "Show fps in game" instead of games would look a bit better, but I like it.

The unit tests do need to succeed.

I'm impressed, though, good work! I'll hopefully find some time to look into the code tomorrow

minigalaxy/launcher.py
```diff
@@ -70,6 +70,9 @@ def get_windows_exe_cmd(game, files):
     exe_cmd = [""]
     prefix = os.path.join(game.install_dir, "prefix")
     os.environ["WINEPREFIX"] = prefix
+    command = game.get_info("command").split(" ")
+    var = game.get_info("variable").split(" ")
+    i = 0
```
makson96, line 75: This can be done easier. You can use split() without any arguments. It will not produce empty strings in the list and will handle things like double spaces and tabs. But be aware that empty list can be result of such an operation. also i=0 is not needed here. For example:
command_list = game.get_info("command").split()
var_list = game.get_info("variable").split()

minigalaxy/launcher.py
```diff
@@ -88,6 +91,18 @@ def get_windows_exe_cmd(game, files):
...
+    # Command can be empty
+    if var[0] != "":
+        while i < len(var):
+            if i < len(var):
+                exe_cmd.insert(i, var[i])
+                i = i + 1
+            else:
+                break
+
+    exe_cmd.extend(command)
```
makson96, line 105: This can be done easier. Be aware that index 0 might not be present in var (var_list), so this first "if" is risky.
exe_cmd = var_list + exe_cmd + command_list

minigalaxy/launcher.py
```diff
@@ -114,8 +129,24 @@ def get_scummvm_exe_cmd(game, files):
...
+    if var[0] != "":
+        while i < len(var):
+            if i < len(var):
+                exec_start.insert(i, var[i])
+                i = i + 1
+            else:
+                break
 
+    exec_start.extend(command)
+    return exec_start
```
makson96, line 149: As above this can be done easier:
command_list = game.get_info("command").split()
var_list = game.get_info("variable").split()
exec_start = var_list + [os.path.join(game.install_dir, "start.sh")] + command_list
return exec_start

minigalaxy/ui/properties.py
```diff
@@ -0,0 +1,96 @@
...
+            webbrowser.open(self.api.get_info(self.game)['links']['support'], new=2)
+        except:
+            self.parent.parent.show_error(
+                _("Couldn't open support page"),
+                _("Please check your internet connection")
+            )
+
+    @Gtk.Template.Callback("on_button_properties_store_clicked")
+    def on_menu_button_store(self, widget):
+        webbrowser.open(self.gogBaseUrl + self.game.url)
```
makson96, line 80: we should probably put it into "try:" statement like in on_menu_button_support()

minigalaxy/ui/properties.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @Gtk.Template.Callback("on_button_properties_cancel_clicked")
+    def cancel_pressed(self, button):
+        self.destroy()
+
+    @Gtk.Template.Callback("on_button_properties_ok_clicked")
+    def ok_pressed(self, button):
+        self.game.set_info("variable", str(self.entry_properties_variable.get_text()))
+        self.game.set_info("command", str(self.entry_properties_command.get_text()))
+        self.game.set_info("show_fps", self.switch_properties_show_fps.get_active())
```
makson96, line 56: please make sure all self.game.set_info() is under:
if game.is_installed():

makson96:
Hi,
I don't want to review Glade project files as I am not the expert on this but some things look weird there like:
"
              <placeholder/>
            </child>
            <child>
              <placeholder/>
            </child>
            <child>
              <placeholder/>
            </child>
            <child>
              <placeholder/>
"

As it goes to Python code it is in good quality. I have put some notes what can be corrected. This will probably also fix unit tests.

I would reorganize launcher a little bit. We are adding var and command more than one time, so this should be separate function. There is also no reason why we should exclude scummvm and dosbox from for example var:
add function like:
"
def get_exe_cmd_with_var_command(exe_cmd):
    command_list = game.get_info("command").split()
    var_list = game.get_info("variable").split()
    exe_cmd = var_list + exe_cmd + command_list
    return exe_cmd
"
and as a final step in get_execute_command() do:
"
exe_cmd = get_exe_cmd_with_var_command(exe_cmd)
"
And do not modify get_windows_exe_cmd() and get_start_script_exe_cmd().

What is more there should be "env" in variables (not sure if just as a first argument or before every variable). And user is not informed that he/she needs to provide "env" to variable field in Properties.
I think that we need to make sure that during var_list preparation "env" is also put in proper places.
Best regards,

minigalaxy/launcher.py
Odelpasso: Oh god, that's more easiest you're right. Thank you Master, still more things to learn with Python :-)
Change done.

minigalaxy/launcher.py
Odelpasso: Change done, TY

minigalaxy/launcher.py
Odelpasso: Same here :D

minigalaxy/ui/properties.py
Odelpasso: What do you with that :
`        try:
            self.game.set_install_dir()
            subprocess.call(["xdg-open", self.game.install_dir])
        except:
            self.parent.parent.show_error(
                _("Couldn't open game's directory"),
                _("Please check if the game is installed")
            )`

minigalaxy/ui/properties.py
Odelpasso: Do you think it's necessary because when the game is not installed, variable/command text bar and show_fps switch are disabled for the user so he can't use them.

makson96: Yep. Clicking "OK" will call self.game.set_info() (even if variable is empty). This will probably create minigalaxy-info.json file in current working directory.

You can alternatively alter game.set_info() method itself to be executed only on installed games.

minigalaxy/launcher.py
```diff
@@ -65,6 +65,12 @@ def determine_launcher_type(files):
         launcher_type = "final_resort"
     return launcher_type
 
+def get_exe_cmd_with_var_command(game, exe_cmd):
+    command_list = game.get_info("command").split()
+    var_list = game.get_info("variable").split()
+    var_list.insert(0, "env") # env must be still in the 1st place
```
makson96, line 71: only if var_list is not empty.

Do you know if "env" needs to placed before every variable or only once at the beginning?
We should also check if user didn't put "env" here already and don't add "env" if it is already present.

minigalaxy/launcher.py
```diff
@@ -88,6 +94,8 @@ def get_windows_exe_cmd(game, files):
         executables.remove(os.path.join(game.install_dir, "unins000.exe"))
         filename = os.path.splitext(os.path.basename(executables[0]))[0] + '.exe'
         exe_cmd = ["wine", filename]
+
+    exe_cmd = get_exe_cmd_with_var_command(game, exe_cmd)
```
makson96, line 98: Please move it from here to the last step in "get_execute_command()"

minigalaxy/launcher.py
```diff
@@ -114,8 +122,9 @@ def get_scummvm_exe_cmd(game, files):
 
 
 def get_start_script_exe_cmd(game, files):
-    return [os.path.join(game.install_dir, "start.sh")]
-
+    exec_start = [os.path.join(game.install_dir, "start.sh")]
+    exec_start = get_exe_cmd_with_var_command(game, exec_start)
```
makson96, line 126: Please move it from here to the last step in "get_execute_command()"

minigalaxy/ui/properties.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @Gtk.Template.Callback("on_button_properties_settings_clicked")
+    def on_menu_button_settings(self, widget):
+        config_game(self.game)
+
+    @Gtk.Template.Callback("on_button_properties_open_files_clicked")
+    def on_menu_button_open_files(self, widget):
+        try:
+            self.game.set_install_dir()
+            subprocess.call(["xdg-open", self.game.install_dir])
```
makson96, line 68: subprocess.call() will not produce exception here. You can remove this "try:". I was referring to on_menu_button_store(), not on_menu_button_open_files().

minigalaxy/launcher.py
Odelpasso: env needs to be placed once at the beginning (env VAR1 VAR2 VAR3 etc....)

minigalaxy/launcher.py
```diff
@@ -65,6 +66,16 @@ def determine_launcher_type(files):
...
 
+def get_exe_cmd_with_var_command(game, exe_cmd):
+    command_list = game.get_info("command").split()
+    var_list = game.get_info("variable").split()
+
+    if not(var_list) or var_list[0] == "env":
+        exe_cmd = var_list + exe_cmd + command_list
+    else:
+        var_list.insert(0, "env")
+        exe_cmd = var_list + exe_cmd + command_list
```
makson96, line 77: This is working solution, but we have same command
```
exe_cmd = var_list + exe_cmd + command_list
```
in both "if" and "else", which is not a good practice. I would rather suggest something like:
```
if var_list:
    if var_list[0] not in ["env"]:
        var_list.insert(0, "env")
exe_cmd = var_list + exe_cmd + command_list
```

Odelpasso: I don't yet have the good habit of simplifying the code.
The code works, i push later the modification.
Thanks

makson96:
OK.
To sum up:
- I haven't run it yet.
- I haven't reviewed GLADE files.
- Issue with renaming "Settings" -> "winecfg", "Show FPS in games" -> "Show FPS in game" remains.
But:
- Unit tests work fine now.
- Python code looks good and clean to me.

Good job.

makson96:
I have no idea why this failed. I am unable to reproduce this fail locally.
Can we re-run it and check if this was some kind of sporadic Python crash?
The other reason which could have some impact on this fail is removal of Config import from launcher. That would be second thing to check.

sharkwouter:
Restarting the tests made them succeed. That is a bit worrisome. Probably unrelated to this PR

Odelpasso:
Do i need to move this def from gametile.py to properties.py since Store button is in Properties windows ?
https://github.com/sharkwouter/minigalaxy/blob/2972451351ea307c6420a3a1c33e3c4151a12e2c/minigalaxy/ui/gametile.py#L436

makson96:
You can remove this function and thread which calls it.
By the time user clicks "Properties", all games should have filled game.url data.

sharkwouter:
I found one issue. It seems that the properties window does not load the game's image at the moment.

sharkwouter:
Thanks for all the work @Odelpasso and @makson96. I made some small changes which fixed some bugs I've noticed and I've merged it.

PR #247, opened 2021-01-05
Fix case register of program name for uniformity

TotalCaesar659:


sharkwouter:
Thanks! I hadn't noticed this

PR #248, opened 2021-01-05
Don't remove extract dir upon creating a new directory

### LINKED ISSUE #244: 2 games installing at the same time causes 1 to fail
Cleaning up the temporary files is currently too aggressive. The files for the other installation will be removed before the installation is finished, making it fail, if 2 games install at the same time. This does happen sometimes if you download a big game first and then a small game. Downloading continues while a game is installing, so if the next installer is downloaded very fast you see this problem.

sharkwouter:
This should fix #244, but it's a bit hard to make sure it does.


sharkwouter:
Thanks for the feedback, that was a much better idea

Issue #245, opened 2021-01-05
Kept installers are not used for installing


sharkwouter:
Currently if the keep installers option is enabled, the saved installers will not be used for installation when a game is installed after having been deleted. Moving the installer to the keep installers directory will also fail, since the file already exists, taking up extra disk space.

PR #249, opened 2021-01-05
Simplify keeping installers

sharkwouter:
This fixes issue #245 and it simplifies the removal process of
installers. It does, however, change the location in which the
installers are stored, which some people will not like.

To summarize this PR:

- Makes remove_installer always remove all installer files.
- Makes sure kept installers are never redownloaded.
- Changes the location of kept installers to ``~/.cache./minigalaxy/download``.
- Simplifies the code for keeping the installer after installation.

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -19,10 +19,10 @@ def install_game(game, installer):
         error_message = move_and_overwrite(game, tmp_dir, game.install_dir)
     if not error_message:
         error_message = copy_thumbnail(game)
-
-    if not Config.get("keep_installers"):
+    if not error_message:
+        error_message = remove_installer(installer)
+    else:
         remove_installer(installer)
-
     if error_message:
         print(error_message)
     return error_message
@@ -122,8 +122,14 @@ def copy_thumbnail(game):
 
 
 def remove_installer(installer):
-    directory = os.path.dirname(installer)
-    shutil.rmtree(directory, ignore_errors=True)
+    error_message = ""
+    if not Config.get("keep_installers"):
+        installer_directory = os.path.dirname(installer)
+        if os.path.isdir(installer_directory):
+            shutil.rmtree(installer_directory, ignore_errors=True)
+        else:
+            error_message = "No installer directory is present: {}".format(installer_directory)
+    return error_message
 
 
 def uninstall_game(game):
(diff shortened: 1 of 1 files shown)


minigalaxy/installer.py
```diff
@@ -19,10 +19,10 @@ def install_game(game, installer):
         error_message = move_and_overwrite(game, tmp_dir, game.install_dir)
     if not error_message:
         error_message = copy_thumbnail(game)
-    if not error_message:
-        error_message = remove_installer(installer)
-    else:
+
+    if not Config.get("keep_installers"):
```
makson96, line 23: I don't like it. Convention:
```
if not error_message:
    error_message = action()
```
Allowed us to obtain error_message from action and report it for user in error message window. We are now cutting of that possibility from remove_installer(). Checking Config.get("keep_installer") can be done inside of remove_installer().

minigalaxy/installer.py
```diff
@@ -122,21 +122,8 @@ def copy_thumbnail(game):
...
-            # It's needed for multiple files
-            for file in os.listdir(download_dir):
-                shutil.move(download_dir + '/' + file, keep_dir + '/' + file)
-        except Exception as ex:
-            print("Encountered error while copying {} to {}. Got error: {}".format(installer, keep_dir, ex))
-    elif os.path.exists(installer):
-        os.remove(installer)
-    return error_message
+    directory = os.path.dirname(installer)
+    shutil.rmtree(directory, ignore_errors=True)
```
makson96, line 126: I would go rather with:
```
error_message = ""
if not Config.get("keep_installers"):
    installer_directory = os.path.dirname(installer)
    if os.path.isdir(installer_directory):
        shutil.rmtree(installer_directory, ignore_errors=True)
    else:
        error_message = "No installer directory is present: {}".format(installer_directory)
return error_message
```

minigalaxy/installer.py
sharkwouter: Perhaps, but we do need to make sure the installer is deleted if the verification fails. Do you think it would be better for verify_installer_integrity to delete the installer by itself when there is an issue?

sharkwouter: I just re-read the old code and I understand now. I'll restore it and update remove_installer

sharkwouter:
Okay, it should be good now.

makson96:
Code looks good to me.
Some tests should probably be run to make sure that shutil.rmtree() will not ever remove any important directory.

Issue #251, opened 2021-01-08
DLC icon missing


stephanlachnit:
The icon on the DLC button is missing. I haven't tried to search for it in the code, but it seems like you use an icon which is not in the default icon set. If you give the code line of where you create the button in Gtk I can take a look as well.

![Screenshot from 2021-01-08 15-25-02](https://user-images.githubusercontent.com/36662302/104026378-1d302100-51c6-11eb-9155-a1805f48ccd9.png)

sharkwouter:
Oh that could be, the code for that can be found here:

https://github.com/sharkwouter/minigalaxy/blob/1dcc82b4f9e2e00a5d58bc4116b5348b75e3df32/minigalaxy/ui/gametile.py#L367-L397

The icons used are gtk-cdrom, gtk-okay and gtk-goto-bottom. I'm not sure which one is not in the standard set.

stephanlachnit:
Hm, I think the issue is that the `gtk-xxx` icons have been replaced with the Adwaita icon theme.

makson96:
Hmm, can we perform some check if gtk-xxx icon is available and if not use adwaita-xxx instead?

sharkwouter:
It seems in Ubuntu these icons are in the humanity-icon-theme package, but Debian doesn't have these.

sharkwouter:
It seems the gnome-human-icon-theme has some of the icons we use, but not all of them it seems.

stephanlachnit:
Any reason you don't use icons from Adwaita? That's the default Gtk icon theme these days.

Some quick replacing I found:
`gtk-cdom` - `media-optical-symbolic`
`gtk-ok` - `emblem-ok-symbolic`
`gtk-goto-bottom` - `document-save-symbolic`

makson96:
I can replace them with Adwaita if @sharkwouter is ok with this.
I would rather use accordingly:
media-optical
emblem-default
go-bottom
Minigalaxy is colourful no need to use "symbolic".
Is Adwaita available in all major distributions?

stephanlachnit:
Adwaita is in every distro: https://repology.org/project/adwaita-icon-theme/versions

Regarding symbolic: I don't think Adwaita has these icons in non-symbolic. These days UI icons are always symbolic, so that shell themes can work properly.

makson96:
I see non-symbolic (all icons I mentioned) here:
https://packages.debian.org/buster/all/adwaita-icon-theme/filelist

[...]

PR #252, opened 2021-01-08
Change icons from default Gtk icon theme to Adwaita symbolic.

makson96:
On my system:
Gtk icon theme (current):
icons_before
Adwaita full color:
icons_non_symbolic
Adwaita symbolic:
icons_symbolic

As pointed in #251 Debian does not come with Gtk icon theme, so our best solution is to switch to Adwaita.
I consider this pull request as important one as mentioned issue is blocker from our side to be incorporated into Debian repository (and as a result to Ubuntu, Mint, etc.).
I consider this pull request as having little regression risk.

As it goes to icons themselves I consider full color to be prettier, but was prompted that they are going to be legacy so symbolic is more future proof choice.
I am also a little bit confused about using symbolic icon. If I use icon-name-symbolic I got full color instead. To get symbolic I need to use icon-name-symbolic.symbolic. Is this a right approach?

Please let me know if you consider that some icons could be better chosen.
Best regards,

### DIFF
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -368,7 +368,7 @@ class GameTile(Gtk.Box):
         if title not in self.dlc_dict:
             dlc_box = Gtk.Box(orientation=Gtk.Orientation.HORIZONTAL)
             image = Gtk.Image()
-            image.set_from_icon_name("gtk-cdrom", 1)
+            image.set_from_icon_name("media-optical-symbolic.symbolic", 1)
             dlc_box.pack_start(image, False, True, 0)
             label = Gtk.Label(label=title, xalign=0)
             dlc_box.pack_start(label, True, True, 0)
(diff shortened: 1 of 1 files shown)


stephanlachnit:
> As it goes to icons themselves I consider full color to be prettier, but was prompted that they are going to be legacy so symbolic is more future proof choice.
> I am also a little bit confused about using symbolic icon. If I use icon-name-symbolic I got full color instead. To get symbolic I need to use icon-name-symbolic.symbolic. Is this a right approach?

That is indeed interesting. I think there was some stuff about symbolic flags in the docs, I will have a look later.

Iirc there was also a way to specify the normal (non-symbolic) name, and Gtk can then pick the symbolic icon as fallback. This maybe what we are looking for as this solves the legacy problem.

stephanlachnit:
> Iirc there was also a way to specify the normal (non-symbolic) name, and Gtk can then pick the symbolic icon as fallback. This maybe what we are looking for as this solves the legacy problem.

Nvm, this only makes stuff more compliacted.

PR #256, opened 2021-01-09
Icons part 2

makson96:
Lately we have replaced some Gtk icons with Adwaita symbolic ones. However Cancel icon remain Gtk icon, so we end up with mixed icons.
I am not also very happy with Adwaita icons symbolic as I don't think that they match Minigalaxy visual theme.
I am opening this pull request as some sort of opinion exchange platform what to do next with this issue. Please do not merge it too quickly.
In this pull request I have incorporated Humanity icons into Minigalaxy files and use those instead of system ones.
The result is:
icons_humanity1
icons_humanity2

We should probably find an answer to the following questions:
- Do we want to bring icons by ourself instead of using system ones?
- Are there any icons missing in this pull request?
- Do we want to switch some of those icons to others?
- Do we want to change present png icons to svg for better quality (Wine, Update)

P.S. Humanity icons are on GPL 2:
https://launchpad.net/humanity
if we use those we should probably mention it somewhere in the files.

### DIFF
--- a/data/images/dialog-apply.svg
+++ b/data/images/dialog-apply.svg
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+<svg id="svg2841" width="22" height="22" version="1.0" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <defs id="defs2843">
+  <radialGradient id="radialGradient4811" cx="11.25" cy="19.031" r="8.0625" gradientTransform="matrix(1 0 0 .28295 0 13.646)" gradientUnits="userSpaceOnUse">
+   <stop id="stop2487" style="stop-color:#0d0d0d" offset="0"/>
+   <stop id="stop2489" style="stop-color:#0d0d0d;stop-opacity:0" offset="1"/>
+  </radialGradient>
+  <linearGradient id="linearGradient4813" x1="-86.129" x2="-93.088" y1="2.0691" y2="31.138" gradientTransform="matrix(.51128 -.12167 .12241 .50821 56.783 -3.2077)" gradientUnits="userSpaceOnUse">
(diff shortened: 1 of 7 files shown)


stephanlachnit:
> I am not also very happy with Adwaita icons symbolic as I don't think that they match Minigalaxy visual theme.

The entire idea of Icon Themes (https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/IconTheme.html), is that they can be set by the user system wide. If you don't like Adwaita, you can set your default Icon Theme to Humanity or any other.

What I'm against is shipping icons, since they are duplicate files and need to be mentioned in the copyright.

stephanlachnit:
To showcase this a bit more, I created a small application to showcase how the icons look on different icon themes. If you want to test it yourself, here's the code: https://gist.github.com/stephanlachnit/7ff19f022324ed546798c425f291c1fa

I didn't use symbolic, Adwaita falls back to symbolic for `go-bottom` and `emblem-ok` anyway, only `media-optical` has a non-symbolic variant.

Adwaita:
adwaita
Elementary:
elementary
Oxygen:
oxygen
Tango:
tango
Humanity isn't packaged for Debian, so I couldn't test it.

The only problem really is `emblem-ok`. It's not available on most icon themes. Humanity has `emblem-OK`, but I don't know if that will be loaded. The docs [mention](https://developer.gnome.org/gtk3/stable/gtk3-Stock-Items.html#GTK-STOCK-OK:CAPS) mentions using mnemonics, I'll see if I can get that to work (earlier attempts resulted in text instead of icons for me, but maybe I'll find an option).

makson96:
Thanks for all your input.
I am really not an icon expert. I will wait for a while till all interested people will have chance to express their opinion and explore this approach if it is the right one.
Best regards,

sharkwouter:
I think @stephanlachnit makes a good argument for not including copies of icons in Minigalaxy and a new PR has been opened which offers an alternative way. I'd suggest we move this discussion to this PR at #258. I'll close this one for now.

PR #257, opened 2021-01-10
Don't crash with broken refresh_token in config.json

makson96:
On discord such an issue was reported:
Bildschirmfoto_vom_2021-01-09_16-51-47
This pull request is to fix it.

How to test:
1. open ~/.config/minigalaxy/config.json as regular text file and remove half of "refresh_token" value.
2. Start Minigalaxy. Crash from bug report will be obtained
3. Start Minigalaxy from this pull request. No error will be received and Minigalaxy will just ask for another login.

### DIFF
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -42,11 +42,14 @@ class Api:
         response = SESSION.get(request_url, params=params)
 
         response_params = response.json()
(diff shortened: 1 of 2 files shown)


sharkwouter:
Works well, thanks!

PR #258, opened 2021-01-11
Icon rework

stephanlachnit:
This PR does two things:
- use normal icon names and icon sizes
- streamlines the updates available icon to be used from the system

As already mentioned in #256, I created a [gist](https://gist.github.com/stephanlachnit/7ff19f022324ed546798c425f291c1fa) to showcase the icons. Here are some examples with various shell and icon themes:

Tango (shell: Adwaita):
tango
Ukui classic (shell: Ukui):
ukui-classical
Adwaita (sell: Adwaita):
adwaita
ePapirus (sehll: Ukui):
epapirus
Moka (shell: Adwaita):
moka
Paper (shell: Ukui dark):
paper
Papirus (shell: Ukui):
papirus

~Btw I think `document-save` is better suited than `go-bottom` themes, for example on the Paper and Papirus themes.~


stephanlachnit:
I changed `emblem-ok` to `object-select`, which I found looked better on some themes. @makson96 can you try the gist and check the looks on humanity?

sharkwouter:
This looks pretty good, thanks for the clear example. I Think ``go-bottom`` is a bit ugly on Ubuntu to be honest, using ``document-save`` may be better indeed.

Here is what I see:

Screenshot from 2021-01-12 09-57-58

This is with ``document-save``:

Screenshot from 2021-01-12 09-57-28

sharkwouter:
On Ubuntu 20.04 ``emblem-ok`` and ``object-select`` are actually the same icon.

minigalaxy/ui/gametile.py
```diff
@@ -381,19 +380,16 @@ def update_gtk_box_for_dlc(self, icon, title, installer):
             self.get_async_image_dlc_icon(icon, title)
         download_info = self.api.get_download_info(self.game, dlc_installers=installer)
         if self.game.is_installed(dlc_title=title):
-            icon_name = "emblem-default-symbolic.symbolic"
+            icon_name = "object-select"
             self.dlc_dict[title][0].set_sensitive(False)
         elif self.game.is_update_available(version_from_api=download_info["version"], dlc_title=title):
```
makson96, line 385: I know this is not strictly icon thing, but can you switch places between this elif and its content and "if self.game.is_installed(dlc_title=title):"

Now if "if self.game.is_installed(dlc_title=title):" is true (game is installed), this elif is never checked, so update icon is never shown on DLC.

makson96:
@stephanlachnit thank you for your work with this icons.
I see some unintentional change with this code. Now update icon is shown on all gametiles. I don't know really why:
update_all

On my system icons look like this:
icons-p

minigalaxy/ui/gametile.py
sharkwouter: This issue is now being addressed in #263

sharkwouter:
@stephanlachnit do you need help with the issue mentioned by Makson?

I'd like to merge this when the following changes have been made:

- The upload icon only shows when needed.
- Use the ``document-save`` icon instead of ``go-bottom``

This is currently the last issue still open in the 1.0.2 milestone.

stephanlachnit:
I'll try to fix it tomorrow.

sharkwouter:
Awesome, thanks!

stephanlachnit:
Works now.

sharkwouter:
Works well, thanks for the contribution!

PR #259, opened 2021-01-11
add DLC list spacing

stephanlachnit:
Text and icon / install button were squeezed together, which looked kinda ugly. Adds standard 8px spacing.


sharkwouter:
Looks good, thanks!

PR #260, opened 2021-01-11
Add missing dependencies

stephanlachnit:
Just noticed that these should be stated separately. It's doesn't really make a difference in practice since Gtk 3.0 depends on both of them, but it's always good to state direct dependencies in case stuff changes in the future.


sharkwouter:
Makes sense. Thanks for the contribution

PR #263, opened 2021-01-12
Actually perform DLC update check.

makson96:
Well,
It seems that DLC update check was not perform, because if DLC was installed it never executed "elif" with update check.
This pull request fix that.

How to test it:
1. Install game with DLC (optionally two DLCs).
2. Exit Minigalaxy.
3. Open minigalaxy-info.json from game files and change version of one of DLCs.
4. Star Minigalaxy - all DLCs still are in installed state. Exit Minigalaxy.
5. Apply this patch and start Minigalaxy again - DLC with altered version will have update available.


sharkwouter:
Thanks! I'll merge this after the icon rework PR.

stephanlachnit:
@sharkwouter can you merge this before the icon PR?

sharkwouter:
Done, it looks good and it's only a small change

PR #268, opened 2021-01-13
Wait for game.url to be ready and then show Store button.

### LINKED ISSUE #265: Opening the store page only works for games which are just installed
In the following code the store page button is hidden when a game tile is created:

https://github.com/sharkwouter/minigalaxy/blob/333774b7ff691450fd0dc29e8e5254162f4b7ec1/minigalaxy/ui/gametile.py#L86-L87

From what I can see, there is no code which unhides it if the store page url is added later. This means the store page button is never shown on already installed games.

makson96:
This one is to fix #265 
Unfortunately it doesn't seem that some easy one-liner will solve this issue.
First gametile with installed game is created and some time after it game data is updated from api, what includes setting game.url.
Because upon creating gametile there is no game.url, Open Store button is always hidden.

This pull request will make gametile to monitor in separate thread for 10 seconds if game.url is set. If so Open Store button will be shown.

### DIFF
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -437,6 +437,7 @@ class GameTile(Gtk.Box):
         while performed_try < tries:
             if self.game.url:
                 GLib.idle_add(self.menu_button_store.show)
+                break
             performed_try += 1
             time.sleep(1)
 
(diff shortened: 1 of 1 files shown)


sharkwouter:
Working well, thanks!

PR #269, opened 2021-01-13
Fix 266

### LINKED ISSUE #266: Missing thumbnails are never downloaded for already installed games
When a games has been installed with the installer manually, Minigalaxy will currently never download the image.

makson96:
This one is to fix #266 
It is not easy either. game.image_url appear some time after gametile is created. So in case of lack of thumbnail, we will monitor for 10 seconds if game.image_url appear and if so download it.

Test:
1. Remove thumbnail.jpg from game and ~/.cache/minigalaxy/thumbnails directories.
2. Start Minigalaxy.

### DIFF
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -7,6 +7,7 @@ import webbrowser
 import threading
 import subprocess
 import re
+import time
 import urllib.parse
 from gi.repository.GdkPixbuf import Pixbuf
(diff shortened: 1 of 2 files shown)


sharkwouter:
Cool, I think I like this solution better than the one I was thinking of. Mine would've involved pressing the refresh button

sharkwouter:
Works well, merging

Issue #157, opened 2020-05-18
Crash after entering credentials


lotth:
Minigalaxy crashes every time after I enter my credentials (email, password and, first time, two-factor 4 gidit code).
minigalaxy version 0.9.4
OS : Fedora 32 x64

Traceback (most recent call last):
  File "/usr/bin/minigalaxy", line 66, in <module>
    main()
  File "/usr/bin/minigalaxy", line 60, in main
    window = Window(APPLICATION_NAME)
  File "/usr/lib/python3.8/site-packages/minigalaxy/ui/window.py", line 51, in __init__
    self.__authenticate()
  File "/usr/lib/python3.8/site-packages/minigalaxy/ui/window.py", line 128, in __authenticate
    authenticated = self.api.authenticate(refresh_token=token, login_code=result)
  File "/usr/lib/python3.8/site-packages/minigalaxy/api.py", line 28, in authenticate
    return self.__get_token(login_code)
  File "/usr/lib/python3.8/site-packages/minigalaxy/api.py", line 63, in __get_token
    self.active_token = response_params['access_token']
KeyError: 'access_token'

sharkwouter:
Try launching with ``MG_DEBUG=1 minigalaxy`` then you can see what the server responds. Do not post the full log with this enabled until after you've removed you passwords and tokens from it, though.

lotth:
Same trace with MG_DEBUG=1. Could it be that Fedora did not compile minigalaxy with this option?

sharkwouter:
Maybe I did not actually make auth priny what is it doing with MG_DEBUG. Could you try starting as ``minigalaxy --reset`` once. It might be an issue with the configuration file.

lotth:
Same effect.

sharkwouter:
Shame, I'll need to look into this some more to see if I can reproduce this then. I'll probably find some time for this tonight, otherwise later this week.

sharkwouter:
Just to make sure, you don't have a firewall configuration which would block this, right?

lotth:
No, I don't. Initial connection with the server seems to work as I receive requests for two-factor code, and associated email is sent correctly.
I also tried with and without VPN, with no success.

sharkwouter:
Okay, thanks. It's probably an issue with Minigalaxy then.

sharkwouter:
I have not been able to reproduce this on the current master branch. I don't think the code for logging in has changed, though.

Could you try cloning the repo and running it?

[...]

PR #275, opened 2021-01-16
fix_157

makson96:
This pull request is to address #157 
I am not sure if user with this error could now log in, but at least Minigalaxy will not crash in this case.
This is also achieved by deduplicating some code, which I hope is welcomed change.


sharkwouter:
Yeah, that looks better and it work well. I'll merge it.

PR #279, opened 2021-01-25
Filesys utils

### LINKED ISSUE #218: Create filesystem_utils.py module
At the moment uninstalling game looks like:
```
def uninstall_game(game):
```
I am always scared when I see code like this. What if due to misconfiguration or bug, we end up with gamw.install_dir being some important user directory? Disaster.

I propose to create filesystem_utils.py module, which will abstract file operation (especially destructive ones: move, remove) and make sure that they executed only in ~/.cache/minigalaxy or ~/.config/minigalaxy or Config.get("install_dir").
Then we can make all modules to use filesystem_utils instead of directly calling files operation.
With such a change it will be a lot safer to perform files operations.
If you approve. I can proceed with the work.
Best regards,

makson96:
This is to fix #218 
Well it took me much longer that I was expecting, but I think that this module is now ready for broader testing and comments.

Basically this module will replace destructive files operations (write, move, remove, copy) with operations, which would double check if any important user files are not affected.
This module has some possibilities to introduce regressions which should be detected before merge, but what is most important, users files should be safe now.

### DIFF
--- a/minigalaxy/filesys_utils.py
+++ b/minigalaxy/filesys_utils.py
@@ -2,12 +2,18 @@ import subprocess
 import os
 import shutil
 import json
-from minigalaxy.config import Config
-from minigalaxy.paths import CONFIG_DIR, CACHE_DIR
+from minigalaxy.paths import CONFIG_DIR, CACHE_DIR, CONFIG_FILE_PATH, DEFAULT_INSTALL_DIR
 
 
 def _get_white_list():
-    return [Config.get("install_dir"), CONFIG_DIR, CACHE_DIR]
+    json_config = {"install_dir": DEFAULT_INSTALL_DIR}
+    if os.path.exists(CONFIG_FILE_PATH):
+        with open(CONFIG_FILE_PATH, "r") as file:
+            try:
+                json_config = json.loads(file.read())
+            except json.decoder.JSONDecodeError:
+                pass
+    return [json_config["install_dir"], CONFIG_DIR, CACHE_DIR]
 
 
 def _get_black_list():
@@ -76,7 +82,7 @@ def _check_if_ok_for_copy_or_move(source="", target="", recursive=False, overwri
         elif not os.path.isdir(os.path.dirname(target)):
             err_msg = "Directory for target operation doesn't exists: {}".format(os.path.dirname(target))
         elif os.path.isdir(source) and not recursive:
-            err_msg = "Non recursive requested on directory:{}".format(target)
+            err_msg = "Non recursive requested on directory: {}".format(target)
         elif os.path.exists(target) and not overwrite:
             err_msg = "Non overwrite operation, but target exists:{}".format(target)
     return err_msg
--- a/minigalaxy/filesys_utils.py
+++ b/minigalaxy/filesys_utils.py
@@ -153,3 +153,16 @@ def write_file(file_content, file_path, overwrite=False, append=False):
             file_to_write.write(file_content)
             file_to_write.close()
     return err_msg
+
+
+def mkdir(target, parents=False):
+    target = os.path.abspath(target)
+    err_msg = _check_if_accordance_with_lists(target)
+    if not err_msg:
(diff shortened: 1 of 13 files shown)


sharkwouter:
Thanks for the PR. I've finally found the time to test this. I did find an issue with it, though. It seems updating doesn't work with it at the moment. I'm seeing this error:

Screenshot from 2021-02-06 13-52-03

minigalaxy/filesys_utils.py
```diff
@@ -0,0 +1,174 @@
...
+            is_blacklisted = True
+            break
+    if not inside_white_list:
+        err_msd = "{} is not inside white list for file operations.".format(target)
+    elif is_blacklisted:
+        err_msd = "{} is on black list for file operations.".format(target)
+    return err_msd
+
+
+def _check_if_ok_for_copy_or_move(source="", target="", recursive=False, overwrite=False):
```
sharkwouter, line 71: For private functions I would use double underscore.

minigalaxy/filesys_utils.py
```diff
@@ -0,0 +1,174 @@
...
+        with open(CONFIG_FILE_PATH, "r") as file:
+            try:
+                json_config = json.loads(file.read())
+            except json.decoder.JSONDecodeError:
+                pass
+    return [json_config["install_dir"], CONFIG_DIR, CACHE_DIR]
+
+
+def _get_black_list():
+    important_user_dirs = ["", "DESKTOP", "DOCUMENTS", "DOWNLOAD", "MUSIC", "PICTURES", "PUBLICSHARE", "TEMPLATES",
```
sharkwouter, line 20: This seems a bit specific. Could you explain the logic behind this? I don't think we are removing directories which are filled with files other than installation directories at the moment.

sharkwouter (commented):
I'm a little bit worried that this might not integrate as well as it would have before. I'm not quite sure if there are any significant merge issues at the moment, though. The idea of having some checks in place before deleting a directory is really good, but I do worry that we're not checking if a directory should be deleted in a way that adds much.

minigalaxy/filesys_utils.py
makson96: Those are standard xdg-user-dir
User technically can remove them, but we for sure don't want to do this. This would be the most damaging thing that can happen by accident.

Those entries get localized to actual user directories. Lets take "VIDEOS" for example:
```
>>> import subprocess
>>> subprocess.check_output(['xdg-user-dir', "VIDEOS"]).decode("utf-8").strip()
'/home/makson/Wideo'
>>> 
```

minigalaxy/filesys_utils.py
makson96: Such a change does produce issues for unittests. Do we really need this?

makson96:
- The issue you mentioned (with updates) is already addressed in this branch.
- Branch is updated with current master
- Changes in master in regard to files operation are addressed here.

It is ok. I know this seems to be big change so we have time to shape it, so it addresses all worries regarding files operations.

At the moment the main thing which this pull request does is during destructive files operations:
- check if we are performing operation only in directory in white list, what includes only:
```
json_config["install_dir"], CONFIG_DIR, CACHE_DIR
```
- standard user directories are on the blacklist. So even if due some misconfiguration, we end up with "install_dir" being "/home/user/Documents", this "Documents" directory won't be removed.
- We got a little more control. For example Python for default overwrites files. With this pull request we overwrites files only if we specifies so (this was the cause of issue you mentioned with updates).
- If we do something incorrect, instead of Python exception, this module return error message, which can be displayed to the user (as on the screen you show).

If you have other suggestions how to improve safety here, please let me know

makson96:
Lets convert it to the draft and check if we can make easier solution.

sharkwouter:
I'm going to close this for now, since I think it doesn't fit that well into the current state of Minigalaxy. We should look into a different solution once I'm done with refactoring.

PR #284, opened 2021-02-06
Remove old general FPS value

TotalCaesar659:
It was a leftover after changing to game-related ones.

### DIFF
--- a/minigalaxy/constants.py
+++ b/minigalaxy/constants.py
@@ -32,7 +32,6 @@ DEFAULT_CONFIGURATION = {
     "install_dir": DEFAULT_INSTALL_DIR,
     "keep_installers": False,
     "stay_logged_in": True,
-    "show_fps": False,
     "show_windows_games": False,
     "keep_window_maximized": False
 }


Odelpasso:
Why do you want delete this option ?
Show_FPS is still used, the option has been moved to Properties from Preference window.

TotalCaesar659:
It's used? Well, FPS value now stores in separate minigalaxy-info.json for the game, but ""show_fps": False" is also saved in main minigalaxy.json. So I thought it's doesn't needed (because it can't be changed in main preferences now), tested and haven't found any bugs.
OK I can close this PR if you want.

sharkwouter:
> Why do you want delete this option ?
> Show_FPS is still used, the option has been moved to Properties from Preference window.

This is a bit strange. Maybe we could change this, since it's now per game it's a bit odd to still need to global configuration value.

TotalCaesar659:
@sharkwouter I'm little bit confused too. If you want to change it, tell me and I'll re-open it.

PR #289, opened 2021-02-13
Don't show gog installer 

Odelpasso:
Hi,

For Windows games and Wine, Minigalaxy use "/dir=" command argument to force the installation in a specific folder.
I think we don't need to see the installation's window and install the game silently is better. A specific command argument exists to do that.
I try it and works correctly.

Example of installation's window for Windows games :
Capture d’écran de 2021-02-13 11-27-05

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -69,7 +69,7 @@ def extract_installer(game, installer, temp_dir):
             os.makedirs(prefix_dir, mode=0o755)
 
         # It's possible to set install dir as argument before installation
-        command = ["env", "WINEPREFIX={}".format(prefix_dir), "wine", installer, "/dir={}".format(temp_dir)]
+        command = ["env", "WINEPREFIX={}".format(prefix_dir), "wine", installer, "/dir={}".format(temp_dir), "/VERYSILENT"]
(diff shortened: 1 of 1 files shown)


sharkwouter:
I just did a quick test, but this doesn't always seem to work with wine version 4.0. The following error happens every time and I'm seeing some installers fail
Screenshot from 2021-02-13 13-22-25
Screenshot from 2021-02-13 13-14-54
Screenshot from 2021-02-06 13-52-03

sharkwouter:
What I do want to say is that if this ends up working that would be amazing

sharkwouter:
I've converted this to a draft, because it's not really done yet.

Odelpasso:
Just a question, apart Debian Stable user, who use Wine-4.x or wine-staging-4.x from now ?^^
1- do you want i add a condition to show gog-installer if wine is <= 4.0
or
2- Wait for the release for new Debian Stable which have Wine-5.0 ?

I tried with wine-5.0 on Debian Sid and the "/VERYSILENT" option works.

sharkwouter:
Well, Debian Stable is still used, I use it. That's not my main worry, though. Something is going wrong sometimes which causes the installer to crash, but there isn't really a good error message as to what happened.

sharkwouter:
I've thought about this some more and I think this is a great addition. I've merged it. Thanks for the work!

PR #291, opened 2021-02-14
Add disk space checking before download part2.

makson96:
Hi,
This is continuation of this pull request:
https://github.com/sharkwouter/minigalaxy/pull/147
Original author stated on Discord, that he won't have time to finish it and it is ok for someone else to do it.

I am pushing it here with some changes as pull request. It looks like this now:
disk_space_check

Please do not merge it yet as I want to:
- Write some unit tests
- Add config option to ignore disk space check (it is based on approximation, so in some cases it might be very bad and user might want to disable it)
- Tune magic_ratio

Before I continue I would like to hear your comments about it.
Best regards,


sharkwouter (commented):
This seems like a pretty good solution, although it does have a disadvantage. When the download actually starts, the disk space available can have changed. I'm not sure if this PR addresses that well. The code does look clean in general.

An alternative method to do this would probably be to also check when the download actually starts. Not sure. The download manager does know the size of a file it downloads when it start downloading something, but it doesn't know the Window so it cannot correctly show an error message at the moment. Something to think about, I'm not sure what the best solution is at the moment.

minigalaxy/ui/gametile.py
```diff
@@ -252,7 +255,12 @@ def __download(self, download_info, finish_func, cancel_to_state):
             )
             self.download.append(download)
 
-        DownloadManager.download(self.download)
+        if check_diskspace(total_file_size, Config.get("install_dir")):
+            DownloadManager.download(self.download)
+        else:
+            GLib.idle_add(self.parent.parent.show_error, _("Download error"),
+                          _("Not enough disk space to install game."))
```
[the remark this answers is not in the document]
makson96: Well, I am doing here also approximation of installed game size (based on installer size), so as a result we actually check if we have space for whole install process to take place.

minigalaxy/installer.py
```diff
@@ -7,6 +7,28 @@
...
+
+def check_diskspace(installer_size, location):
+    """This method will return True when the disk space available is sufficient
+    for the Download and Install. If not sufficient, it returns False."""
+    magic_ratio = 1.5
+    installed_game_size = int(float(installer_size) * magic_ratio)
+    total_required_disk_space = installer_size + installed_game_size
+    diskspace_available = get_availablediskspace(location)
+    # Reserved space can be used to mimic a full disk or prevent filling a disk completely.
+    reserved_space = 0
```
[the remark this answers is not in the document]
makson96: What do you thing about 500MB as a safe reserve?

sharkwouter:
I've converted this to a draft, since you're saying not to merge this yet.

makson96:
Sorry for long wait. Code is now ready for merge. Please check it.

makson96:
To test it manipulated "reserved_space" in check_diskspace().
After receiving error (during game installation), change "Ignore disk space" in preferences and retry to install game.

makson96:
OK. Changes requested on Discored are implemented. Please review.

minigalaxy/installer.py
```diff
@@ -7,6 +7,28 @@
...
+
+def check_diskspace(installer_size, location):
+    """This method will return True when the disk space available is sufficient
+    for the Download and Install. If not sufficient, it returns False."""
+    magic_ratio = 1.5
+    installed_game_size = int(float(installer_size) * magic_ratio)
+    total_required_disk_space = installer_size + installed_game_size
+    diskspace_available = get_availablediskspace(location)
+    # Reserved space can be used to mimic a full disk or prevent filling a disk completely.
+    reserved_space = 500*1024*1024
```
sharkwouter, line 28: I'd leave out reserved space. We can't really determine that and not all partitions have it, only the root partition.

minigalaxy/ui/gametile.py
```diff
@@ -252,7 +254,17 @@ def __download(self, download_info, finish_func, cancel_to_state):
             )
             self.download.append(download)
 
-        DownloadManager.download(self.download)
+        ds_ok = check_diskspace(total_file_size, Config.get("install_dir"))
```
sharkwouter, line 257: This doesn't really need a variable I think. It's only used once in the line below.

minigalaxy/installer.py
```diff
@@ -7,6 +7,28 @@
 from minigalaxy.config import Config
 
 
+def get_availablediskspace(location):
```
sharkwouter, line 10: Could you change this to get_available_disk_space?

minigalaxy/installer.py
```diff
@@ -7,6 +7,28 @@
...
+    absolute_location = os.path.realpath(location)
+    disk_status = os.statvfs(os.path.dirname(absolute_location))
+    available_diskspace = disk_status.f_frsize * disk_status.f_bavail
+    return available_diskspace
+
+
+def check_diskspace(installer_size, location):
+    """This method will return True when the disk space available is sufficient
+    for the Download and Install. If not sufficient, it returns False."""
+    magic_ratio = 1.5
```
sharkwouter, line 23: I don't think we'll need the magic ratio in the future. Not sure if we should keep it for now or not.

sharkwouter (changes requested):
Thanks for addressing it! The code looks really good, I just had a couple of small nitpicks.

makson96:
Please review. You can change line:
return size_value
in get_game_size_from_unzip()
to:
return size_value * 1000000000000
to generate new error.

sharkwouter (changes requested):
This is really good now, awesome job. I've done some testing and it works really well. The method you use to check the disk space is very cool actually. I just found one minor nitpick, then it can be merged.

makson96:
Done. Please review.

sharkwouter:
Thanks! I've merged it.

PR #292, opened 2021-02-15
Attempt to fix semi-randomly failing unit test.

makson96:
Could you re run few times unit tests on GitHub with this patch to check if it helps to fix "test_save_minigalaxy_info_json" unit tests which fails randomly?


sharkwouter:
Sure, I'll re-run it now.

sharkwouter:
Ran this 4 times now. Seems to fix the issue.

PR #293, opened 2021-02-15
fix_290

### LINKED ISSUE #290: Intregrety check fails
Hello,

There is a bug since integrity check is implemented in minigalaxy-git.
This issue does not appears if the Linux/Windows installer has ONLY one file. Linux games are never touched by this game, like Windows game which have an only .exe file.

Step to reproduce : 
- Enable Windows games
- Download a Windows which have 2 and more files (.exe and .bin)

Cheers,
Maxime

makson96:
This is attempt to fix #290 
Now game.md5sum will be dictionary with key - file name, value - file md5sum.
This way we will be sure that we always check md5sum of correct file.

@Odelpasso could you check if this one solves the issue?

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -34,15 +34,20 @@ def verify_installer_integrity(game, installer):
     if not os.path.exists(installer):
         error_message = _("{} failed to download.").format(installer)
     if not error_message:
-        hash_md5 = hashlib.md5()
-        with open(installer, "rb") as installer_file:
-            for chunk in iter(lambda: installer_file.read(4096), b""):
-                hash_md5.update(chunk)
-        calculated_checksum = hash_md5.hexdigest()
-        if game.md5sum[os.path.basename(installer)] == calculated_checksum:
-            print("{} integrity is preserved. MD5 is: {}".format(installer, calculated_checksum))
-        else:
-            error_message = _("{} was corrupted. Please download it again.").format(installer)
+        for installer_file_name in os.listdir(os.path.dirname(installer)):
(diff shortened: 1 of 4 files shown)


minigalaxy/installer.py
```diff
@@ -39,7 +39,7 @@ def verify_installer_integrity(game, installer):
             for chunk in iter(lambda: installer_file.read(4096), b""):
                 hash_md5.update(chunk)
         calculated_checksum = hash_md5.hexdigest()
-        if game.md5sum == calculated_checksum:
+        if game.md5sum[os.path.basename(installer)] == calculated_checksum:
```
sharkwouter, line 42: Might want to catch a KeyError here, just in case.

sharkwouter (commented):
Looks pretty good, still need to test it, but I'll be able to do that later.

sharkwouter:
It seems it resolves the error, but I don't think all packages are tested. I tried with Price of Persia and it only tested the EXE:
```
/home/wouter/.cache/minigalaxy/download/Prince of Persia The Sands of Time/setup_prince_of_persia_-_the_sands_of_time_181_(28548).exe integrity is preserved. MD5 is: 008b4359cafb21dd4e437860dc23e2cf
```

makson96:
And does it work now as intended?

sharkwouter:
Yeah, that works! Should I merge it, or would you like to make changes still?

makson96:
I am done with this one. You can merge it. Thanks.

sharkwouter:
Great! Thanks for the work!

PR #303, opened 2021-04-02
Make 'Installed' filter permanent upon restarts.

makson96:
As the titles indicates.
This pull request will make "Installed" filter to be preserved upon Minigalaxy restart.
I can't find issue about it, but I recall it was request on Discord.


sharkwouter:
Thanks! That's a cool feature. It seems to work well too

PR #304, opened 2021-04-02
Attempt to fix Config interuption for unittests

makson96:
In GitHub unittests fails semi randomly. It is because Config import already starts daemon thread, which fails in this artificial environment and as a result it might be received as unittest fail.
This pull request delay Config thread initialization to the moment get() set() or unset() are used for the first time. As a result, simple Config import does not start new threads and hopefully will make all unittests always pass.
End user should not fell any differences.


sharkwouter:
Thanks! Lets see if that solves the problem

Issue #277, opened 2021-01-23
Manually select game version


Angell1993:
I'd like the option to select either Linux or Windows version of a game if available..

Seeing that I've got a game that will only install the broken Linux version.. While the Windows version works perfectly in wine..

sharkwouter:
Thanks for the report, we are building up to this feature. The first part in the process would be #243. After that has been merged, we plan on working on this.

Angell1993:
Well the new properties window, looks nice.. And I can use an alternative way to download the games that has issues at this time.. I already can't wait to see how the update turns out..

LiamDawe:
Don't suppose there was any update on this? Would be real helpful for a lot of things.

GB609:
Sorry for the late response.
An option to chose between windows and linux has been asked for multiple times in the past. 

It is not yet done, but i've actually had some code for this in preparation for quite a while. I'm just not done and had little time recently.

Issue #273, opened 2021-01-14
Add option to hide games


makson96:
As on Discord request:
"
Jero09.01.2021
Hi, just a feature idea since I would really like to see it. I would like to hide some games from the list. I did claim some free games on gog which I don't want to play or to see, so I don't want to see them in my list
"
There should be obviously option to unhide games. Maybe in hamburger menu on top panel?

sharkwouter:
We can add it to the properties menu and add "show hidden games" to the hamburger menu

PR #305, opened 2021-04-02
New game configuration directory

makson96:
This pull request moves games minigalaxy-info.json file to ~/.config/minigalaxy/<Game-Name>.json.
This allows us to do some configuration to games which are not yet install and as a result adds opportunity to implement for example:
Hiding games - #273 
Choosing between Windows and Linux game version to install - #277 

We preserve compatibility with previous minigalaxy-info.json files (Minigalaxy 1.0.1 and 1.0.2) and those configs are translated to new once.

We break compatibility with minigalaxy-dlc.json (Minigalaxy 1.0.0). If someone for some reason have such a config, DLCs will be marked as not installed. Upon reinstalling them everything should be back to normal.

### DIFF
--- a/minigalaxy/game.py
+++ b/minigalaxy/game.py
@@ -3,6 +3,7 @@ import re
 import json
 
 from minigalaxy.config import Config
+from minigalaxy.paths import CONFIG_DIR
 
 
 class Game:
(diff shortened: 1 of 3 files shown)


sharkwouter:
This seems to work quite well and I think it is something we really need. The only thing that bothers me is the directory the files are in. Could this be changed to ``~/.config/minigalaxy/games/`` or something similar? At the moment it makes it harder to find the general minigalaxy configuration.

makson96:
Done.

sharkwouter:
Thanks!

PR #307, opened 2021-04-07
Initial redesign of Properties Window

makson96:
This pull request is to address #106
It is done by redesigning Properties Window, which can be also shown for not installed games.
My vision is pretty much complete so I am ready for some output if look needs some changes:
new_properties
This is only draft as code still needs some polish:
- offline mode won't work
- no unit tests
- some PEP-8 improvements should be also put here
- description is available in multiple languages, so we should use it
- Do we want to download picture asynchronous from the rest of the window (performance)?

So what do you think about it?

### DIFF
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -224,3 +224,23 @@ class Api:
             print("Response body: {}".format(response.text))
             print("")
         return response.json()
+
+    @staticmethod
+    def __request_gamesdb(game: Game):
+        request_url = "https://gamesdb.gog.com/platforms/gog/external_releases/{}".format(game.id)
+        response = SESSION.get(request_url)
+        return response.json()
+
+    def get_gamesdb_info(self, game: Game) -> dict:
+        gamesdb_dict = {"cover": "", "vertical_cover": "", "background": ""}
+        response_json = self.__request_gamesdb(game)
+        if "game" in response_json:
+            for gamesdb_key in gamesdb_dict:
+                if gamesdb_key in response_json['game']:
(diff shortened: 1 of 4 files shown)


sharkwouter:
That looks pretty good. I'd say the most important part in getting this working well is to store the text in the info json file, so we don't need a network connection.

makson96:
My idea was that in lack of network connection we will fall back to old thumbnail picture and no description text.
But I will also investigate keeping description text in json file. Please note that I am heading for multilanguage support with this one.

makson96:
OK. I am ready for the review.

sharkwouter:
Thanks! I'll make time to review it tomorrow evening

sharkwouter:
I really love how this is looking at the moment

minigalaxy/api.py
```diff
@@ -224,3 +225,32 @@ def __request(self, url: str = None, params: dict = None) -> dict:
             print("Response body: {}".format(response.text))
             print("")
         return response.json()
+
+    @staticmethod
```
sharkwouter, line 229: Why is this static? It uses the session, right? Same above here.

sharkwouter:
It looks good. If you resolve the conflict and clear up the one comment I'll merge it.

minigalaxy/api.py
makson96: It is static, because it doesn't use "self" anywhere. As a result it can be placed outside of the class with no consequences. PyCharm marks such a things as violation of PEP-8 and wants to mark method as static or place it as a function outside of the class.

makson96:
OK. It should be now compatible with "Hide game".

sharkwouter:
Thanks!

PR #309, opened 2021-04-11
Add option to hide games in library

TotalCaesar659:
I'm tired to wait this option and tried to make it by myself. @makson96 please check this.
Fixes https://github.com/sharkwouter/minigalaxy/issues/273.

### DIFF
--- a/minigalaxy/ui/library.py
+++ b/minigalaxy/ui/library.py
@@ -11,6 +11,7 @@ from minigalaxy.paths import UI_DIR
 from minigalaxy.api import Api
 from minigalaxy.config import Config
 from minigalaxy.game import Game
+from minigalaxy.launcher import config_game
 from minigalaxy.ui.gametile import GameTile
 from minigalaxy.translation import _
 
@@ -67,7 +68,7 @@ class Library(Gtk.Viewport):
             self.search_string = widget.get_text()
         self.flowbox.set_filter_func(self.__filter_library_func)
 
-    def __filter_library_func(self, child):
+    def __filter_library_func(self, child, game):
         tile = child.get_children()[0]
         if self.search_string.lower() not in str(tile).lower():
             return False
@@ -75,6 +76,9 @@ class Library(Gtk.Viewport):
         if self.show_installed_only:
             if tile.current_state in [tile.state.DOWNLOADABLE, tile.state.INSTALLABLE]:
                 return False
+
+        if Config.get("show_hidden_games") and not game.get_info("hide_game"):
+            return False
 
         return True
 
@@ -95,8 +99,9 @@ class Library(Gtk.Viewport):
 
         for game in self.games:
             if game not in games_with_tiles:
-                if Config.get("show_hidden_games") or not game.get_info("hide_game"):
-                        self.__add_gametile(game)
+                self.__add_gametile(game)
+                #if Config.get("show_hidden_games") or not game.get_info("hide_game"):
+                        #self.__add_gametile(game)
 
     def __add_gametile(self, game):
         self.flowbox.add(GameTile(self, game))
--- a/minigalaxy/ui/library.py
+++ b/minigalaxy/ui/library.py
@@ -11,7 +11,6 @@ from minigalaxy.paths import UI_DIR
 from minigalaxy.api import Api
 from minigalaxy.config import Config
 from minigalaxy.game import Game
-from minigalaxy.launcher import config_game
 from minigalaxy.ui.gametile import GameTile
 from minigalaxy.translation import _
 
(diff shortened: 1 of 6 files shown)


makson96:
I had slightly different idea about implementation, but such an approach seems to be fine too (I was thinking about button in hamburger menu to unhide all game).

There are some issues in the implementation (I will mark it in the code):
- games are hidden/unhidden from user only after Minigalaxy restart (or new show_hidden_games option change in Preferences). Games should be hidden after clicking proper option in game Properties and then OK instead.
- In library.py please do not change "__create_gametiles()". Please put proper implementation in "__filter_library_func()" and use "filter_library()" whenever you want to hide or unhide game.
- It would be desired to be also able to hide not installed games. Proper config support is now in place for this, but it is blocked in your code.

Thanks for your work.

minigalaxy/ui/library.py
```diff
@@ -95,7 +95,8 @@ def __create_gametiles(self) -> None:
 
         for game in self.games:
             if game not in games_with_tiles:
-                self.__add_gametile(game)
+                if Config.get("show_hidden_games") or not game.get_info("hide_game"):
+                        self.__add_gametile(game)
```
makson96, line 99: Please put proper implementation in __filter_library_func() instead of __create_gametiles()

minigalaxy/ui/preferences.py
```diff
@@ -98,6 +100,10 @@ def save_pressed(self, button):
         Config.set("keep_installers", self.switch_keep_installers.get_active())
         Config.set("stay_logged_in", self.switch_stay_logged_in.get_active())
 
+        if self.switch_show_hidden_games.get_active() != Config.get("show_hidden_games"):
+            Config.set("show_hidden_games", self.switch_show_hidden_games.get_active())
+            self.parent.reset_library()
```
makson96, line 105: Once implementation is done in __filter_library_func() we will not need this "if" condition and reset_library(). Just filter_library() instead.

minigalaxy/ui/properties.py
```diff
@@ -105,6 +110,7 @@ def button_sensitive(self, game):
             self.entry_properties_variable.set_sensitive(False)
             self.button_properties_regedit.set_sensitive(False)
             self.switch_properties_show_fps.set_sensitive(False)
+            self.switch_properties_hide_game.set_sensitive(False)
```
makson96, line 113: We don't need it. It should be possible to hide not installed games now.

minigalaxy/ui/properties.py
```diff
@@ -56,6 +60,7 @@ def ok_pressed(self, button):
             self.game.set_info("variable", str(self.entry_properties_variable.get_text()))
             self.game.set_info("command", str(self.entry_properties_command.get_text()))
             self.game.set_info("show_fps", self.switch_properties_show_fps.get_active())
+            self.game.set_info("hide_game", self.switch_properties_hide_game.get_active())
```
makson96, line 63: This doesn't need to be under "if" condition (should work for not installed games too).

Once implementation is done in __filter_library_func() use filter_library() here.

makson96 (changes requested):

TotalCaesar659:
> I was thinking about button in hamburger menu to unhide all game.

It's better to move it to options window, cause there are less chance that somebody behind your back will see it. The window behind two mouseclicks is more safer than hamburger menu behind one I guess.

I tried to move the condition to __filter_library_func(), but it shows me "TypeError: __filter_library_func() missing 1 required positional argument: 'game'".

makson96:
Instead of using "game" try "tile.game" here.
For example:
tile.game.get_info("hide_game")

TotalCaesar659:
The condition works now, but saving shows "TypeError: filter_library() missing 1 required positional argument: 'switch'".

makson96:
Hmm. Could you check if calling:
self.parent.library.filter_library()
instead of:
self.parent.filter_library()
solves this issue?

TotalCaesar659:
The warning is gone, but library isn't updated and setting isn't saved.

makson96:
You should add:
Config.set("show_hidden_games", self.switch_show_hidden_games.get_active())
in:
def save_pressed() (preferences.py)
If you have further issues, please remember to update your branch.

TotalCaesar659:
I should have guessed. In properties.py (https://github.com/sharkwouter/minigalaxy/pull/309/files#diff-ad7117370a7d1ef849b4e36d90fd4bbade6985c915a37846ab3f0b7fc61eb4f5R64):
- self.parent.filter_library() shows "'GameTile' object has no attribute 'filter_library'";
- self.parent.library.filter_library() shows 'GameTile' object has no attribute 'library'.

makson96:
I think that "Window" i parent of "Game Tile". Which makes it grandparent of "Properies" ;)
Try:
self.parent.parent.library.filter_library()

TotalCaesar659:
`self.parent.parent.library.filter_library()` shows "'Library' object has no attribute 'library'", but `self.parent.parent.filter_library()` works. Looks like it's ready for review.

makson96 (approved):
I have tested it and it works as expected. Code also looks good for me.
@sharkwouter can you take a look?

sharkwouter:
Looks really good now, thanks!

Issue #271, opened 2021-01-14
Couldn't connect to GOG servers


voltumna:
Hi all, 
I downloaded and installed minigalaxy app, awesome idea! 
I am having an issue: when I started the app the first time, it showed me the login prompt for GOG, and that was fine, but then right away after logging in I get this message:

"Couldn't connect to GOG servers"

any suggestions?

Thanks

sharkwouter:
Thanks for the report, I've not seen this before. Which version is this? Can you access the GOG website? If you launch minigalaxy from the terminal, do you see any output?

voltumna:
Hi, 
thank you for the fast feedback. 
Via the browser, I can totally connect, i can log in and everything you can do on the website,

I have tried via the commandline, but there is no output at all, 

I have version 1.0.2
here is a screenshot with the output of the command line
![image](https://user-images.githubusercontent.com/1274922/104614020-20288700-5688-11eb-9439-c4f40a31670e.png)

sharkwouter:
Now that is strange. Maybe running Minigalaxy from the terminal like so will give some more information: ``MG_DEBUG=1 minigalaxy``

voltumna:
Thank you, i have done it and it looks like the took does get a response but it is not recognising it:
![image](https://user-images.githubusercontent.com/1274922/104620912-cf1c9100-568f-11eb-8a71-67f634a35040.png)

makson96:
OK. Lets debug this a little.
This error is provided by:
minigalaxy/ui/library.py
from "__add_games_from_api()" function.
It is becausethere were no games retrived by minigalaxy/api.py api.get_library().
Now I think this is because active_token is false for some reason.
I believe that something went wrong "response_params = response.json()" either in __refresh_token() or in __get_token()

@voltumna could you please modify minigalaxy/api.py file from line 35 to 72 so it would look like that (add additional prints):
```
    def __refresh_token(self, refresh_token: str) -> str:
        request_url = "https://auth.gog.com/token"
        params = {
            'client_id': self.client_id,
            'client_secret': self.client_secret,
            'grant_type': 'refresh_token',
            'refresh_token': refresh_token,
        }
        response = SESSION.get(request_url, params=params)

        response_params = response.json()
        print("refresh token")
        print(response_params)
        if "access_token" in response_params and "expires_in" in response_params and "refresh_token" in response_params:
            self.active_token = response_params["access_token"]
            expires_in = response_params["expires_in"]
[...]

PR #310, opened 2021-04-12
Fix error with empty library

makson96:
This pull request is to address issue #271 
Empty retrieved_games list is no reason to display errors (user for example can have only Windows games, which are not displayed by default).
But I still want to preserve possibility for get_library() to return error message if something went wrong. As a result I moved some things a little bit around.


sharkwouter:
What I can test seems to work, so I'll just merge this and hope for the best.

PR #314, opened 2021-04-15
Add logout confirmation window

TotalCaesar659:
This PR adds confirmation window that prevents accidental logout if you missed "Preferences" button in menu.
@makson96 I tried but `self.parent.__authenticate()` shows "AttributeError: 'Window' object has no attribute '_Logout__authenticate'".


sharkwouter:
Oh, that's a good idea actually! Thanks for the work!

Some feedback to this, look at the code. It seems you created a new window for the question, but we actually have a function which does this in window.py here:

https://github.com/sharkwouter/minigalaxy/blob/d2453488f3396a55f7d7af4cc14c433f73bef704/minigalaxy/ui/window.py#L123-L135

Here is an example of an implementation:

https://github.com/sharkwouter/minigalaxy/blob/d2453488f3396a55f7d7af4cc14c433f73bef704/minigalaxy/ui/gametile.py#L138-L142

TotalCaesar659:
Ah, thanks, I assumed there are more simple decision. Looks like it works now, ready for review.

data/ui/properties.ui
```diff
@@ -208,7 +208,7 @@
                     </child>
                     <child>
                       <object class="GtkButton" id="button_properties_open_files">
-                        <property name="label" translatable="yes">Open Files</property>
+                        <property name="label" translatable="yes">Open files</property>
```
sharkwouter, line 211: This could break translations, so I'm not sure if this is a good idea

sharkwouter (changes requested):

data/ui/properties.ui
TotalCaesar659: OK, I'll create separate PR then.

sharkwouter:
Thanks! Looking good!

PR #315, opened 2021-04-15
Fix translations

TotalCaesar659:
Fix translations (and updated Russian one).


TotalCaesar659:
It's better to be postponed for a while.

PR #316, opened 2021-04-16
Game version in Properties

makson96:
After yesterdays big Stellaris update, I figured out that it would be great if game version would be displayed in Properties for installed games.
version


sharkwouter:
Good idea, I'll test it

PR #317, opened 2021-04-16
Adapting Windows games to run native Linux

makson96:
At this point it is proof of concept, that requires further development.

It works like this:
In constants.py beside "IGNORE_GAME_IDS" there is also "ADAPTED_GAMES", which is the list of games that are adapted to nativly run on Linux. For each game, list contains info about its name, id and require. In case of Theme Hospital, the requirement is dosbox. If requirement is met, the game is listed with other native titles (no wine icon, etc.):
innoextract1a
When "download" button is pressed, game is normally downloaded:
innoextract2a
then innoextract (downloaded from web) is used to extract game data.
During installation, there is additional step "additional_configuration()", which will run custom script for each adapted game. Script will result in creating "minigalaxy-start.sh" which will be universal name for scripts to launch adapted games.
When you press "play" this script will run:
innoextract4b

So what now?
- Switching between extracting Windows games by wine and innoextract requires improvements.
- Choosing between Linux, Windows and Adapted game version needs to be added to "Properties".
- Add unit tests.
- Add more games to this mechanic.

This is still early stage, so I am curious about your opinion. At this point it will be still easy to change some stuffs.

### DIFF
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -4,7 +4,7 @@ from urllib.parse import urlencode
 import requests
 import xml.etree.ElementTree as ET
 from minigalaxy.game import Game
-from minigalaxy.constants import IGNORE_GAME_IDS, SESSION
+from minigalaxy.constants import IGNORE_GAME_IDS, ADAPTED_GAMES, SESSION
 from minigalaxy.config import Config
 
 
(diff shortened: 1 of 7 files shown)


makson96:
I have added ability to choose game platform from Properties:
platform
Linux (native) is blocked because this game doesn't provide official support for Linux.
Windows (wine) is blocked because I don't have wine installed.

Odelpasso:
Hi Makson, how are you ?

It is possible to update this commit ?
I'm interested to try the installation of Windows version of a game when a Linux version is available.

Thanks,

yoshimo:
In the case of Twopoint Hospital and an increasing number of other games we have the option to run a third party open source engine instead like https://github.com/corsixth/corsixth.

Issue #86, opened 2020-03-11
Add a language selection option for the interface


sharkwouter:
This could be a simple settings menu item, but it should make sure that a language can actually be used before letting the user pick it. It is not possible to load a translation for a locale which isn't installed.

PR #318, opened 2021-04-17
Add language switch to options

TotalCaesar659:
Locale changing applies only to preferences wndow (not to main one), I don't know why.
Fixes https://github.com/sharkwouter/minigalaxy/issues/86.

### DIFF
--- a/data/ui/preferences.ui
+++ b/data/ui/preferences.ui
@@ -51,13 +51,38 @@
           </packing>
         </child>
         <child>
-          <!-- n-columns=3 n-rows=5 -->
+          <!-- n-columns=2 n-rows=7 -->
           <object class="GtkGrid">
             <property name="visible">True</property>
             <property name="can-focus">False</property>
             <property name="row-spacing">2</property>
             <property name="row-homogeneous">True</property>
             <property name="column-homogeneous">True</property>
+            <child>
+              <object class="GtkLabel">
+                <property name="visible">True</property>
+                <property name="can-focus">False</property>
+                <property name="tooltip-text" translatable="yes" context="program_language_tooltip">The preferred language on Minigalaxy start</property>
+                <property name="halign">start</property>
+                <property name="valign">center</property>
+                <property name="label" translatable="yes" context="program_language" comments="Program language. Ends with &quot;: &quot;.">Program language: </property>
+              </object>
+              <packing>
+                <property name="left-attach">0</property>
+                <property name="top-attach">0</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkComboBox" id="combobox_program_language">
+                <property name="visible">True</property>
+                <property name="can-focus">False</property>
+                <property name="valign">center</property>
+              </object>
+              <packing>
+                <property name="left-attach">1</property>
+                <property name="top-attach">0</property>
+              </packing>
+            </child>
             <child>
               <object class="GtkLabel">
                 <property name="visible">True</property>
@@ -69,7 +94,7 @@
               </object>
               <packing>
                 <property name="left-attach">0</property>
-                <property name="top-attach">0</property>
+                <property name="top-attach">1</property>
               </packing>
             </child>
             <child>
@@ -80,7 +105,7 @@
               </object>
               <packing>
                 <property name="left-attach">1</property>
-                <property name="top-attach">0</property>
+                <property name="top-attach">1</property>
               </packing>
             </child>
             <child>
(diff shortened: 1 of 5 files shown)


sharkwouter:
Thanks for the PR. This is an interesting one. I really like the idea of having this.

I see a couple of problems, though:

- By default no language is selected in the menu. Try to detect which one is used.
- If a locale is enabled which is not installed on the user's machine, it is no longer possible to start Minigalaxy.
- Unsupported locales are shown in the preferences menu.
- Changing the language requires Minigalaxy to be restarted, but the user is not informed of this.

I'm not sure if it would be possible to load a translation independent of the currently selected locale, but that might be worth looking into. Even if without the locale the characters required might not always load.

I fixed one small issue which made it not possible to start Minigalaxy without manually setting the language in the properties file.

TotalCaesar659:
Point 4 is fixed. For point 2, I tried to create the array of system locales, but it shows in line 69: "NameError: name 'system_locale_list' is not defined".
Well, I don't know about independent translation, so it's better postpone it for a while. If Point 2 will be fixed, wrong locales jist won't appear, and user can't set them in menu at least.

TotalCaesar659:
@sharkwouter I've added warning messages for non-existent locales. It was easier for me than trying to get list of system locales, besides user will see all locales supported by program.

sharkwouter:
I've just done another test with this, but it seems to work quite well if you do a restart of the application. Would it be possible to force the recreation of the application window?

data/ui/preferences.ui
```diff
@@ -51,13 +51,38 @@
...
             <property name="row-homogeneous">True</property>
             <property name="column-homogeneous">True</property>
+            <child>
+              <object class="GtkLabel">
+                <property name="visible">True</property>
+                <property name="can-focus">False</property>
+                <property name="tooltip-text" translatable="yes" context="program_language_tooltip">The preferred language on Minigalaxy start</property>
+                <property name="halign">start</property>
+                <property name="valign">center</property>
+                <property name="label" translatable="yes" context="program_language" comments="Program language. Ends with &quot;: &quot;.">Program language: </property>
```
sharkwouter, line 68: I think this field should probably be named "Language" and the field below it should be "Preferred game language".

sharkwouter (changes requested):
I'm very impressed by how well this works and how good the code looks right now. It's a big upgrade from when this PR was first made. I'm sorry for not looking at it for a while, that did add some extra work for you I see.

If this is rebased I'm willing to merge it. My comment about recreating the Window class object can be addressed in a separate MR if it's a significant amount of work.

TotalCaesar659:
I tried to rebase but it shows me this:
```
git rebase master
First, rewinding head to replay your work on top of it...
Applying: Add language switch to options (draft)
Applying: Do not set locale to nothing before a locale has been chosen
Applying: Fix locale saving (and more)
Applying: Add messages for non-existent locales
Applying: Fix change GUI locale
Applying: Remove game config together with game uninstall.
Applying: Simplified and removed comments.
Using index info to reconstruct a base tree...
M       minigalaxy/download_manager.py
Falling back to patching base and 3-way merge...
Auto-merging minigalaxy/download_manager.py
CONFLICT (content): Merge conflict in minigalaxy/download_manager.py
error: Failed to merge in the changes.
Patch failed at 0007 Simplified and removed comments.
hint: Use 'git am --show-current-patch' to see the failed patch
Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".
```


> I'm very impressed by how well this works and how good the code looks right now. It's a big upgrade from when this PR was first made. I'm sorry for not looking at it for a while, that did add some extra work for you I see.
> 
> If this is rebased I'm willing to merge it. My comment about recreating the Window class object can be addressed in a separate MR if it's a significant amount of work.

Yes, it's better to create another PR for this and just rebase here.

> I think this field should probably be named "Language" and the field below it should be "Preferred game language".

I want to make big PR to fix translations (similar to https://github.com/sharkwouter/minigalaxy/pull/315) right after so let's delay it for this.

sharkwouter:
That would work, thanks!

TotalCaesar659:
I wanted to ask how to rebase correctly, I do it first time and hardly undertstand sequence. Or do you want to do it by yourself?

sharkwouter:
No problem, I've rebased it. Rebasing is applying your commits on top of the current state. There were a few conflict with imports, but not many.

TotalCaesar659:
Is it ready for merge then? I want to start polish translations.

sharkwouter:
I guess my comment can be addressed in that PR then.

PR #319, opened 2021-04-19
Remove game config together with game uninstall.

makson96:
When game config was in game directory it was removed together with game uninstall.
Now it stays even if game is uninstalled.
Some really bad consequences of this behaviour is that DLC info is stored in this config file. So if game is installed again (without DLCs), Minigalaxy will think that DLCs are installed, because it will use old config.

This pull request solves this issue by removing game config file together with game uninstall.


sharkwouter:
Makes sense for now, thanks!

PR #326, opened 2021-06-05
Code cleanup and linter improvements

viacheslavka:
I'm interested in improving integration with different compatibility layers (proton, wine, whatever), thought this would be a good introduction to the project.

I've gone through the flake8 output, fixed the errors, added some changes of my own, and updated the Github workflow script. Now the linter is run twice: once with some errors ignored, but if it fails, the whole job fails; and then with no issues ignored and without allowing it to fail the whole job. See example here: https://github.com/viacheslavka/minigalaxy/runs/2750413482

Some of the changes here may be questionable, don't hesitate to let me know if I'm doing something wrong.


makson96:
Wow, it was 10 days since this pull request with no response. First of all sorry for that. Due to personal reasons (home renovation) I am temporary not able to work too much for the project (I don't even have a desk to work at home at the moment).

Secondary welcome to Minigalaxy community. It is great to have you here. I am glad that you want to work with wine and proton support. Users are heavily using those technologies, while I... don't even have wine installed (only native games for me).

I checked your changes. I am impressed by amount of changes you have made. They all seems to be good code quality improvements. I will leave one comment in the code for your consideration.

All units tests passed, which is great, but due to extensive changes, I think we need to do some extended manual testing to make sure no regressions are introduced.
I will let here know when I am done with testing.

Please consider joining our Discord channel. I visit Discord daily, so I can give you quick responses for your questions.
Best regards,

minigalaxy/game.py
```diff
@@ -29,45 +29,36 @@ def get_install_directory_name(self):
...
     @staticmethod
     def __strip_string(string):
         return re.sub('[^A-Za-z0-9]+', '', string)
 
     def is_installed(self, dlc_title="") -> bool:
-        installed = False
         if dlc_title:
             dlc_version = self.get_dlc_info("version", dlc_title)
-            installed = True if dlc_version else False
+            return bool(dlc_version)
```
makson96, line 49: Here we introduced second return statement to the function which previously had only one return. I prefer functions to have only one return. That way if future modifications are introduced we will make sure that they are executed in every function run. I know that we already have many functions with multiple returns, but I think that we should work on reducing that number.

viacheslavka: I'm not sure I understand the reasoning but I'll revert these changes if that's how it's done in this project

viacheslavka:
Don't worry about the delay, I know we all have our own business to attend to. I thought about joining the Discord server, I'll probably do it later when I have something to show in regards to compatibility layers support.

I'm glad you liked the changes, I saw your comment and will adjust them later (probably sometime closer to the weekend).

viacheslavka:
I've updated the commit you reviewed, hopefully I understood you correctly

makson96:
Thanks. It looks great. I will let you know when I am done with testing.

makson96:
I haven't found any regressions during manual testing.
@sharkwouter I am recommending a merge.

minigalaxy/ui/gametile.py
```diff
@@ -441,108 +442,123 @@ def reload_state(self):
...
+        tooltip_text = "{} (update{})".format(self.game.name, ", Wine" if self.game.platform == "windows" else "")
+        self.image.set_tooltip_text(tooltip_text)
+        self.menu_button_update.show()
+        if self.game.platform == "windows":
+            self.wine_icon.set_margin_left(22)
+
+    def __state_updating(self):
+        self.button.set_label(_("updating…"))
+
+    STATE_UPDATE_HANDLERS = {
```
sharkwouter, line 549: Nice, I love this

minigalaxy/ui/__init__.py
```diff
@@ -1,4 +1,5 @@
 """ MiniGalaxy gui windows """
-from minigalaxy.ui.window import Window
-from minigalaxy.ui.preferences import Preferences
-from minigalaxy.ui.gametile import GameTile
+# Flake8 thinks these imports are unused
+from minigalaxy.ui.window import Window             # noqa: F401
```
sharkwouter, line 3: I was wondering how you could exclude single lines from flake8. Looks good

minigalaxy/ui/gtk.py
```diff
@@ -0,0 +1,4 @@
+import gi
+
+gi.require_version('Gtk', '3.0')
+from gi.repository import Gtk, Gdk, Gio, GLib, GdkPixbuf  # noqa: E402,F401
```
sharkwouter, line 4: Ah, this is how you solved this. I've been wondering how to do this for a while. Looks like a good solution

sharkwouter (approved):
Great Work! I really appreciate this PR. I've been wanting to do this for a while, but it's such a daunting task. I'm merging this

minigalaxy/ui/gtk.py
viacheslavka: Yeah, there is a little problem with this solution, though: I couldn't find a way to "forward" the package imports. In other words, I don't think there's a way to allow something like this:
```python
from minigalaxy.ui.gtk.GdkPixbuf import Pixbuf
...
pixbuf = Pixbuf(...)
```
You always have to do it like this:
```python
from minigalaxy.ui.gtk import GdkPixbuf
...
pixbuf = GdkPixbuf.Pixbuf(...)
```
However, it's rarely needed (I've only made a couple changes across the whole codebase), so I think this solution should be good enough for a long while. If someone figures this out, that'd definitely be an improvement. It could probably be done by mirroring the GTK package structure, but I didn't think it was pretty enough. Anyway, as I said, the current solution should be good enough. I'm glad you like it

PR #332, opened 2021-06-18
Adds creation of .desktop files during installation

### LINKED ISSUE #54: Option for .desktop-files
When you install a game via the installer script you can create .desktop files on desktop and for the application menu. It would be nice to have this choice too, or add desktop files afterwards via a button.

itwasonlyabug:
* Adds a new function that handles the creation of .desktop files for Linux games. Wine games create them automatically.
* The function is called during installation and the file is removed after uninstalling the game.

Closes #54

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -2,8 +2,9 @@ import os
 import shutil
 import subprocess
 import hashlib
+import textwrap
 from minigalaxy.translation import _
-from minigalaxy.paths import CACHE_DIR, THUMBNAIL_DIR
+from minigalaxy.paths import CACHE_DIR, THUMBNAIL_DIR, DESKTOP_DIR
 from minigalaxy.config import Config
 
 
@@ -50,6 +51,8 @@ def install_game(game, installer):
         error_message = extract_installer(game, installer, tmp_dir)
     if not error_message:
         error_message = move_and_overwrite(game, tmp_dir, game.install_dir)
+    if not error_message:
+        error_message = create_desktop_file(game)
     if not error_message:
         error_message = copy_thumbnail(game)
     if not error_message:
@@ -152,6 +155,32 @@ def move_and_overwrite(game, temp_dir, target_dir):
     shutil.rmtree(temp_dir, ignore_errors=True)
     return error_message
 
+def create_desktop_file(game):
+    error_message = ""
+    if game.platform == "linux":
+        desktop_file_path = os.path.join(DESKTOP_DIR, game.name+".desktop")
+        # Create desktop file
+        desktop_context = {
+            "game_bin_path":os.path.join(game.install_dir,'start.sh'),
+            "game_name":game.name,
+            "game_icon_path":os.path.join(game.install_dir,'support/icon.png')
+            }
+        desktop_definition = """\
+        [Desktop Entry]
+        Type=Application
+        Terminal=false
+        StartupNotify=true
+        Exec="{game_bin_path}"
+        Name={game_name}
+        Icon={game_icon_path}""".format(**desktop_context)
+        if not os.path.isfile(desktop_file_path):
+            try:
+                with open(desktop_file_path, 'w+') as desktop_file:
+                    desktop_file.writelines(textwrap.dedent(desktop_definition))
+            except Exception as e:
+                os.remove(os.path.join(DESKTOP_DIR,game.name+".desktop"))
+                error_message = e
+    return error_message
 
 def copy_thumbnail(game):
     error_message = ""
@@ -183,3 +212,5 @@ def uninstall_game(game):
     shutil.rmtree(game.install_dir, ignore_errors=True)
     if os.path.isfile(game.status_file_path):
         os.remove(game.status_file_path)
+    if os.path.isfile(os.path.join(DESKTOP_DIR,game.name+".desktop")):
+        os.remove(os.path.join(DESKTOP_DIR,game.name+".desktop"))
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -1,10 +1,10 @@
 import os
 import shutil
 import subprocess
-import hashlib
 import textwrap
+import hashlib
 from minigalaxy.translation import _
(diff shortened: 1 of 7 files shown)


makson96:
Hi,
Thank you for your pull request. This is an important new feature.

I would like to provide some review for this one. i think that creating desktop shortcuts should not be part of every installation. In fact wine games behaviour (which does exactly that) is considered to be a bug.
Instead I would prefer to have this option i "Properties" window:
properties

I would also leave some comments in code itself.
Best regards,

minigalaxy/installer.py
```diff
@@ -152,6 +155,32 @@ def move_and_overwrite(game, temp_dir, target_dir):
     shutil.rmtree(temp_dir, ignore_errors=True)
     return error_message
 
+def create_desktop_file(game):
```
makson96, line 158: Please leave two empty lines between functions.

minigalaxy/installer.py
```diff
@@ -152,6 +155,32 @@ def move_and_overwrite(game, temp_dir, target_dir):
...
+        Name={game_name}
+        Icon={game_icon_path}""".format(**desktop_context)
+        if not os.path.isfile(desktop_file_path):
+            try:
+                with open(desktop_file_path, 'w+') as desktop_file:
+                    desktop_file.writelines(textwrap.dedent(desktop_definition))
+            except Exception as e:
+                os.remove(os.path.join(DESKTOP_DIR,game.name+".desktop"))
+                error_message = e
+    return error_message
```
makson96, line 183: Please leave two empty lines between functions.

minigalaxy/installer.py
```diff
@@ -152,6 +155,32 @@ def move_and_overwrite(game, temp_dir, target_dir):
     shutil.rmtree(temp_dir, ignore_errors=True)
     return error_message
 
+def create_desktop_file(game):
+    error_message = ""
+    if game.platform == "linux":
+        desktop_file_path = os.path.join(DESKTOP_DIR, game.name+".desktop")
+        # Create desktop file
+        desktop_context = {
+            "game_bin_path":os.path.join(game.install_dir,'start.sh'),
```
makson96, line 164: start.sh is not the only way we use to start a game. We should import proper launch command from launcher instead.

minigalaxy/installer.py
```diff
@@ -152,6 +155,32 @@ def move_and_overwrite(game, temp_dir, target_dir):
...
+            "game_name":game.name,
+            "game_icon_path":os.path.join(game.install_dir,'support/icon.png')
```
makson96, line 166: Please leave empty space after ":" and "," in lines above.

minigalaxy/installer.py
```diff
@@ -152,6 +155,32 @@ def move_and_overwrite(game, temp_dir, target_dir):
...
+        StartupNotify=true
+        Exec="{game_bin_path}"
+        Name={game_name}
+        Icon={game_icon_path}""".format(**desktop_context)
+        if not os.path.isfile(desktop_file_path):
+            try:
+                with open(desktop_file_path, 'w+') as desktop_file:
+                    desktop_file.writelines(textwrap.dedent(desktop_definition))
+            except Exception as e:
+                os.remove(os.path.join(DESKTOP_DIR,game.name+".desktop"))
```
makson96, line 181: please leave empty space after ",".

minigalaxy/installer.py
```diff
@@ -183,3 +212,5 @@ def uninstall_game(game):
     shutil.rmtree(game.install_dir, ignore_errors=True)
     if os.path.isfile(game.status_file_path):
         os.remove(game.status_file_path)
+    if os.path.isfile(os.path.join(DESKTOP_DIR,game.name+".desktop")):
+        os.remove(os.path.join(DESKTOP_DIR,game.name+".desktop"))
```
makson96, line 216: please leave empty space after ",".

minigalaxy/paths.py
```diff
@@ -11,6 +11,7 @@
 CACHE_DIR = os.path.join(os.getenv('XDG_CACHE_HOME', os.path.expanduser('~/.cache')), "minigalaxy")
 
 THUMBNAIL_DIR = os.path.join(CACHE_DIR, "thumbnails")
+DESKTOP_DIR = os.path.expanduser("~/.local/share/applications")
```
makson96, line 14: This doesn't look like desktop dir. It is rather application menu directory. One cleaver way I know to obtain desktop dir (for every language) is as follow:
"
import subprocess
subprocess.check_output(['xdg-user-dir', "DESKTOP"]).decode("utf-8").strip()
"
Nevertheless adding menu application launcher might be a good idea too.

makson96 (changes requested):
If you have any specific question do not hesitate to ask here or on our Discord channel.

itwasonlyabug:
Thanks for the feedback @makson96 ! I tried to follow the existing formatting but missed the details, I'll make the proposed changes.

To clarify on the folder / paths - my goal was to create the `.desktop` file in the Applications path, as that's the default location for `.desktop` files (or `/usr/share/applications`). Is the goal to have the icon on the actual Desktop (for DE's that support it) and/or both?

I'll do some more reading first, regarding the Properties windows, as I saw you have some extra considerations when creating buttons / toggles.

minigalaxy/paths.py
sharkwouter: I would recommend using ``DESKTOP_DIR = os.path.join(os.getenv('XDG_DATA_HOME', os.path.expanduser('~/.local/share')) + /applications`` here, so we adhere to the XDG standard.

sharkwouter:
I actually like this approach. I would say just making menu shortcuts is pretty good for now. I'd assume that if the user wants shortcuts to be created when installing games, he'll want it for every game he/she installs. I would, however, still like to see the following added before I'd consider this mergable:

- Add an option in the preferences to enable/disable the creation of these desktop files.
- Make sure desktop files created are removed when a game is uninstalled.

There are also some edge cases which might be annoying here. For example games using wine, scummvm or dosbox might require some additional work. An option to create desktop files for games which are already installed could be nice, but that should probably not be in this PR.

itwasonlyabug:
Closing as it's superceded by other PRs and this one is generally a mess.

Issue #313, opened 2021-04-13
Windows installers create useless shortcuts


sharkwouter:
This is especially annoying now you can no longer disable the option in the installer. The shortcuts which are created in the menu and on the desktop link to files which don't exist, since the installed files are moved after installation. According to the documentation you should be able to prevent the icons from being created using the ``/NOICONS`` option, but I tested this and it is not the case.

Here is the documentation: https://jrsoftware.org/ishelp/index.php?topic=setupcmdline

It may be possible to explicitly not run certain steps (tasks) in the installation process.

makson96:
This issue is now partially solved as we can now use innoextract to prepare wine game.
I believe that proper way to address it is to remove those wrong shortcuts. #343 will deal with creation of proper shortcuts.

It seems that wrong shortcuts are localized in (for example):
"
~/.local/share/applications/wine/Programs/GOG.com/Battle\ Realms\ Complete/'Battle Realms Complete.desktop'
"
And they example content is:
"
[Desktop Entry]
Name=Battle Realms Complete
Exec=env WINEPREFIX="/home/itwasonlyabug/GOG Games/Battle Realms  Winter of the Wolf/prefix" wine C:\\\\windows\\\\command\\\\start.exe /Unix /home/itwasonlyabug/GOG\\ Games/Battle\\ Realms\\ \\ Winter\\ of\\ the\\ Wolf/prefix/dosdevices/c:/ProgramData/Microsoft/Windows/Start\\ Menu/Programs/GOG.com/Battle\\ Realms\\ Complete/Battle\\ Realms\\ Complete.lnk
Type=Application
StartupNotify=true
Path=/home/itwasonlyabug/GOG Games/Battle Realms  Winter of the Wolf/prefix/dosdevices/z:/home/itwasonlyabug/.cache/minigalaxy/extract/1207659111/
Icon=92D0_gfw_high.0
StartupWMClass=battle_realms_f.exe
"

I think that during installation we should check all .desktop files in "~/.local/share/applications/wine/Programs/GOG.com/" and if the contain "/.cache/minigalaxy/extract/" remove them.

sharkwouter:
If Innoextract is installed this shouldn't happen anymore.

sharkwouter:
This issue makes me wonder if we should add Innoextract as a dependency. We probably should, although, wine is also optional.

GB609:
This is caused by wine's default desktop integration behaviour.
[...]

PR #334, opened 2021-06-20
Innoextract

makson96:
We can address #313 by extracting windows installers using innoextract instead of wine.

Please check this approach.

Code can use a little polishing before merge, so I will mark it as a draft for now.

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -2,6 +2,9 @@ import os
 import shutil
 import subprocess
 import hashlib
+import tarfile
+
+from minigalaxy.constants import SESSION, DOWNLOAD_CHUNK_SIZE
 from minigalaxy.translation import _
 from minigalaxy.paths import CACHE_DIR, THUMBNAIL_DIR
 from minigalaxy.config import Config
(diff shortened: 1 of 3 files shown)


sharkwouter:
I did a quick test of this and it seems to work very well for some games and not so well for others.

So far I tested:
- Absolute Drift, which starts, but doesn't work. Probably a wine issue
- Beyond Good and Evil, which doesn't work, the executable is not found
- Driver Parallel Lines, which works, but installs in Polish somehow

I'll be honest, though, I'm not a fan of the Innnoextract executable being downloaded instead of the system one being used.

sharkwouter:
It seems a couple of directories are being extracted which aren't needed. For instance tmp and _redist. I'm not sure if app and commonappdata are very usefull either.

sharkwouter:
This closes #344. It also closes #313

sharkwouter:
My main question with this PR, which does make me very reluctant to merge it, is if there are older versions of Innoextract which do not work with the current GOG archives. I'm really not a big fan of it downloading it as a binary instead of having it as a dependency.

makson96:
Thanks for the comment. I still did not find the time to finish this one.
At the moment innoextract binary is downloaded only if there is no innoextract already installed.
I can change this behavior. I will make innoextract as recommended dependence and fall back to wine if it is not installed. I will test it with innoextract 1.8 because this is what is present in Debian 11.

makson96:
Hi,
I am ready for review.
Tested with:
Debian 11 (with innoextract package installed from repository):
- Absolute Drift
- Aliens versus Predator Classic 2000
- Europa Universalis II
- Heroes of Might and Magic 3 Complete

minigalaxy/installer.py
```diff
@@ -106,51 +107,82 @@ def make_tmp_dir(game):
...
-def move_and_overwrite(game, temp_dir, target_dir):
+
+def extract_by_wine(game, installer, temp_dir):
+    err_msg = ""
+    # Set the prefix for Windows games
+    prefix_dir = os.path.join(game.install_dir, "prefix")
+    if not os.path.exists(prefix_dir):
+        os.makedirs(prefix_dir, mode=0o755)
+    # It's possible to set install dir as argument before installation
+    command = ["env", "WINEPREFIX={}".format(prefix_dir), "wine", installer, "/dir={}".format(temp_dir), "/VERYSILENT"]
```
sharkwouter, line 164: This shouldn't use a temp dir, because it breaks the shortcuts which are made. I would however, not require this to be done within this PR.

sharkwouter (approved):
Looks good and works well, thanks!

PR #341, opened 2021-07-24
Show Uninstall button after game installation.

makson96:
At the moment after game installation "Uninstall" button remain hidden. Uninstall button is only shown after Minigalaxy restart.

This pull request fixes this isssue by showing "Uninstall" button, when gametile status is changed to "installed".


sharkwouter:
Oh, good catch! Looks good to me.

PR #342, opened 2021-07-25
Center properties window

TotalCaesar659:


sharkwouter:
Thanks! That works even better than I expected

TotalCaesar659:
Great! Could you look at https://github.com/sharkwouter/minigalaxy/pull/318 then?

PR #343, opened 2021-07-30
Adds creation of desktop icons to Preferences dialog

itwasonlyabug:
- Adds a new function that handles the creation of .desktop files for Linux games.
- The function is called during installation and the file is removed after uninstalling the game.
- The function depends on a setting found in Preferences. The setting has two states - Enabled / Disabled. If enabled, newly installed games get a .desktop shortcut.

This PR is split from PR #332 and closes issue #54

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -39,7 +39,7 @@ def check_diskspace(required_size, location):
     return diskspace_available >= installed_game_size
 
 
-def install_game(game, installer):
+def install_game(game, installer): # noqa: C901
     error_message = ""
     tmp_dir = ""
     if not error_message:
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -57,7 +57,7 @@ def install_game(game, installer):
     if not error_message:
         error_message = create_applications_file(game)
     if not error_message:
-        error_message = remove_installer(installer)
+        error_message = remove_installer(game, installer)
     else:
         remove_installer(game, installer)
     if not error_message:
@@ -208,9 +208,13 @@ def create_applications_file(game):
     error_message = ""
     if Config.get("create_applications_file"):
         path_to_shortcut = os.path.join(APPLICATIONS_DIR, "{}.desktop".format(game.name))
+        exe_cmd_list = get_execute_command(game)
+        for element in exe_cmd_list:
+          element = element.replace(" ", "\\ ")
+        exe_cmd = element
         # Create desktop file definition
         desktop_context = {
-            "game_bin_path": get_execute_command(game),
+            "game_bin_path": exe_cmd,
             "game_name": game.name,
             "game_icon_path": os.path.join(game.install_dir, 'support/icon.png')
             }
@@ -219,7 +223,7 @@ def create_applications_file(game):
             Type=Application
             Terminal=false
             StartupNotify=true
-            Exec="{game_bin_path}"
+            Exec=/bin/bash -c "{game_bin_path}"
             Name={game_name}
             Icon={game_icon_path}""".format(**desktop_context)
         if not os.path.isfile(path_to_shortcut):
@@ -246,33 +250,6 @@ def remove_installer(game, installer):
         shutil.rmtree(installer_directory, ignore_errors=True)
     else:
         error_message = "No installer directory is present: {}".format(installer_directory)
-    if Config.get("create_applications_file") and not error_message:
-        path_to_shortcut = os.path.join(APPLICATIONS_DIR, "{}.desktop".format(game.name))
-        exe_cmd_list = get_execute_command(game)
-        for element in exe_cmd_list:
-          element = element.replace(" ", "\\ ")
-        exe_cmd = element
-        # Create desktop file definition
-        desktop_context = {
-            "game_bin_path": exe_cmd,
-            "game_name": game.name,
-            "game_icon_path": os.path.join(game.install_dir, 'support/icon.png')
-            }
-        desktop_definition = """\
-            [Desktop Entry]
-            Type=Application
-            Terminal=false
-            StartupNotify=true
-            Exec=/bin/bash -c "{game_bin_path}"
-            Name={game_name}
-            Icon={game_icon_path}""".format(**desktop_context)
-        if not os.path.isfile(path_to_shortcut):
-            try:
-                with open(path_to_shortcut, 'w+') as desktop_file:
-                    desktop_file.writelines(textwrap.dedent(desktop_definition))
-            except Exception as e:
-                os.remove(path_to_shortcut)
-                error_message = e
     return error_message
 
 
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -215,9 +215,13 @@ def create_applications_file(game):
     error_message = ""
     if Config.get("create_applications_file"):
         path_to_shortcut = os.path.join(APPLICATIONS_DIR, "{}.desktop".format(game.name))
+        exe_cmd_list = get_execute_command(game)
+        for element in exe_cmd_list:
+          element = element.replace(" ", "\\ ")
+        exe_cmd = element
         # Create desktop file definition
         desktop_context = {
-            "game_bin_path": get_execute_command(game),
+            "game_bin_path": exe_cmd,
             "game_name": game.name,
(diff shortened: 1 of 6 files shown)


minigalaxy/installer.py
```diff
@@ -167,6 +170,35 @@ def copy_thumbnail(game):
     return error_message
 
 
+def create_applications_file(game):
+    error_message = ""
+    preferences_switch = Config.get("create_applications_file")
+    if game.platform == "linux" and preferences_switch:
```
makson96, line 176: We can easily support more platforms. Please remove this Linux check to:
"
    if Config.get("create_applications_file"):
"
(lines 175 and 176 combined into one)

minigalaxy/installer.py
```diff
@@ -167,6 +170,35 @@ def copy_thumbnail(game):
...
+        path_to_shortcut = os.path.join(APPLICATIONS_DIR, game.name+".desktop")
```
makson96, line 177: This is really cosmetic, but can you please change it to:
"
path_to_shortcut = os.path.join(APPLICATIONS_DIR, "{}.desktop".format(game.name))
"

minigalaxy/installer.py
```diff
@@ -167,6 +170,35 @@ def copy_thumbnail(game):
...
+        # Create desktop file definition
+        desktop_context = {
+            "game_bin_path": os.path.join(game.install_dir, 'start.sh'),
```
makson96, line 180: We support way more methods to launch the game (for multiple platforms, etc.) please add at the top of the fail (import section)
"
from minigalaxy.laucher import get_execute_command
"

And change this line to:
"
"game_bin_path": get_execute_command(game),
"

minigalaxy/installer.py
```diff
@@ -167,6 +170,35 @@ def copy_thumbnail(game):
...
+            "game_name": game.name,
+            "game_icon_path": os.path.join(game.install_dir, 'support/icon.png')
```
makson96, line 182: Hmm, I will need to check if this icon file exists also on Windows games.

minigalaxy/installer.py
```diff
@@ -167,6 +170,35 @@ def copy_thumbnail(game):
...
+            StartupNotify=true
+            Exec="{game_bin_path}"
+            Name={game_name}
+            Icon={game_icon_path}""".format(**desktop_context)
+        if not os.path.isfile(path_to_shortcut):
+            try:
+                with open(path_to_shortcut, 'w+') as desktop_file:
+                    desktop_file.writelines(textwrap.dedent(desktop_definition))
+            except Exception as e:
+                os.remove(path_to_shortcut)
```
makson96, line 197: I am not sure about this...
If we were unable to write to file, we will probably crash also on remove file attempt.

minigalaxy/installer.py
```diff
@@ -167,6 +170,35 @@ def copy_thumbnail(game):
...
+            "game_bin_path": os.path.join(game.install_dir, 'start.sh'),
+            "game_name": game.name,
+            "game_icon_path": os.path.join(game.install_dir, 'support/icon.png')
+            }
+        desktop_definition = """\
+            [Desktop Entry]
+            Type=Application
+            Terminal=false
+            StartupNotify=true
+            Exec="{game_bin_path}"
```
makson96, line 189: Hmm, I think we will need to alter it a little bit for more complex commands. Maybe:
"
Exec=bash -c "{game_bin_path}"
"
It will require some testing...

minigalaxy/installer.py
```diff
@@ -182,3 +214,5 @@ def uninstall_game(game):
     shutil.rmtree(game.install_dir, ignore_errors=True)
     if os.path.isfile(game.status_file_path):
         os.remove(game.status_file_path)
+    if os.path.isfile(os.path.join(APPLICATIONS_DIR, game.name+".desktop")):
+        os.remove(os.path.join(APPLICATIONS_DIR, game.name+".desktop"))
```
makson96, line 218: The same cosmetic request in both lines as above. ("{}.desktop."format(game.name))

makson96 (changes requested):
Thank you for your work. It looks clean and nice.
We support multiple ways of starting games. Please address review in order to utilize it.

I will also need to perform some testing later.
Great job.

minigalaxy/installer.py
itwasonlyabug: Done!

minigalaxy/installer.py
itwasonlyabug: Done in all places within the file.

minigalaxy/installer.py
itwasonlyabug: Done in all places within the file.

minigalaxy/installer.py
itwasonlyabug: Done, now it uses the function instead of start.sh.

minigalaxy/installer.py
itwasonlyabug: The support/icon.png? Good point, in Heroes of Might and Magic it exists, but it's .ico. I'll check some other games.

Also, for the time being Wine auto-creates the .desktop files and we should skip generating it for Windows games anyway.

minigalaxy/installer.py
itwasonlyabug: I'll do some testing with less permissions to verify.

minigalaxy/installer.py
itwasonlyabug: Not sure about this as bash might not be installed. But then again, the list of supported OSes should have bash by default. I'll test it the way you mentioned and commit it if there are no issues.

minigalaxy/installer.py
itwasonlyabug: Ok, I did some tests and the WINE-generated shortcuts do not work. I'll think how to replace them via minigalaxy.

sharkwouter:
Is this still a draft or can this be reviewed?

minigalaxy/installer.py
itwasonlyabug: Added /bin/bash -c and a loop to properly escape whitespaces in the path.

minigalaxy/installer.py
itwasonlyabug: I created a brand new user and installed some games. There were no permission issues but I'm OK to remove it from the exception if you think that's fine.

Also, I referenced https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html and some StackOverflow posts and I don't see a way that the user will not have permissions for ~/.local/share

sharkwouter:
I found something odd. With Linux games this works just fine, but with Dosbox and wine games the result is the following:

```
[Desktop Entry]
Type=Application
Terminal=false
StartupNotify=true
Exec=/bin/bash -c "exit"
Name=Dragonsphere
Icon=/home/wouter/GOG Games/Dragonsphere/support/icon.png
```

sharkwouter:
Could you call the option in preferences "Create menu shortcuts"? It seems more appropriate for what it's actually doing.

sharkwouter:
I've rebased your branch on master and force pushed to it. All your changes should be intact. They seem to work in the same way as before the rebase.

sharkwouter (changes requested):
I'm quite impressed by this, well done. There are still some minor issues, but this is very close to being usable. Here a short summary of the changes which are required before I'll merge this:

- Change the label in preferences to say "Create menu shortcuts
- Make sure all tests run (preferably also add a test for the create_application_file function, but this is optional)
- Fix the shortcuts for non-native games (Dosbox, Scummvm and Wine) if possible, otherwise prevent shortcuts from being created for these.

itwasonlyabug:
> I found something odd. With Linux games this works just fine, but with Dosbox and wine games the result is the following:
> 
> ```
> [Desktop Entry]
> Type=Application
> Terminal=false
> StartupNotify=true
> Exec=/bin/bash -c "exit"
> Name=Dragonsphere
> Icon=/home/wouter/GOG Games/Dragonsphere/support/icon.png
> ```

I tried with `Heretic` and see the following shortcut created:

```
[Desktop Entry]
Type=Application
Terminal=false
StartupNotify=true
Exec=/bin/bash -c "DOSBOX\dosbox.exe"
Name=Heretic: Shadow of the Serpent Riders
Icon=/home/itwasonlyabug/GOG Games/Heretic Shadow of the Serpent Riders/support/icon.png

Which is still busted, but in a different way, I'll do some more testing.
```

Issue #186, opened 2020-11-11
Full Throttle Remastered doesn't work after installing


berarma:
The game's main data file `game/full.data` is split in two files after installation. The game crashes when launched.

This doesn't happen when installing with the installer downloaded from GOG.

The split files are:
```
-rw-r--r-- 1 berarma staff 4194304000 may 24  2018 full.data.split00
-rw-r--r-- 1 berarma staff 1270343297 may 24  2018 full.data.split01
```

The correct install has this file instead:
```
-rw------- 1 berarma staff 5464647297 nov 11 19:59 full.data
```

Using release 0.9.4.

sharkwouter:
Thanks for the report, I'm really sorry for not noticing this issue sooner. Since this issue has been reported a couple of new versions have been released. Could you try again? Version 1.0.0 fixes a ton of similar issues.

berarma:
Still the same. I've installed the latest version 1.0.0 but in the about window I'm still seeing 0.9.4. I'm pretty sure I have just one instance installed.

I guess the file is splitted to avoid some 4GB file size limit on the zip format and then the game installer joins the files on installation.

I'm a programmer with basic Python knowledge. I've taken a look at the Minigalaxy code and I don't see anywhere that handles joining the files or any processing that needs to be done on installation.

sharkwouter:
Yeah, that's a bug. I forgot to change the version number in the 1.0.0 release.

I'm suprised by this issue, though. I've not seen this before. I'll take a look.

berarma:
I've found `postinst.sh` and `preuninst.sh` scripts in `data/noarch/support/`.

The `postinst.sh` script joins the split files according to the contents of `data/noarch/support/split_files`. The `preuninst.sh` script removes the resulting file before uninstall, I guess because the uninstall script knows nothing about that file created by `postinst.sh`.

The scripts can be called from wherever since they change the working directory to the script directory and use relative paths from there. They call `yad`, a binary inside the support directory that displays a window with a progression bar while the files are joined.

freaky-m0:
I have the same problem with "X4 Foundations". There is also a `postinst.sh` and `preuninst.sh` in `X4 Foundations/support/`. The game only runs after starting `postinst.sh` manually.

sharkwouter:
Thanks for the reply. It's good to have some extra games to test this with.

makson96:
Can anyone affected by this issue check if this branch solves problem:
[...]

PR #346, opened 2021-08-14
Fix 186.

makson96:
This pull request will add running support/posinst.sh during game installation if such a file exists.
This should solve #186


sharkwouter:
Working well! Thanks!

PR #350, opened 2021-08-18
Fix issue with different games names from API and GOG files.

### LINKED ISSUE #349: Game names from goggame.info and API can be different
As you can see on picture below:
names
Directory name is created from name from API, but after game is installed name is taken from installed files. This results in for example constant "update available" status. I am working on cleaver solution.

makson96:
This fixes #349 
Unittest are also added to prevent reintroducing this issue in the future.

### DIFF
--- a/minigalaxy/game.py
+++ b/minigalaxy/game.py
@@ -151,9 +151,11 @@ class Game:
         # Compare names with special characters and capital letters removed
         if self.get_stripped_name().lower() == other.get_stripped_name().lower():
             return True
-        if self.install_dir and other.get_stripped_name() in self.__strip_string(self.install_dir):
+        if self.install_dir and \
+                other.get_install_directory_name() == os.path.basename(os.path.normpath(self.install_dir)):
             return True
-        if other.install_dir and self.get_stripped_name() in self.__strip_string(other.install_dir):
+        if other.install_dir and \
+                self.get_install_directory_name() == os.path.basename(os.path.normpath(other.install_dir)):
             return True
         return False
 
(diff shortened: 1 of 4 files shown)


sharkwouter:
It doesn't seem to fix the issue for Torchlight 2. Left if "Torchlight 2", right is "Torchlight II":

Screenshot from 2021-08-19 08-55-18

The directory is called "Torchlight II". Is this the same issue?

makson96:
Thanks for the check. Local installed games and games from Api are identified as two separate games if both:
- Games IDs are different
- Games names are different

Could you please provide goginfo content, so we can check what is game ID there. Also could you paste here game ID returned from API: in ui/library.py change line 153 from:
for game in retrieved_games:
to
for game in retrieved_games:
    print(game.name)
    print(game.id)

sharkwouter:
Torchlight 2 actually only has a gameinfo file. The contents of which are:

```
Torchlight 2
2.0.0.2
1.25.9.7
```

sharkwouter:
I've also seem a similar mismatch with Pirates! Gold. The gameinfo file contains the following:

```
Pirates! Gold
2.0.0.9
n/a
```

But the directory is called ``Pirates GOLD``. Making it appear twice. This one does not show there is an update, though. Not sure how related this is to this particular fix.

makson96:
Thanks for the info. Those games don't have id in their gameinfo, so they differ in both id and name with games from API.
But I still can provide some tricks to handle this issue.
Check if this pull request fixes issue now.
Proper unit test is added.

sharkwouter:
Thanks! It works well for me now

PR #352, opened 2021-08-28
Some improvements

makson96:
Well, this pull request is in for preparation for fix for 327.
Improving 327 looks to be more challenging than I initially thought, so better to do it in two fazes.
This pull request:
- Fixes installers not being properly kept.
- Pass save_location to finish_function(). It helps to clean some code and make finish_functions more reliable.

P.S. I am still not super happy with get_keep_executable_path(). It would probably fail if multiple installers (for example different versions) of same game are kept. But fixing it totally is outside of my current scope.
Best regards,

### DIFF
--- a/minigalaxy/download.py
+++ b/minigalaxy/download.py
@@ -23,7 +23,7 @@ class Download:
     def finish(self):
         if self.__finish_func:
             try:
-                self.__finish_func()
+                self.__finish_func(self.save_location)
             except (FileNotFoundError, BadZipFile):
                 self.cancel()
 
--- a/minigalaxy/download.py
+++ b/minigalaxy/download.py
@@ -2,7 +2,8 @@ from zipfile import BadZipFile
 
 
 class Download:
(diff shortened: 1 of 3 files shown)


sharkwouter:
Thanks for this PR. It looks like an improvement in general, but I'm still seeing an issue with it. If I enable keeping installers, install a game and then uninstall it, trying to install the game again causes the following exception:

```
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
TypeError: __install_game() takes 2 positional arguments but 69 were given
Exception in thread Thread-1085:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
TypeError: __install_game() takes 2 positional arguments but 69 were given
```

makson96:
Thanks for testing. Args for threading should be tuple. This issue should be fixed now.

sharkwouter:
Great, seems to be working like expected now!

PR #355, opened 2021-08-29
Fix freeze for games generating a lot of output

lmeunier:
When using minigalaxy, some games tend to freeze after a few minutes.
When running them manually, from a terminal, I can't reproduce the
freeze. This is the case, at least, for "Darkest Dungeon" and "Slay the
Spire".
This issue seems to be related to how the `subprocess.Popen` call is
done when launching the game. `stdout` and `stderr` are set to
`subprocess.PIPE`, but `subprocess.PIPE.communicate` is never called and
the child process hangs when a buffer is filled.
Removing `stdout` and `stderr` args from the Popen call seems to fix
this issue. When not specified, stdout and stderr of child process are
attached to the parent process, and games logs are now displayed in the
terminal.

### DIFF
--- a/minigalaxy/launcher.py
+++ b/minigalaxy/launcher.py
@@ -30,6 +30,8 @@ def start_game(game):
         error_message, process = run_game_subprocess(game)
     if not error_message:
         error_message = check_if_game_started_correctly(process, game)
+    if not error_message:
+        send_game_output_to_stdout(process)
     if error_message:
         print(_("Failed to start {}:").format(game.name))
         print(error_message)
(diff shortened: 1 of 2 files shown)


sharkwouter:
Good catch, I wasn't expecting that would be an issue. The proposed solution probably breaks our error reporting a little bit. We do use communicate, but only when something goes wrong. The code for that is here: https://github.com/sharkwouter/minigalaxy/blob/1c53a0e6d3e8bba30a12fa4bd9592f61b2fd1d53/minigalaxy/launcher.py#L202

Are there any other ways to solve this without causing breakage there?

lmeunier:
You were right, the proposed solution break the error reporting (sorry about that, I'm not yet used to the code base).
I've just pushed a new commit. Error reporting should still work as expected now.

makson96:
Some general notes:
1. Please fix failing test.
2. Could we please keep error_message always as str type.
3. Please keep current convention for start_game() function. Instead of adding new code to this function, please create separate function which will be run "if not error_message". Please add single return error_message at the end of this function.

(I haven't tested it yet, but rest of the code looks great. Thanks!)

makson96:
```
    # Set the error message to what's been received in stdout if not yet set
    if error_message:
        stdout, _ = process.communicate()
        error_message = stdout.decode("utf-8")
```
Why are attaching stdout as error_message without stderr?

lmeunier:
I changed the call to `subprocess.Popen` to combine both stdout and stderr in one stream: stdout.
```
            stdout=subprocess.PIPE,
            stderr=subprocess.STDOUT,
```

This way, all game outputs are merged in one stream. Usually, it is useful to have both outputs to help debug a game that refuse to start. 
Furthermore, it's easier to consume the output (to avoid child process to hang) as you have to consume only one stream.

makson96:
Thanks for the explanation, it makes sense.

Code looks good. I would like to do some manual testing before merge, but I don't know when I will find time for it...

PR #356, opened 2021-08-29
Fix crash when auth.gog.com is down

### LINKED ISSUE #354: minigalaxy couldn't start because auth.gog.com is down
Hello,

Tried to start minigalaxy and got the following Python stack trace:
```
$ MG_DEBUG=1 ./bin/minigalaxy 
  File "/home/laurent/Devel/github/minigalaxy/./bin/minigalaxy", line 64, in <module>
    main()
  File "/home/laurent/Devel/github/minigalaxy/./bin/minigalaxy", line 58, in main
    window = Window(APPLICATION_NAME)
  File "/home/laurent/Devel/github/minigalaxy/minigalaxy/ui/window.py", line 70, in __init__
    self.__authenticate()
  File "/home/laurent/Devel/github/minigalaxy/minigalaxy/ui/window.py", line 172, in __authenticate
    authenticated = self.api.authenticate(refresh_token=token, login_code=url)
  File "/home/laurent/Devel/github/minigalaxy/minigalaxy/api.py", line 29, in authenticate
    return self.__refresh_token(refresh_token)
  File "/home/laurent/Devel/github/minigalaxy/minigalaxy/api.py", line 43, in __refresh_token
    response_token = self.__get_refresh_token(params)
  File "/home/laurent/Devel/github/minigalaxy/minigalaxy/api.py", line 62, in __get_refresh_token
    response_params = response.json()
  File "/usr/lib/python3.9/site-packages/requests/models.py", line 900, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib64/python3.9/site-packages/simplejson/__init__.py", line 525, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.9/site-packages/simplejson/decoder.py", line 370, in decode
    obj, end = self.raw_decode(s)
  File "/usr/lib64/python3.9/site-packages/simplejson/decoder.py", line 400, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

It seems that auth.gog.com is currently down, I've random 503/504 errors when I try to access https://auth.gog.com/ in my browser.

Having auth.gog.com down shouldn't prevent minigalaxy from starting. I should still be able to play installed games.

lmeunier:
Fix #354

### DIFF
--- a/minigalaxy/ui/window.py
+++ b/minigalaxy/ui/window.py
@@ -67,12 +67,14 @@ class Window(Gtk.ApplicationWindow):
             os.makedirs(THUMBNAIL_DIR, mode=0o755)
 
         # Interact with the API
-        try:
-            self.__authenticate()
-        except Exception as e:
-            print(e)
-            self.offline = True
-        self.HeaderBar.set_subtitle(self.api.get_user_info())
+        self.offline = not self.api.can_connect()
+        if not self.offline:
+            try:
+                self.__authenticate()
+                self.HeaderBar.set_subtitle(self.api.get_user_info())
+            except Exception as e:
+                print(e)
+                self.offline = True
         self.sync_library()
 
     @Gtk.Template.Callback("filter_library")
--- a/minigalaxy/ui/window.py
+++ b/minigalaxy/ui/window.py
@@ -67,7 +67,11 @@ class Window(Gtk.ApplicationWindow):
             os.makedirs(THUMBNAIL_DIR, mode=0o755)
 
         # Interact with the API
-        self.__authenticate()
+        try:
+            self.__authenticate()
+        except Exception as e:
+            print(e)
+            self.offline = True
         self.HeaderBar.set_subtitle(self.api.get_user_info())
         self.sync_library()
 
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -206,11 +206,15 @@ class Api:
 
     @staticmethod
     def can_connect() -> bool:
-        url = "https://embed.gog.com"
-        try:
-            SESSION.get(url, timeout=5)
-        except requests.exceptions.ConnectionError:
-            return False
+        urls = [
+            "https://embed.gog.com",
+            "https://auth.gog.com",
+        ]
+        for url in urls:
+            try:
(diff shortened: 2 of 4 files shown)


lmeunier:
This is a naive approach to fix the issue with auth.gog.com. Maybe you could fix this issue with a way better solution than mine.

makson96:
Hi,
Thanks for your support. I think that better place to address this issue is in line 61 in api.py. For example:
```
        try:
            response_params = response.json()
        except Exception as e:
            print(e)
            response_params = {}
```
This will help to address this exception if other function enters this code here.

lmeunier:
I'm not sure about moving the try/except in api.py instead of window.py. Doing this will still display the Login dialog, but with an error message. And the only way to hide/close this dialog is the close button, which will trigger `Gtk.main_quit()` in https://github.com/sharkwouter/minigalaxy/tree/master/minigalaxy/ui/window.py#L180 and close the app.

minigalaxy/ui/window.py
```diff
@@ -67,7 +67,11 @@ def __init__(self, name="Minigalaxy"):
             os.makedirs(THUMBNAIL_DIR, mode=0o755)
 
         # Interact with the API
-        self.__authenticate()
+        try:
+            self.__authenticate()
+        except Exception as e:
+            print(e)
+            self.offline = True
         self.HeaderBar.set_subtitle(self.api.get_user_info())
```
sharkwouter, line 75: Wouldn't this cause an exception in this case?

sharkwouter:
Thanks for the PR. I don't think window.py is a bad place to have this check, since that is where we keep track of the offline mode. It would be good if we had a test for this, though, to made sure there are no other exceptions when the authentication fails.

minigalaxy/ui/window.py
lmeunier: No, I don't think. From what I saw, the Login dialog stay open as long as the authentication is not successful.

sharkwouter:
I'm just going to merge this, because I've not seen any problems with it and it looks good. Thanks for the contribution!

PR #359, opened 2021-09-04
Update the Finnish translation

jonnelafin:


sharkwouter:
Thanks, I really appreciate the update! It is a little bit hard to merge for me like this, though. Would it be possible to make a new PR where you just commit the ``data/po/fi.po`` file?

jonnelafin:
Certainly!
My bad, seems like I accidentally saved all of the file at the same time with poedit or something

Issue #362, opened 2021-09-12
minigalaxy failed to install Pathfinder: WOTR


rilian-la-te:
Steps to reproduce:
1. Download latest master of Minigalaxy (1.0.2 works fine)
2. Try to install Pathfinder:WOTR from GoG
3. See freeze (tried waiting for 24 hours, no change)

Game downloaded correctly and can be manually innoextracted and played. Or even wineinstalled and played.

sharkwouter:
Do you see any output if you launch Minigalaxy from the terminal and try to install this game?

rilian-la-te:
It finished on Resuming download <name of sixth part of the game.bin>. And I do not see this message with .exe file name.

rilian-la-te:
If I remove all downloaded file, then I see only usual terminal messages about cannot find proper DLC icon and it is all.

sharkwouter:
Strange, then I'm not quite sure what goes wrong. If anyone else owns this game too, please try it and update this issue.

sharkwouter:
Are there any other games where you see this?

rilian-la-te:
@sharkwouter no.

I see some bugs with this game even in 1.0.2 (like, downloading and installing entire game when DLC requested). This persisits, but installing is also does not work in master. 

Maybe other Wine games will show similar issues. I will try with AoW:Planetfall this evening.

rilian-la-te:
Seems fixed in master)

sharkwouter:
Great! Thanks for testing

PR #364, opened 2021-09-29
Fix execution of installer for multipart downloads.

graag:
Since changes in 70791bd28fc, for multipart download the `finish_func` is not called. Download stays in DOWNLOADING state forever. Probably explains #362.

In this PR:
- Reverse the order of downloads to make sure that the executable part is downloaded last and proper path is passed to `finish_func`.
- Fix progress reporting for restarted downloads where parts were finished previously.


sharkwouter:
Oh wow, that's a good catch. Thanks a lot! I'll need to test this.

sharkwouter:
Okay, I just did a quick test after finding a game which is multi-part. It actually works. Thanks!

PR #365, opened 2021-09-29
Improve handling of installer storage

graag:
There are two issues with current implementation of "do not remove installers". One is that only the exe part of the installer is moved to `keep_dir` which breaks multipart installers. The other is that installer directory is always removed even when it was called from the `keep_dir`. This results in following sequence:
- download and install the game
  - the installer exe is moved to the `keep_dir`
  - the is installer directory is removed
- remove the game
- install the game
   - the installer exe is moved from `keep_dir` to `keep_dir`
   - the installer directory is removed, which is the `keep_dir`
- Remove the game
- The game has to be downloaded again

This PR:
- Handle multipart installers by moving the whole installer directory
- Do not remove installer when it was called from the "keep" directory

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -242,20 +242,51 @@ def create_applications_file(game):
     return error_message
 
 
+def compare_directories(dir1, dir2):
+    files_1 = []
+    files_2 = []
+
+    if os.path.isdir(dir1):
+        files_1 = os.listdir(dir1)
+    if os.path.isdir(dir2):
+        files_2 = os.listdir(dir2)
+
+    if not set(files_1).issubset(set(files_2)):
+        return False
+
+    result = True
+    for f in files_1:
+        if os.path.getsize(os.path.join(dir1, f)) != \
+           os.path.getsize(os.path.join(dir2, f)):
+            result = False
+
+    return result
+
+
 def remove_installer(game, installer):
     error_message = ""
+    installer_directory = os.path.dirname(installer)
+
+    if not os.path.isdir(installer_directory):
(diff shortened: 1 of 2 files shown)


sharkwouter:
I wasn't even aware of this issue, wow. Thanks for figuring it out and coming up with a solution. I will test this later.

I think this really speaks to us needing more tests in around the downloading code, but it's not easy to set that up. If you'd be willing to take a spin at that, you'd be welcome to do so. Otherwise, I'll just review this later and merge it. It's appreciated either way.

sharkwouter:
Would it be possible to rebase this and fix the flake8 errors found by the test?

graag:
I'll fix the flake8 errors and add a test or two shortly.

graag:
Done. Fixed linter warnings and implemented some unittests for `remove_installer`.

graag:
There's still a problem with some games that provide inconsistent `keep_path`. Eg:
`/home/konrad/GOG Games/installer/Age of Wonders 3` vs `/home/konrad/GOG Games/installer/Age of Wonders III`

Starting minigalaxy with AoW3 installed and the clicking uninstall returns the `Age of Wonders III`. Which is not the name of the actual installer dir and leads to a "Download" button.

Restarting minigalaxy after uninstall or clicking uninstall just after downloading and installing the game results in `Age of Wonders 3`. Which **is** the proper installer directory and allows to install the game without re-downloading.

I would leave it for a future PR as at the moment I do not see what's the cause of this error ...

sharkwouter:
Thanks! This seems to work well. The code looks good too

PR #371, opened 2021-09-30
Updated Turkish Translation

fuzunspm:


sharkwouter:
Thanks! That was fast. Could you remove the mo file? Then I'll merge it, since it looks pretty good.

data/po/tr.po
```diff
@@ -513,12 +501,8 @@ msgstr "{} bozuk. Lütfen yeniden indirin."
 #~ msgstr "Ayarlar"
 
 #~ msgctxt "show_fps_tooltip"
-#~ msgid ""
-#~ "Show the framerate while playing games. Only works with AMD and Nvidia "
-#~ "graphics cards."
-#~ msgstr ""
-#~ "Oyun oynarken kare hızını göster. Sadece AMD ve Nvidia ekran kartlarında "
-#~ "çalışır."
+#~ msgid "Show the framerate while playing games. Only works with AMD and Nvidia graphics cards."
```
sharkwouter, line 504: This part was changed, but it's still commented out. Is that intended?

fuzunspm: No, actually I didn't commented that part. Should i uncomment?

sharkwouter: Seems these are commented out everywhere. I think the string doesn't exist anymore, so it should be fine.

PR #374, opened 2021-09-30
Update pl.po

ArturWroblewski:
PL - Update


sharkwouter:
Thanks for the fast update!

Issue #369, opened 2021-09-29
Update translations for 1.1.0 release


sharkwouter:
Currently the date for the release has been set to the 7th of November and the strings on master shouldn't change much anymore. So this would be the perfect time to update some translations.

We currently have the following translations in need of updates:

- [ ] Brazilian Portuguese
- [x] Dutch (@sharkwouter)
- [x] Finnish
- [ ] French
- [x] German
- [ ] Italian
- [ ] Norwegian Bokmål
- [x] Norwegian Nynorsk
- [x] Polish
- [x] Russian
- [x] Simplified Chinese
- [x] Spanish
- [x] Swedish
- [ ] Taiwanese Mandarin
- [x] Turkish
- [x] Ukrainian

I'll be tagging all people who've helped translating Minigalaxy before below. If anyone feels like updating one of these translations, please do. All help is welcome here.

sharkwouter:
Tagging all translators here: @s8321414 @BlindJerobine @fuzunspm @thomasb22 @protheory8 @tim77 @EsdrasTarsis @ArturWroblewski @kimmalmo @LordPilum @juanborda @dummyx @Newbytee @jonnelafin @karaushu @heidiwenger

Now would be a good time for an update if you'd like to.

tim77:
Latest version from `master` in Flathub `beta` channel [available](https://flathub.org/builds/#/builders/32/builds/61623) for testing. Could be safely installed on system without worrying to broke something, but please backup your user data provided by Minigalaxy if you already using Flatpak version. To install Minigalaxy dev pre-1.1.0 version:

```bash
$ flatpak remote-add --user flathub-beta https://flathub.org/beta-repo/flathub-beta.flatpakrepo
$ flatpak install --user flathub-beta io.github.sharkwouter.Minigalaxy
```

To run:

```bash
$ flatpak run io.github.sharkwouter.Minigalaxy//beta
```

sharkwouter:
I forgot to tag @koraynilay for the Italian translation. Sorry about that.

PR #377, opened 2021-10-01
Update sv_SE.po

Newbytee:
Relates to https://github.com/sharkwouter/minigalaxy/issues/369

### DIFF
--- a/data/po/sv_SE.po
+++ b/data/po/sv_SE.po
@@ -8,14 +8,14 @@ msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2021-09-30 11:20+0200\n"
-"PO-Revision-Date: 2020-11-30 08:52+0100\n"
+"PO-Revision-Date: 2021-10-01 12:46+0200\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "Language: sv_SE\n"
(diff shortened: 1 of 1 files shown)


Newbytee:
Side note: Whatever I do, it seems Minigalaxy displays in English, so I can't verify that the translations look good. I've tried setting LC_ALL, LANGUAGE, LANG; I've tried changing the in-app setting to something else; and I've tried settings my entire session to a different language. Regardless, Minigalaxy stays in English. Do you have any ideas?

sharkwouter:
Did you run ``scripts/compile-translations.sh`` afterwards? Otherwise you might have set the language in the preferences to something other than system default. The translation looks good, though.

There are some issues with certain translations not being applied when not using LC_ALL at the moment and the properties dialog doesn't get translated for some reason, though.

sharkwouter:
Thanks a lot, though, I've merged it.

Newbytee:
> Did you run `scripts/compile-translations.sh` afterwards? Otherwise you might have set the language in the preferences to something other than system default. The translation looks good, though.
> 
> There are some issues with certain translations not being applied when not using LC_ALL at the moment and the properties dialog doesn't get translated for some reason, though.

Oh no, I didn't. I suppose that's why then. I recall you removed the output from the version control?

sharkwouter:
Makes sense, I've checked it off

PR #378, opened 2021-10-01
Fix #375

### LINKED ISSUE #375: Image in properties dialog is put at the top instead of the left
I really don't know why this happened, I think it changed after this commit: https://github.com/sharkwouter/minigalaxy/commit/76ae21dd5ced53925d026812f94cf45f4d226917#diff-4620d9229d4b96e6e0951552cb506af353e64efd69d7be360695157c916b7c84

Screenshot from 2021-09-30 11-39-35

makson96:
I noticed that Glade can sometimes remove orientation. We need to be super careful about that.

### DIFF
--- a/data/ui/properties.ui
+++ b/data/ui/properties.ui
@@ -13,6 +13,7 @@
     <child internal-child="vbox">
       <object class="GtkBox">
         <property name="can_focus">False</property>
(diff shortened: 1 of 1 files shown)


sharkwouter:
That works, thanks! We'll probably need a better solution for this in the future, but for now this will work.

PR #380, opened 2021-10-03
Implement icon extraction from windows ico files

graag:
The PR is marked as a Draft as I'm not sure what would be the best way to handle required dependencies.

The idea is to extract a png icon for games that provide only `goggame-*.ico`, a png version would be useful for `.desktop` files.
Current PR does not introduce any new mandatory dependencies and tries two approaches to extracting the icon:
- use python pillow module if available as default extractor,
- use imagemagick as a fallback.

There are two issues I would like to discuss:
- The code could be simplified a lot if we would add `pillow` as mandatory dependency to `requirements.txt`, is that acceptable?
- How to handle error messages? Which should remain as prints to the console and which should be displayed in a popup dialog? What about translations?

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -3,6 +3,8 @@ import shutil
 import subprocess
 import hashlib
 import textwrap
+import glob
+import re
 from minigalaxy.translation import _
 from minigalaxy.launcher import get_execute_command
 from minigalaxy.paths import CACHE_DIR, THUMBNAIL_DIR, APPLICATIONS_DIR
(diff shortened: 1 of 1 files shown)


sharkwouter:
In this case I would show an error message in the terminal, but not as a pop-up. Error messages in the terminal should not be translatable.

Adding Pillow as a dependency would make it so Imagemagick would not be needed, right?

graag:
> In this case I would show an error message in the terminal, but not as a pop-up. Error messages in the terminal should not be translatable.

OK

> Adding Pillow as a dependency would make it so Imagemagick would not be needed, right?

Yes. With Pillow as a dependency the part with imagemagick could be dropped.

However it looks that it would be enough to point to proper .ico file - although KDE could select a better resolution.

sharkwouter:
Oh that would be good. Much simpler too.

sharkwouter:
@graag do you think this could be completed before the 7th of November? It would be nice to include it in the next release.

GB609:
Should we close this now that an alternative solution was implemented?

sharkwouter:
Yeah, that makes sense.

PR #381, opened 2021-10-04
Update no_NN.po

LordPilum:
Relates to #369


sharkwouter:
Thanks for the update!

Issue #139, opened 2020-04-23
Musl compiled Python support


mdgomes:
Hello,

Due to musl differences to glibc handling of locales, the following python locale method functions should not be used and gettext ones should be used:

locale.gettext(msg)
locale.dgettext(domain, msg)
locale.dcgettext(domain, msg, category)
locale.textdomain(domain)
locale.bindtextdomain(domain, dir)

Otherwise the following error occurs:

```
$ minigalaxy 
Traceback (most recent call last):
  File "/usr/bin/minigalaxy", line 66, in <module>
    main()
  File "/usr/bin/minigalaxy", line 57, in main
    from minigalaxy.ui import Window
  File "/usr/lib/python3.8/site-packages/minigalaxy/ui/__init__.py", line 2, in <module>
    from minigalaxy.ui.window import Window
  File "/usr/lib/python3.8/site-packages/minigalaxy/ui/window.py", line 5, in <module>
    from minigalaxy.ui.login import Login
  File "/usr/lib/python3.8/site-packages/minigalaxy/ui/login.py", line 8, in <module>
    from minigalaxy.translation import _
  File "/usr/lib/python3.8/site-packages/minigalaxy/translation.py", line 8, in <module>
    locale.bindtextdomain(TRANSLATION_DOMAIN, LOCALE_DIR)
AttributeError: module 'locale' has no attribute 'bindtextdomain'
```
Commenting "locale.bindtextdomain(TRANSLATION_DOMAIN, LOCALE_DIR)" in minigalaxy/translation.py fixes the issues and allows minigalaxy to launch :)

This change makes it Alpine Linux (and probably other musl distros) ready :)

sharkwouter:
That's a shame. For me when I don't use locale.bindtextdomain, the translations aren't fully loaded. I can add a small check though to make sure it doesn't crash in Alpine.

sharkwouter:
Could you clone master and test again? I've added a workaround.

mdgomes:
It works, many thanks :)

PR #385, opened 2021-10-13
translation.py: call locale.textdomain

### LINKED ISSUE #372: Translations are not being loaded for properties
When opening a properties window for a game, it seems translations are not applied.

zweif:
Closes #372.

Adding a call to 'locale.textdomain' seems to do the trick.
I also added some error handling which might need to be tested (see #139).


sharkwouter:
Thanks! I tested it and it works well for me. I'll merge it

PR #387, opened 2021-10-14
Fix translations not being applied

### LINKED ISSUE #373: Many translations are not being loaded after changing the language in properties
~~In the properties menu this is visible. The list of languages is not translated, even if there are translations available.~~

After setting the language in the properties menu and restarting Minigalaxy. The play button on game times and anything in constants.py are not being translated. There are probably other places where this is a problem.

zweif:
Closes #373.

Not sure if this is the best way.
Moved DEFAULT_CONFIGURATION from constants to config to avoid circular dependencies.

Some further notes/questions:

https://github.com/sharkwouter/minigalaxy/blob/bf7a7f3c3a4b6ab5487374cdb10e7391e6bdeb7d/minigalaxy/translation.py#L6-L9
and
https://github.com/sharkwouter/minigalaxy/blob/bf7a7f3c3a4b6ab5487374cdb10e7391e6bdeb7d/minigalaxy/ui/gametile.py#L44-L52
might not be needed.

https://github.com/sharkwouter/minigalaxy/blob/bf7a7f3c3a4b6ab5487374cdb10e7391e6bdeb7d/minigalaxy/ui/window.py#L36
should be locale.Error?

### DIFF
--- a/minigalaxy/config.py
+++ b/minigalaxy/config.py
@@ -2,8 +2,23 @@ import os
 import threading
 import json
 import time
-from minigalaxy.paths import CONFIG_DIR, CONFIG_FILE_PATH
-from minigalaxy.constants import DEFAULT_CONFIGURATION
+from minigalaxy.paths import CONFIG_DIR, CONFIG_FILE_PATH, DEFAULT_INSTALL_DIR
+
+# The default values for new configuration files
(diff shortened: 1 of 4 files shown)


zweif:
I looked into this again and updated test_config.py. Seems to work now.
Is this ok?

PR #395, opened 2021-10-29
DLC Download Queuing

flagrama:
Made changes allowing DLC downloads to be queued up.

Should not delete the entire installer directory if other files exist in the directory. This would cause a race condition `remove_installer` in `installer.py` and `prepare_location` in `download_manager.py` so `prepare_location` will now create a placeholder file so that the directory will be much less likely be deleted between preparation and download starting.

The Menu button will not be hidden anymore, instead the Uninstall and Update buttons will be hidden while in a DOWNLOADING state, leaving Properties and DLC still accessible. Cancel button has been made to appear to the left of the Menu button now that they will appear together and would have overlapped previously.

`set_progress` in `gametile.py` has been modified to ensure downloads will update the tile's state back to DOWNLOADING with a progress bar after being left in INSTALLED after installing the previous download.

DLC Download buttons will be disabled after clicking on them.

Current Known Issues:
- ~~Canceling a download will say it is cancelling the download of the game, not the DLC, this is the same as current behaviour, but is more confusing now due to the queuing.~~
   - Cancels all downloads for a game now, a little overkill but makes sense regarding the message box. Can be improved in the future.
- ~~The change in `prepare_locations` may have introduced an issue where downloads may end up corrupt after downloading, though this could also easily be because of closing the application by killing the debugger.~~
   - Just not deleting the directory on install, similar to how cancelling a download doesn't delete the directory.
- ~~DLC Download button will not remain disabled. After a call to `__check_for_update_dlc` the buttons will be returned to their states dictated by `update_gtk_box_for_dlc`~~
- I have no idea what I'm doing with Glade. Hopefully I didn't break anything.

### DIFF
--- a/minigalaxy/download_manager.py
+++ b/minigalaxy/download_manager.py
@@ -3,7 +3,6 @@ import shutil
 import time
 import threading
 import queue
-from os.path import exists
 
 from requests.exceptions import ConnectionError
 from minigalaxy.config import Config
--- a/minigalaxy/download_manager.py
+++ b/minigalaxy/download_manager.py
@@ -3,6 +3,8 @@ import shutil
 import time
 import threading
 import queue
+from os.path import exists
+
 from requests.exceptions import ConnectionError
 from minigalaxy.config import Config
 from minigalaxy.constants import DOWNLOAD_CHUNK_SIZE, MINIMUM_RESUME_SIZE, SESSION
@@ -110,7 +112,8 @@ class __DownloadManger:
             shutil.rmtree(save_location)
             print("{} is a directory. Will remove it, to make place for installer.".format(save_location))
 
-        open(save_location, 'a+').close()
+        if not exists(save_location):
+            open(save_location, 'w').close()
 
     def get_start_point_and_download_mode(self, download):
         # Resume the previous download if possible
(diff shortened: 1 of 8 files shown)


flagrama:
Will work on any issues the unit tests run into if other aspects of the PR seem fine.

flagrama:
This should be good for review, the tests should pass if the `os.remove` is mocked instead of `shutil.rmtree`. Though if cleaning up the download directory is still preferred I'll need to find a away to do it that doesn't suffer from the race condition.

minigalaxy/download_manager.py
```diff
@@ -3,6 +3,8 @@
 import time
 import threading
 import queue
+from os.path import exists
```
sharkwouter, line 6: This import is not used and it making the linting test fail.

sharkwouter (changes requested):
I found one small issue with the code by looking at the test, besides that the code looks okay. I will try to find time to test this a bit later. Thanks for the work!

flagrama:
Okay, I'll take a look at it soon and try to fix that.

flagrama:
Ah, you must have deleted it with DLC already installed and redownloaded it. I'll look into how to fix that.

flagrama:
This is why I hate resolving merge conflicts with GitHub.

flagrama:
Not sure why the test failed
image

makson96:
I haven't found anything suspicious in the code, which could affected that test. I have rerun it and it seems to work.

sharkwouter:
I'm going to merge this then. Sorry for not checking it out again sooner. I just tested it and it does work quite well.

PR #407, opened 2021-11-09
If locale cannot be determined, fallback to en translation

flagrama:
https://docs.python.org/3/library/locale.html#locale.getdefaultlocale
"Except for the code 'C', the language code corresponds to RFC 1766. language
code and encoding may be None if their values cannot be determined."


sharkwouter:
Can you check what happens if you set it to C to make sure it doesn't crash?

sharkwouter:
I just did a small test, it doesn't seem to crash with C. I'll merge it.

PR #413, opened 2021-11-12
cleanup commandline parser

zocker-160:
This PR cleans up the CLI parsing by using `argparse` instead of reimplementing it.

The output it almost the same:

```
$ bin/minigalaxy -h
usage: minigalaxy [-h] [--reset] [-v]

A simple GOG Linux client

optional arguments:
  -h, --help     show this help message and exit
  --reset        reset the configuration of Minigalaxy
  -v, --version  show program's version number and exit
  ```
```
$ bin/minigalaxy -v
1.1.0
```
```
$ bin/minigalaxy --fail
usage: minigalaxy [-h] [--reset] [-v]
minigalaxy: error: unrecognized arguments: --fail
```

Using `argparse` is cleaner, easier to understand, easier to extend and less error prone.

### DIFF
--- a/bin/minigalaxy
+++ b/bin/minigalaxy
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 import sys
 import os
-import getopt
+import argparse
 import shutil
 from os.path import realpath, dirname, normpath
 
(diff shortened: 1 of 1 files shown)


sharkwouter:
Oh cool, thanks! That looks quite a bit cleaner. Is there a way to test this in a unit test?

zocker-160:
While I am all for unit tests, I am not quite sure how a meaningful test would look like in this case.

We could parse stdout after calling it with a given parameter, but I am not sure how useful that actually is.

Maybe you have an idea?

sharkwouter:
Okay, makes sense. thanks for making this change. I've merged it.

PR #414, opened 2021-11-12
Handle LANGUAGE the same as Gtk

flagrama:


sharkwouter:
@flagrama is this still relevant? With drafts I tend to assume the work is not ready for review yet, but this has been here for a while.

flagrama:
Honestly I'm not sure, I haven't really paid attention to development efforts to know if it's still a problem.

I think it was a draft because I wanted to get your opinion on it in the discord, but I don't remember what the outcome of that was either.

I'd think if there are still issues with minigalaxy by default displaying the right language for certain ui elements, this is probably the most consistent to how the backend gtk library code does it (IIRC), but if you have/had a cleaner way that manages to work that's probably fine too.

sharkwouter:
Honestly, I really doubt the original problems were completely fixed. #530 was reported yesterday for example. We also have some super ugly code at the beginning of most classes which use Gtk like here:
https://github.com/sharkwouter/minigalaxy/blob/257cd94ec28d15b169683a7264ec1556d972d2d4/minigalaxy/ui/window.py#L30-39

I would like to be able to get rid of that and have this work, but I don't remember anymore which problem was being solved here specifically and if it would allow for changes in other places.

flagrama:
This code is loosely based on [this code in glib](https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gcharset.c#L680). I'm pretty sure this is what the properties set in `.ui` files use to decide the set language.

That seems to be where the discrepancy comes from. In #530 the title text is [set in Python code](https://github.com/flagrama/minigalaxy/blob/8a95d90bf007a442f998059bc3cdecb168c08abb/minigalaxy/ui/about.py#L13) while all the other text is [set using properties in the .ui file](https://github.com/flagrama/minigalaxy/blob/8a95d90bf007a442f998059bc3cdecb168c08abb/data/ui/about.ui#L6-L38). Since the python code for getting language/locale doesn't follow the same rules as gtk does using glib you end up with these differences in which string is chosen.

While this code worked at the time, it also prevented you from changing the language when using the `LANGUAGE` envvar which wasn't very user friendly, and I'm also not entirely sure that it follows the logic glib uses closely enough. Maybe it's possible to just set the `LANGUAGE` envvar during translation setup to get them both using the same translation files.

I don't really have the time to fix it myself currently, so I'm going to close this draft, but if anyone else wants to use it as reference to fix the problem that's fine.

sharkwouter:
Okay, thanks for the update! I'll probably not get to this one for a while either, since I'm in the progress of refactoring some code to be more maintainable, testable and to make more information about GOG games available on disk.

PR #416, opened 2021-11-13
Split game properties and preferences

TotalCaesar659:
This PR splits game properties and preferences to different windows. It resolves two issues:
- when you want to change some preferences, you have to wait for downloading game properties (cover art, description etc) from GOG API;
- only window may become overloaded when adding new properties and preferences in future.

Before:

After:


sharkwouter:
Thanks for this PR, I really appreciate the work you put in here and the code itself is pretty good. The issue I have here is that even though I agree with you on that there is a problem here, namely that the properties menu can take a while to open, which is a bad user experience, I don't think we agree on a solution.

What I would like to see is the information in the properties menu to get cached instead. The description could be stored in the configuration json of the game upon downloading. The image could be stored where thumbnails are saved. What do you think?

makson96:
We could also load window first and add picture later, when it is downloaded. This solution is the case in DLC icons. Their download does not affect drawing rest of the interface:
```
    def get_async_image_dlc_icon(self, icon, title):
        if icon:
            url = "http:{}".format(icon)
            response = Gio.File.new_for_uri(url)
            response.read_async(3, None, self.set_proper_dlc_icon, title)

    def set_proper_dlc_icon(self, source, async_res, user_data):
        response = source.read_finish(async_res)
        pixbuf = GdkPixbuf.Pixbuf.new_from_stream(response)
        self.dlc_dict[user_data][1].set_from_pixbuf(pixbuf)
```

TotalCaesar659:
@sharkwouter I didn't want to complicate for trying to cache the properties picture and text in next PR, but if you want I may to do it here (and add makson96's proposal). Now it's great time to make two windows for future additions to them (I already have a coulple of ideas), or we can face many problems with overloaded joint window (e.g. for people who use big font size like me).

sharkwouter:
Could you rename them to game info and properties instead?

TotalCaesar659:
Made quick rebase and renamed menu points and window names (it's better to change class names in next PR). By the way, I'd name them just "Information" instead "Game info", one-word as other menu points.

sharkwouter:
Thanks! I agree with you on the naming. Lets call it information. I'd prefer to class names to be changed within this PR if that's possible.

TotalCaesar659:
Got it.

PR #420, opened 2021-11-26
Fix properties screen for games without a genre

mareksapota:
Summary: Not all games have the `genre` field populated in gamesdb info.
For example "Jazz Jackrabbit Collection" has an empty list of genres:
https://gamesdb.gog.com/platforms/gog/external_releases/1808582759

Currently trying to open the properties screen for a game without any
listed genres throws an exception.  This PR changes the API part to
accept an empty list of genres as valid and changes the UI to display
`unknown` if no genres are available.

Test plan: Added a unit test.
```
pytest tests/
```

UI screenshot after changes: jazz-genre

### DIFF
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -259,8 +259,9 @@ class Api:
             for summary_key in response_json["game"]["summary"]:
                 gamesdb_dict["summary"][summary_key] = response_json["game"]["summary"][summary_key]
             gamesdb_dict["genre"] = {}
-            for genre_key in response_json["game"]["genres"][0]["name"]:
-                gamesdb_dict["genre"][genre_key] = response_json["game"]["genres"][0]["name"][genre_key]
+            if len(response_json["game"]["genres"]) > 0:
+                for genre_key in response_json["game"]["genres"][0]["name"]:
+                    gamesdb_dict["genre"][genre_key] = response_json["game"]["genres"][0]["name"][genre_key]
         else:
             gamesdb_dict["summary"] = {}
             gamesdb_dict["genre"] = {}
--- a/minigalaxy/ui/properties.py
+++ b/minigalaxy/ui/properties.py
@@ -130,11 +130,14 @@ class Properties(Gtk.Dialog):
                 description = "{}...".format(self.gamesdb_info["summary"][desc_lang][:description_len])
             else:
                 description = self.gamesdb_info["summary"][desc_lang]
-            genre_lang = "*"
-            for genre_key in self.gamesdb_info["genre"].keys():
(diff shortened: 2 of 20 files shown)


makson96:
Hi,
This seems to be very important fix. I am not very good at translation. Can we keep changes in translation files in separate pull request so I can check only described fix here?
Best regards,

flagrama:
Like 99% of the translation files is just automatically updating the comments referring to line numbers in gametile, mainly from my PR being merged. Otherwise it is just adding the new string "unknown" used by this PR and it's kind of a part of it.

mareksapota:
> Can we keep changes in translation files in separate pull request so I can check only described fix here?

I can split the PR into two if you'd like.  The PR has 2 commits though, first one with the fixes, second one only with the auto generated translation updates.

minigalaxy/api.py
```diff
@@ -259,8 +259,9 @@ def get_gamesdb_info(self, game: Game) -> dict:
             for summary_key in response_json["game"]["summary"]:
                 gamesdb_dict["summary"][summary_key] = response_json["game"]["summary"][summary_key]
             gamesdb_dict["genre"] = {}
-            for genre_key in response_json["game"]["genres"][0]["name"]:
-                gamesdb_dict["genre"][genre_key] = response_json["game"]["genres"][0]["name"][genre_key]
+            if len(response_json["game"]["genres"]) > 0:
```
makson96, line 262: This can be simpler:
```
if response_json["game"]["genres"]:
```

minigalaxy/ui/properties.py
```diff
@@ -130,11 +130,14 @@ def load_description(self):
                 description = "{}...".format(self.gamesdb_info["summary"][desc_lang][:description_len])
             else:
                 description = self.gamesdb_info["summary"][desc_lang]
-            genre_lang = "*"
-            for genre_key in self.gamesdb_info["genre"].keys():
+            if "*" in self.gamesdb_info["genre"]:
+                genre = self.gamesdb_info["genre"]["*"]
+            else:
+                genre = _("unknown")
```
makson96, line 136: I would do such a simple if/else in one line:
```
genre = self.gamesdb_info["genre"]["*"] if "*" in self.gamesdb_info["genre"] else _("unknown")
```

makson96 (changes requested):
I have done review.
Thanks for your contribution. This is important fix as it cause application exception. It looks great, but I think that some things can be done simpler. Please check them.

Thanks for new unit test!

I haven't checked translations.
Best regards,

sharkwouter:
Thanks for the work on this

PR #429, opened 2021-12-12
Add changes for v1.1.0 to appstream data

flukejones:


sharkwouter:
Thanks! I had forgotten about this when I did the release. I'll make a branch based on the release tag with this commit applied to it for @tim77

Issue #427, opened 2021-12-09
Innoextract does not respect the game language set in Preference


Kzimir:
Hello,

This issue does not affect new games because it's now possible to change the language directly in-game. For old game as "XCom Enemy Unknow : Complete Edition", there is no option in-game to change the language. 

In fact for this game when you install it with Wine, the language must be changed at the beginning in the setup option : 
[https://postimg.cc/VrVqY6C5](https://postimg.cc/VrVqY6C5)
For example, Vampyr and Dragon Age : Origins Ultimate Edition have not this option because the language can be changed directly in game.

If MiniGalaxy extract the game with Innoextract, all languages are extracted and for XCom EU, the game's language by default is Chinese. To change the language, you need : 
```
- Launch the game once.
- Go to ~/Documents/My Games/XCOM - Enemy Unknown/XComGame/Config (in my case)
- Edit "XComEngine.ini" file and change "Language=" line.
```

The good news is that innoextract can list the languages supported by the installer.
```
kasimir@debian-desktop:~/Jeux/XCOM Enemy Unknown/$ innoextract setup_xcom_enemy_unknown_401776_\(19762\).exe --list-languages
Inspecting "XCOM - Enemy Unknown" - setup data version 5.5.7 (unicode)
 - de-DE
 - en-US
 - es-ES
 - fr-FR
 - it-IT
 - ja-JP
 - ko-KR
 - pl-PL
 - ru-RU
 - zh-Hant
Done.
```

There is an option with innoextract to extract only a specific language : `--language ARG`
I extracted XCom game like this : 
`innoextract setup_xcom_enemy_unknown_401776_\(19762\).exe -d XCom --language fr-FR`
and at the 1st launch (after to delete XCom option folder in ~Documents), the game is correctly in French.

For game which have not language option in the setup like Vampyr or Dragon Age Origins, the `--list-languages`command returns only **en_US**

I think we need to create a function which check if the setup has more languages but only for game which have only english in installer['language'] to avoid potential issues. (see [here](https://github.com/sharkwouter/minigalaxy/blob/6d5cb94c14637a37139e7c87cfa08c37f9e2f963/minigalaxy/api.py#L158))

sharkwouter:
Thanks for the really detailed report. Seems implementing this might be a bit harder than I thought, though, since there are more languages available for that game than we currently have in the preferences. We should probably ask the user to choose the language upon install.

makson96:
Thanks for the issue report.
[...]

Issue #428, opened 2021-12-10
Prefer Wine instead of Innoextract for Windows games


Kzimir:
Hello,

I write this message to explain why i think it's preferable to choose Wine instead of Innoextract to install Windows games.

1- It's impossible to install patch with Wine if the game has not installed with it. Because some entries are added in the Wine Registry (Regedit) after installation to localize the install path. 

With "The Witcher 3 : GOTY", the offline setup install the 1.31 version and there is a patch to update the game to 1.32. Firstly, MiniGalaxy does not detect that a patch is available and secondly, if i try to install the patch manually with Wine, i have this error : 
https://postimg.cc/qg2S2KXx

2- Install Fallout 3 or Fallout New Vegas with innoextract prevents playing because the game's launcher think the game is not installed. 
https://postimg.cc/cgnj9tqZ

3- GoG has over 3000 games, and the setup installer can be different between more games. I have 15 games in my GoG library and the innoextract installation from MiniGalaxy works for 14 games... It does not work for Dragon Age : Origins because the installer needs to pass `--gog` argument to the innoextract command + the game is extracted in a folder named "game"
--> So the innoextract function must be rewrite to support this type of installer. And i don't know if there are others type ...

4- See the bug report that i opened yesterday #427 

I don't know if these points can change something but i'm agree that innoextract is more quick and easy than wine to install Windows games but in my case, i think that all these points are critical ... It's the main reason that i use minigalaxy git and modify "Installer.py" to choose Wine instead of Innoextract.

sharkwouter:
Thanks again for the detailed report. While I see the problems, I do disagree with this. Innoextract solves a lot of issues which cannot be solved easily with the wine installer. What I'd prefer to do is have a list of applications which do not work with Innoextract and use th gog option by default.

Kzimir:
So point 1 & 2 will be never fixed... Good luck with that

sharkwouter:
We currently do not update with patches, but by redownloading the installer. GOG does update the installers, so this should work. A user manually applying a patch should not happen.

The Fallout games can be exempt from having to be installed using wine. I do wonder how Steam solves the problems with these games now, they also do not run an installer and instead just download the files.

Kzimir:
Steam and Galaxy don't have this issue when it download the files instead of installer ...
That's why implement this solution can be benefit ...

PR #430, opened 2021-12-12
Prefer Wine instead of Innoextract to install Windows games.

Kzimir:
Fix issue #427 and #428 

To fix the first and second point the issue #428, it's necessary to install the game directly in the game's install_dir instead of in a temp_dir and move it. With the actual solution with Wine, it's impossible to install a patch for "The Witcher 3" (MG does not detect it) and Fallout 3/New Vegas launcher think that the game is not installed.


Kzimir:
Innoextract still better than Wine - Close this PR to do an other.

PR #436, opened 2021-12-16
Add the possibility to use custom Wine instead of Wine system.

Kzimir:
Hi,

With this PR, the user have the possibility to use a custom wine version for their game instead of wine system.

In my case, it's very interesting since i have a 4K monitor and old games do not have high resolution. Xorg is old too and does not set the game to fullscreen if the game's resolution is < desktop's resolution.
This screenshot can explain the issue : https://i.postimg.cc/WzkrZCs4/Capture-d-cran-de-2021-12-16-08-05-30.png

Wine-Wayland (which is still in active development and not upstream) or Proton have patches/solution to fix this kind of issue.
Of course if the game desktop's resolution is the same as game's resolution, you don't have this issue.

For users, it can be interesting to use Proton because it has more patches to increase game's performance which are not implemented in Wine.

To finish my explanation, a video to show you how it works : https://youtu.be/5MxLDiIK6Fw

### DIFF
--- a/minigalaxy/launcher.py
+++ b/minigalaxy/launcher.py
@@ -7,27 +7,22 @@ import glob
 from minigalaxy.translation import _
 
 
-def wine_path(game):
-    custom_path = game.get_info("custom_wine")
-    if custom_path == "":
-        custom_wine_path = shutil.which("wine")
-    else:
-        custom_wine_path = custom_path
-    return custom_wine_path
-
+def get_wine_path(game):
+    custom_wine_path = game.get_info("custom_wine")
+    return custom_wine_path if custom_wine_path else shutil.which("wine")
 
 def config_game(game):
     prefix = os.path.join(game.install_dir, "prefix")
 
     os.environ["WINEPREFIX"] = prefix
-    subprocess.Popen([wine_path(game), 'winecfg'])
+    subprocess.Popen([get_wine_path(game), 'winecfg'])
 
 
 def regedit_game(game):
     prefix = os.path.join(game.install_dir, "prefix")
 
     os.environ["WINEPREFIX"] = prefix
-    subprocess.Popen([wine_path(game), 'regedit'])
+    subprocess.Popen([get_wine_path(game), 'regedit'])
 
 
 def start_game(game):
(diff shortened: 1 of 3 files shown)


minigalaxy/launcher.py
```diff
@@ -7,18 +7,27 @@
...
 
+def wine_path(game):
+    custom_path = game.get_info("custom_wine")
+    if custom_path == "":
+        custom_wine_path = shutil.which("wine")
+    else:
+        custom_wine_path = custom_path
+    return custom_wine_path
+
+
```
makson96, line 18: Function should be probably called:
get_wine_path
And I think that body of this function can be shorter:
custom_wine_path = game.get_info("custom_wine")
return custom_wine_path if custom_wine_path else shutil.which("wine")

minigalaxy/ui/properties.py
```diff
@@ -71,6 +75,16 @@ def ok_pressed(self, button):
         self.parent.parent.filter_library()
         self.destroy()
 
+    @Gtk.Template.Callback("on_button_properties_wine_file_set")
+    def on_button_properties_wine(self, widget):
+        self.game.set_info("custom_wine", str(self.button_properties_wine.get_filename()))
+
+    @Gtk.Template.Callback("on_button_properties_reset_clicked")
+    def on_menu_button_reset(self, widget):
+        if self.game.get_info("custom_wine") is not None:
```
makson96, line 84: Wouldn't be enough?
if self.game.get_info("custom_wine"):

makson96 (commented):
I can see only 2 minor code issues.
Of course unit tests needs to be updated as well.

I have not performed any dynamic testing yet.

I can see that you put a lot of effort into this one and video looks good to me. Great job!

minigalaxy/launcher.py
Kzimir: Thanks, I never knew I could make it so short.

minigalaxy/ui/properties.py
Kzimir: It's enough, thanks :-)

sharkwouter:
Thanks for the work on this PR. I've tested this change now. Would you really us a different wine version per game? I think setting the wine executable to use for all games would probably add more value. The dialog does look a little bit weird, I'll add a screenshot:

Screenshot from 2021-12-17 15-19-39

So for me to consider this one, the following will be needed:
- Confirm that people would want to use this on a per game basis
- Rebase on master
- Fix the linting issue
- Make sure all tests run
- Make the GtkFileChooserButton shows the full path somehow (can be in the tooltip)
- Set the default in the GtkFileChooserButton to the result of ``which wine`` 

If we merge this, we will need to allow to setting the default in preferences as well.

paradisebird:
Hi there,
First off all congratulations and thanks for this project and the effort from the team.
The best scenario Will use the custom wine on per game basis.
Becuase One versoin of proton ge could work great in One game but borked another!
The same can append on wine tweaks as well.
Thanks One more time and keeping the Nice work.
Regards.

sharkwouter:
This one is at the top of my list to fix. For reference, the patch file from the PR can be downloaded from here: https://github.com/sharkwouter/minigalaxy/pull/436.patch

PR #440, opened 2021-12-24
Add progressbar download tooltip with percentage

TotalCaesar659:
This PR shows downloading percentage when hovering mouse over progressbar.


sharkwouter:
Good idea, but make sure you use ``GLib.idle_add`` for this, otherwise it will cause crashes.

TotalCaesar659:
Got it.

sharkwouter:
How do I use this? For me the tooltip does not seem to appear

TotalCaesar659:
Try to hover mouse cursor over game picture first for showing game name and to progressbar after. This is weird  but works for me.

sharkwouter:
Thanks!

PR #442, opened 2021-12-27
Minor README tweaks and reformatted licence

joesmullen:
Hi,

I made a few small changes to the README, including installation instructions for Void Linux, as well as updated installation information for Debian and Ubuntu (since Minigalaxy is in the official repositories for both of them now) and some minor grammar tweaks.

I also copied and pasted the licence text [directly from gnu.org](https://www.gnu.org/licenses/gpl-3.0.txt) so GitHub should correctly recognise it as the GPLv3.

Thanks and keep up the good work. Minigalaxy is a godsend!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,12 +1,11 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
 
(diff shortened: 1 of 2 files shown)


sharkwouter:
Thanks!

PR #443, opened 2021-12-31
Add option to enable/disable check for game updates

TotalCaesar659:
It would be useful when you want to install certain game version and not to update it.
P.S. Looks like this the last PR of 2021, so merry Christmas and happy New Year! Let Minigalaxy reach new heights in 2022.


sharkwouter:
Thanks! This looks good and works well

PR #444, opened 2022-01-04
Remove old general FPS value

TotalCaesar659:
It was a leftover after changing to game-related ones (remake of https://github.com/sharkwouter/minigalaxy/pull/284).

### DIFF
--- a/minigalaxy/launcher.py
+++ b/minigalaxy/launcher.py
@@ -156,8 +156,6 @@ def get_final_resort_exe_cmd(game, files):
 def set_fps_display(game):
     error_message = ""
     # Enable FPS Counter for Nvidia or AMD (Mesa) users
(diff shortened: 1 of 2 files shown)


minigalaxy/launcher.py
```diff
@@ -156,6 +156,8 @@ def get_final_resort_exe_cmd(game, files):
 def set_fps_display(game):
     error_message = ""
     # Enable FPS Counter for Nvidia or AMD (Mesa) users
+    if game.get_info("show_fps") == "":
```
sharkwouter, line 159: This seems unnecessary. In python an empty string is equal to False.

TotalCaesar659: Yeah, I was too careful, looks like it works without it.

sharkwouter:
Thanks!

PR #447, opened 2022-01-19
Use thumbnail cache

TotalCaesar659:
After start, Minigalaxy downloads thumbnails from GOG even they are already there (it just rewrites them). This PR fixes that for saving some traffic and requests to GOG.

### DIFF
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -161,23 +161,27 @@ class GameTile(Gtk.Box):
         download_thread.start()
 
     def load_thumbnail(self):
-        set_result = self.__set_image("")
-        if not set_result:
-            tries = 10
-            performed_try = 0
-            while performed_try < tries:
-                if self.game.image_url and self.game.id:
-                    # Download the thumbnail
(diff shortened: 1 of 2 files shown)


sharkwouter:
I see, that's really surprising to me. Could you fix this in the ``__set_image`` function instead, though? It seems like the bug is there.

TotalCaesar659:
I see that `__set_image` copies thumbnail for installed games to `thumbnail.jpg` in install location. Is it really necessary? It looks strange cause thumbnails in cache are already used for all games.

sharkwouter:
The reason is that if there is no internet connection, Minigalaxy does not know game IDs. So there is no way to figure out which thumbnail fits to which game in that case. That's why each game needs to have the thumbnail in the game directory.

TotalCaesar659:
Ah, got it.

TotalCaesar659:
@sharkwouter Do you have any plans to merge this PR and other cache ones? It would be great to do that before Steam Deck release.

sharkwouter (approved):

PR #457, opened 2022-02-05
Added Spanish from Spain translation

mbarrio:
Basically the existing es.po Spanish translation with some verb tenses and terms changed to better suit the Spanish spoken in Spain. I have also corrected and completed some strings.

### DIFF
--- a/data/po/es_ES.po
+++ b/data/po/es_ES.po
@@ -1,0 +1,594 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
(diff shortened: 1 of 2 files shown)


sharkwouter:
Thanks for the PR. This seems to be good work. I'm a little bit conflicted about the "Spanish from Spain" here as I would expect Spanish to be that unless otherwise specified. Do you know where the Spanish in the other file is from?

sharkwouter:
That is totally fair. Thanks for the information!

Could you make one small change? I think it would look nicer if the locale was called ``Spanish (Spain)`` in ``constants.py`` instead of ``Spanish from Spain``.

mbarrio:
> That is totally fair. Thanks for the information!
> 
> Could you make one small change? I think it would look nicer if the locale was called `Spanish (Spain)` in `constants.py` instead of `Spanish from Spain`.

Done ;-)

sharkwouter:
Thanks a lot!

PR #458, opened 2022-02-07
Fix creation of Desktop Entries (menu items)

otaconix:
According to the [FreeDesktop Desktop Entry
Specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html),
if the value under the `Exec` key contains a space (among others), that
value must be quoted.
Moreover, on my Gnome desktop environment, the Desktop Entries generated
by minigalaxy prior to this commit do not appear to be valid (they don't
show up in the applications menu): apparently, one needs to put in an
absolute path, or a bare name. It looks like relative paths aren't
supported.

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -218,7 +218,7 @@ def create_applications_file(game):
         exe_cmd = get_exec_line(game)
         # Create desktop file definition
         desktop_context = {
-            "game_bin_path": exe_cmd,
+            "game_bin_path": os.path.join('"{}"'.format(game.install_dir.replace('"', '\\"')), exe_cmd),
             "game_name": game.name,
             "game_install_dir": game.install_dir,
             "game_icon_path": os.path.join(game.install_dir, 'support/icon.png')


sharkwouter:
Thanks for finding this out and making this PR. It's really appreciated. I think I did find a small issue with it in my short test, though. It does work, but the resulting desktop file looks like this at the moment:

```
wouter@wouter-XPS-15-9500:~/.local/share/applications$ cat Not\ a\ Hero.desktop 
[Desktop Entry]
Type=Application
Terminal=false
StartupNotify=true
Exec="/home/wouter/GOG Games/Not a Hero"/./start.sh
Path=/home/wouter/GOG Games/Not a Hero
Name=Not a Hero
Icon=/home/wouter/GOG Games/Not a Hero/support/icon.png
```

It's a bit strange to have the quotation marks end in the middle of the exec statement.

otaconix:
I agree it's a little strange, but it's allowed, and it's (kinda) intentional on my part. Looking at [`get_exec_line`](https://github.com/sharkwouter/minigalaxy/pull/458/files#diff-05456035cefa0e73620258cd8b75929137b2f2f7f30dc16f032368c875dfc9a6R207-R211), it appears to be possible to have multiple arguments end up in `exe_cmd`. Also, the contents of `exe_cmd` appear to already have space characters escaped. So, by doing it like this, I can end up with something like this:

```ini
Exec="/path/to/GOG Games/MyGame"/./entrypoint.sh someargument\ containing\ spaces
```

Does that make sense? I tested this with the following desktop entry:

```ini
[Desktop Entry]
Type=Application
Terminal=true
Name=Stefan
Exec="/home/psYchotic/space dir/"test.sh contains\ spaces
```

and the following script in `/home/psYchotic/space dir/test.sh`:

```python
#!/usr/bin/env python

import sys

print("Args:")
for arg in sys.argv[1:]:
  print("  - {}".format(arg))

print()
print("<Press Enter to terminate>")
sys.stdin.readline()
```

Out of curiosity: do the desktop entries as they're generated by the current version on `master` work for you? If so, what desktop environment do you use?

Issue #367, opened 2021-09-29
Opening the properties dialog for a game cause a short freeze


sharkwouter:
This is because content is downloaded from the API when this is done. This information is not stored for afaik either.

sharkwouter:
This information could be stored in a game's configuration json when it it downloaded. The image could be stored as well.

PR #460, opened 2022-02-08
Use information cover cache

TotalCaesar659:
Partly fixes https://github.com/sharkwouter/minigalaxy/issues/367.

### DIFF
--- a/minigalaxy/ui/window.py
+++ b/minigalaxy/ui/window.py
@@ -6,7 +6,7 @@ from minigalaxy.ui.preferences import Preferences
 from minigalaxy.ui.about import About
 from minigalaxy.api import Api
 from minigalaxy.config import Config
-from minigalaxy.paths import UI_DIR, LOGO_IMAGE_PATH, THUMBNAIL_DIR
+from minigalaxy.paths import UI_DIR, LOGO_IMAGE_PATH, THUMBNAIL_DIR, COVER_DIR
 from minigalaxy.translation import _
 from minigalaxy.ui.library import Library
 from minigalaxy.ui.gtk import Gtk, Gdk, GdkPixbuf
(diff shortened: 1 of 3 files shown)


sharkwouter:
Thanks for this PR, this is a great feature to have. I did find an issue with it, though, even though it does seem to mostly work.

When opening information for a game which does not have the cover stored yet, I see the following exception pop up in the terminal:

```
Exception in thread Thread-379:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
  File "/home/wouter/Personal/minigalaxy/minigalaxy/download_manager.py", line 86, in __download_file
    result = self.download_operation(download, start_point, download_mode)
  File "/home/wouter/Personal/minigalaxy/minigalaxy/download_manager.py", line 134, in download_operation
    file_size = int(download_request.headers.get('content-length'))
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

minigalaxy/ui/window.py
```diff
@@ -66,6 +66,10 @@ def __init__(self, name="Minigalaxy"):
         if not os.path.exists(THUMBNAIL_DIR):
             os.makedirs(THUMBNAIL_DIR, mode=0o755)
 
+        # Create the covers directory
+        if not os.path.exists(COVER_DIR):
```
sharkwouter, line 70: Could you move the creation of empty directories for thumbnails and covers to a ``make_directories`` function? Flake8 is starting to find this function to complex now.

TotalCaesar659: Fixed.

TotalCaesar659:
Hm, I don't see this exception so I don't know what to do with it, please retest.

jgerrish:
Content-length isn't a required header:

[14.13 Content-Length](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html)

So, maybe just a test for None before casting?

I don't know why it's failing sometimes.  Interrupted connections, GOG proxies running different versions or with different views of the cache, I don't know.  

I saw the behavior a few times myself.

PR #461, opened 2022-02-10
Fix issue 367 https://github.com/sharkwouter/minigalaxy/issues/367

jgerrish:
This adds code to cache cover images to disk, fixiing issue #367.

It also adds new AssetManager and Asset classes to manage assets like
thumbnails and covers.  This simplifies the codebase some and separates interfaces between layers somewhat.

Tests have been added for Asset, some additional mocking needs to be added for network tests to test AssetManager.

I want to add tests to this to test the full AssetManager class, but I need to mock out the requests library.  There are a ton of options for doing this, including several Python libraries.  Before I add another library, I'd appreciate input on your preferred method.

Currently, this commit is failing flake8 for window.py  I think that needs to be fixed in another branch.

This can be integrated into the other PR once that is cleaned up.

Thank you again for your time looking at this and the project.

Thanks,

Josh

### DIFF
--- a/minigalaxy/asset_manager.py
+++ b/minigalaxy/asset_manager.py
@@ -1,0 +1,220 @@
+from datetime import datetime, timedelta
+from enum import Enum
+import os
+from pathlib import PurePath
+import urllib
+from minigalaxy.download import Download
+from minigalaxy.download_manager import DownloadManager
+from minigalaxy.paths import COVER_DIR, THUMBNAIL_DIR
+from minigalaxy.ui.gtk import Gio, GdkPixbuf
+
+
+class ItemNotCachedError(Exception):
+    """Raised when an item is not found or not cached"""
+    pass
+
+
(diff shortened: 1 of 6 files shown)


jgerrish:
Fixed up the unit tests failing and added a comment about the reason.

Thanks again for your time.

Related, there's a single unit test that fails when run alone:

python -m unittest tests/test_ui_window.py

It's missing some mocks that get setup in earlier unit tests and not deleted.  I can add an issue for it and look at it later if you want.

sharkwouter:
Som the problem this PR addresses has been solved now, but the asset manager is an interesting idea. I'm going to try to remove a lot of business logic from the GameTile, Library and Window classes, after that I'll re-evaluate the ideas in this PR.

jgerrish:
No problem.  Thanks again for the great application and have a good week.  Feel free to close this PR if you want.

Issue #392, opened 2021-10-28
Unresponsive while launched game is running


flagrama:
The UI is unresponsive while a launched game is running. This appears to be because of it reading stdout from the launched game forever here:
https://github.com/sharkwouter/minigalaxy/blob/8a95d90bf007a442f998059bc3cdecb168c08abb/minigalaxy/launcher.py#L223-L227
This behaviour seems to have been introduced in the merge of PR #355 as in the commit prior (829da4953423499f199dbf2eb5918cc0c4c5f052) to the merge commit for that PR (ea4f0a0b6a35841b7bfae71ec34d3802aca8fb37) the UI is responsive as expected.

Steps to reproduce:
1. Launch a game
2. Alt-Tab to minigalaxy
3. Use the mouse on the UI

Expected:
- Hovering the mouse over UI elements causes them to respond, clicking the UI fires the associated event for the clicked element.

Actual:
- Hovering the mouse over UI elements, and clicking them does nothing (though it does queue an action that will fire when the launched game is closed)

PR #474, opened 2022-04-07
Fix freeze for games generating a lot of output

lmeunier:
When using minigalaxy, some games tend to freeze after a few minutes.
When running them manually, from a terminal, I can't reproduce the
freeze. This is the case, at least, for "Darkest Dungeon" and "Slay the
Spire".
This issue seems to be related to how the `subprocess.Popen` call is
done when launching the game. `stdout` and `stderr` are set to
`subprocess.PIPE`, but `subprocess.PIPE.communicate` is called only once
if the game is not running after a 3s delay. The subprocess output is
filling an internal buffer and the game hangs when this buffer is full.

Now, stderr is merged in stdout, and a new thread is started to consume
the subprocess stdout and send it to the parent process stdout. A thread
is created in order to avoid blocking the main UI thread.

### DIFF
--- a/minigalaxy/launcher.py
+++ b/minigalaxy/launcher.py
@@ -4,6 +4,7 @@ import shutil
 import re
 import json
 import glob
+import threading
 from minigalaxy.translation import _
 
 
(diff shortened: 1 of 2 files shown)


lmeunier:
This PR is similar to https://github.com/sharkwouter/minigalaxy/pull/355 except that a new thread is created in order to avoid blocking the main UI thread (see: https://github.com/sharkwouter/minigalaxy/issues/392)

sharkwouter:
Thanks!

sharkwouter (approved):

PR #476, opened 2022-04-07
Allow to download/install Windows version for a Linux native game.

Kzimir:
I don't know if you're gonna like this pull request :D 
My PR is based on the Makson's PR #317 

Why i did this ?
- Some native games works better with Windows version and Wine.
- Native games can have an oldest version than Windows like [Ruiner](https://www.gogdb.org/product/1637928515#downloads) or [Pathfinder Kingmaker](https://www.gogdb.org/product/1982293831#downloads)
- And maybe others thing.

I tried and use it since a long time and it works (Download/Installation and DLC download/installation)


minigalaxy/api.py
```diff
@@ -134,7 +137,7 @@ def get_info(self, game: Game) -> dict:
         return response
 
     # This returns a unique download url and a link to the checksum of the download
-    def get_download_info(self, game: Game, operating_system="linux", dlc_installers="") -> dict:
+    def get_download_info(self, game: Game, operating_system="", dlc_installers="") -> dict:
```
sharkwouter, line 140: Why was this change made?

sharkwouter (changes requested):
Thanks for this PR! I like this feature and would like to merge it. It just needs a rebase and I have on tiny nitpick. In the settings for installed games, the option probably shouldn't appear.

minigalaxy/api.py
Kzimir: You force "operating_system" = "linux".

In gametile.py, you have "download_info = self.api.get_download_info(self.game, self.api.get_download_version(self.game))"

but if you set "operating_system" = "linux" in apy.py, you always download the linux version of a game even if "self.api.get_download_version(self.game) = windows"

sharkwouter: Okay, thanks for explaining!

sharkwouter (changes requested):
Okay, I tested it now. It works well. The only issue I have with it at the moment is that 4 tests seem to fail. This should be fixed before merging.

sharkwouter:
This is a requested feature and I will try to get this working before the next release.

PR #478, opened 2022-04-09
DXVK/VKD3D installation for Windows games.

Kzimir:
Hi,

This commit add the possibility to install DXVK/VKD3D for each Windows games. 
- The option will be hidden for Linux games.
- Use a single switch for dxvk/vkd3d-proton. Even if d3d12.dll (from vkd3d-proton) does not need dxgi.dll from dxvk, i prefer to install both in the same time. Most users don't know if the game is dx11, dx12 or both.
- Wine prefix is created before game installation to avoid dxvk/vkd3d installation issues.
- Each time the user clicks on "button_properties_ok" to close Properties window, install_uninstall_{dxvk,vkd3d} functions are called and check if a new version is available. So the Windows game which use dxvk/vkd3d will have always the up-to-date version.


sharkwouter:
Thanks for the work. I'm going to say I'd prefer not to have this PR merged, though, simply because it downloads binaries. The user could install dxvk themselves and point to it in the game settings since we now allow setting a custom wine version.

PR #481, opened 2022-04-13
Fix error in Properties window when wine is not installed

lmeunier:
When wine is not installed `shutil.which()` returns `None` and this exception is raised:
```
  File "/home/laurent/Projects/src/github.com/lmeunier/minigalaxy/minigalaxy/ui/gametile.py", line 134, in show_properties
    properties_window = Properties(self, self.game, self.api)
  File "/home/laurent/Projects/src/github.com/lmeunier/minigalaxy/minigalaxy/ui/properties.py", line 51, in __init__
    self.button_properties_wine.set_filename(shutil.which("wine"))
```

This error prevents the Properties window to be shown.


sharkwouter:
Thanks!

PR #482, opened 2022-04-24
Fixed localized DLC names

TotalCaesar659:
GOG has localized names for some games and DLCs (e.g. [Frostpunk: The Rifts](https://www.gog.com/game/frostpunk_the_rifts)), but it provides them based on user's IP. It looks heterogeneous when you want to use English everywhere. This PR fixes it.
Before:

After:

I thought about making a switcher for that behavior, but I have a doubt that it'll be in demand. @sharkwouter what do you think?

### DIFF
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -129,8 +129,8 @@ class Api:
 
     # Get Extrainfo about a game
     def get_info(self, game: Game) -> dict:
-        request_url = "https://api.gog.com/products/{}?locale=en-US&expand=downloads,expanded_dlcs,description,screenshots,videos," \
-                      "related_products,changelog".format(str(game.id))
+        request_url = "https://api.gog.com/products/{}?locale=en-US&expand=downloads,expanded_dlcs,description,screenshots," \
+                      "videos,related_products,changelog".format(str(game.id))
         response = self.__request(request_url)
         return response
 
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -129,8 +129,8 @@ class Api:
 
     # Get Extrainfo about a game
     def get_info(self, game: Game) -> dict:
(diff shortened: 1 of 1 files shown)


minigalaxy/api.py
```diff
@@ -129,8 +129,8 @@ def get_redirect_url(self) -> str:
 
     # Get Extrainfo about a game
     def get_info(self, game: Game) -> dict:
-        request_url = "https://api.gog.com/products/{}?expand=downloads,expanded_dlcs,description,screenshots,videos," \
-                      "related_products,changelog ".format(str(game.id))
+        request_url = "https://api.gog.com/products/{}?locale=en-US&expand=downloads,expanded_dlcs,description,screenshots," \
```
sharkwouter, line 132: I'd say in this case the default is probably preferred over always setting en-US. Is there a way to get a list from the API of supported locales?

sharkwouter (changes requested):

minigalaxy/api.py
TotalCaesar659: > Is there a way to get a list from the API of supported locales?

I haven't found easy way for it (perhaps more skilled one can). [API document](https://gogapidocs.readthedocs.io/en/latest/galaxy.html) says that `locale` is just query parameter, not database entry.

> I'd say in this case the default is probably preferred over always setting en-US.

Well, Minigalaxy already processes base game library this way: e.g. it shows [Syberia](https://api.gog.com/products/1207658848?locale=en_US&expand=downloads,expanded_dlcs,description,screenshots,videos,related_products,changelog), not [Сибирь](https://api.gog.com/products/1207658848?locale=ru&expand=downloads,expanded_dlcs,description,screenshots,videos,related_products,changelog) (in Russian). So I just got DLCs to the common denominator.
If we want to use localized names, so it should be for both base games and DLCs. But for me this is the work for future PRs.

TotalCaesar659:
@sharkwouter Long time no see, congrats with new release! Could you then merge this?

PR #492, opened 2022-05-14
Add language option for Innoextract installation.

Kzimir:
## Description
Fix issue #427 

Create a function which search if the gog installer has the possibility to extract specific language.

## Checklist
 
 - [X] _CHANGELOG.md_ was updated (**format**: - Change made (thanks to Kzimir))

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -140,8 +140,10 @@ def extract_windows(game, installer, temp_dir):
 
 def extract_by_innoextract(installer, temp_dir):
     err_msg = ""
+    lang = lang_install(installer)
+    print(lang)
     if shutil.which("innoextract"):
-        cmd = ["innoextract", installer, "-d", temp_dir, "--gog"]
+        cmd = ["innoextract", installer, "-d", temp_dir, "--gog", lang]
         stdout, stderr, exitcode = _exe_cmd(cmd)
         if exitcode not in [0]:
             err_msg = _("Innoextract extraction failed.")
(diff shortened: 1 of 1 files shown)


minigalaxy/installer.py
```diff
@@ -334,3 +335,28 @@ def _mv(source_dir, target_dir):
...
+        if not line.startswith(' -'):
+            continue
+        languages.append(line[3:])
+    for lang in languages:
+        if "-" in lang:  # lang must be like "en-US" only.
+            if Config.get("lang") == lang[0:2]:
+                arg = "--language={}".format(lang)
+                break
+            else:
+                arg = "--language=en-US"
```
sharkwouter, line 360: I would leave arg empty if the language isn't found. Then the default should be used, right?

sharkwouter (changes requested):
Thanks for this! This is something we really need. I do think it is a bit risky, though. It could break some things when a language doesn't quite match or partially matches. I would like to see some tests for this.

In this case you would have to get the output of the ``--list-languages`` option for one or more installers and then create some tests with mock the output of ``subprocess.Popen`` in this case. You'd test what the output of the lang_install function is with different configured languages.

minigalaxy/installer.py
Kzimir: No. As i explained in #427, in arg is empty a random language is set.
For example with XCom Enemy Unknow, without "arg", the game has Asiatic language...
That's why i force to use en_US by default.

sharkwouter:
Thanks for this PR. Sorry for taking forever. I've been in between jobs and moved. I should have a bit more time now to work on Minigalaxy, but will do a release first.

PR #494, opened 2022-06-02
Added translation greek

Pyrofanis:
## Description

added the greek translation /data/po/el.po
and added an entry for greek in the section of SUPPORTED_LOCALES in minigalaxy/constants.py

## Checklist
 
 - [ ] _CHANGELOG.md_ was updated (**Greek** (thanks to pyrofanis))

### DIFF
--- a/minigalaxy/constants.py
+++ b/minigalaxy/constants.py
@@ -23,6 +23,7 @@ SUPPORTED_DOWNLOAD_LANGUAGES = [
     ["es", _("Spanish")],
     ["sv", _("Swedish")],
     ["tr", _("Turkish")],
+   
 ]
 
 SUPPORTED_LOCALES = [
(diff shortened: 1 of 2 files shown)


Pyrofanis:
what should i do to resolve those issues.it is my first public pull request so i am a tiny bit lost

minigalaxy/constants.py
```diff
@@ -23,6 +23,7 @@
     ["es", _("Spanish")],
     ["sv", _("Swedish")],
     ["tr", _("Turkish")],
+
```
sharkwouter, line 26: This empty new empty line is causing an issue with the linting test. Could you remove it?

sharkwouter (changes requested):
Thanks for the PR! It looks pretty good to me, it just needs a tiny fix. See my comment.

Would it be possible to add something like ``- Added Greek translation (thanks to Pyrofanis)`` to the ``CHANGELOG.md`` file?

minigalaxy/constants.py
Pyrofanis: yeah thanks i am at work r8 now going to correct it when i get home :)

Pyrofanis:
i am closing it to assign the new corrected pull request sry for any issues

PR #500, opened 2022-06-18
Resolve #389

### LINKED ISSUE #389: Hard WebKit dependency?
WebKit is a pretty heavy dependency, avoided by some systems due to performance and security concerns. Is there no chance of making the WebKit dependency optional, as is the case with similar tools such as `lgogdownloader`?

waltercool:
## Description

- Moved WebkitGTK to load only when UI login is needed. This prevents app crashing if the user have a valid token, without dependency.
- Added a quick tutorial how to do login without using WebkitGTK by using lgogdownloader. Stating this project won't support lgogdownloader in any way, this should be entirely user responsibility.

## Checklist
 

waltercool:
OK took another option to resolve the problem and avoid messing checks.

At first, I was avoiding WebKitGTK to be used if not needed, but tests tried those functions and failed.

Now, I'm preventing in runtime to use Login() if you are not able load that class. Class will fail for people who doesn't have Webkit installed, and provided a message at stdout.

- If you have Webkit, everything should work as intended.
- If you don't have Webkit, but your token is valid, everything works as intended.
- If you don't have Webkit, but your token is invalid, it will load into offline mode.

Let me know if that works.

sharkwouter:
Thanks, I think this is pretty good. I think the instructions are a bit more complex than they have to be, though.

If you'd log in at [this url](https://auth.gog.com/auth?client_id=46899977096215655&redirect_uri=https%3A%2F%2Fembed.gog.com%2Fon_login_success%3Forigin%3Dclient&response_type=code&layout=client2) in any browser. It will redirect to a blank page once you've logged in. In the address bar, copy the code after "code=" and paste that in the config.json.

waltercool:
I agree on that. Will remove that from the README file.

This should be an unsupported feature, and intended as that.

waltercool:
Oh, I'm very sorry! Completely forgot to remove my changes from README, my bad.

sharkwouter:
No worries, I've merged it and made the change

PR #527, opened 2022-09-24
Fix connection error texts

TotalCaesar659:
## Description

Fix copy-pasted connection error texts.

## Checklist
 

sharkwouter:
Thanks! I hadn't noticed, good catch!

Issue #512, opened 2022-07-24
Some Wine games have issues caused by moving game files after install


LeXofLeviafan:
Currently, when installing a Wine game, Minigalaxy appears to be doing the following:
1. Create a folder named `~/.minigalaxy/extract/$ID`
2. Create a wineprefix inside
3. Run installation process via Wine
4. Move the contents of said folder into designated installation path (e.g. `~/Games/GOG/$NAME`)

I'm not sure how necessary it is to do the install in a different folder than the files actually end up, but some games (mostly old ones, e.g. _Return to Krondor_, _Betrayal in Antara_, _Rage of Mages 1/2_, _Escape from Monkey Island_, _Die By the Sword_) refuse to start up due to not finding the files where they're supposed to be (e.g. instead of `Z:\home\$USER\Games\GOG\Return to Krondor\GameData\RtkGame.def` it's looking for `Z:\home\$USER\.minigalaxy\extract\$ID\GameData\RtkGame.def`).

I can think of a way to avoid this issue: add virtual drive into the prefix (e.g. `D:`) pointing at the extraction folder _before the 3rd step (Wine installer)_, then change it to point at the designated installation path at the end of installation process; this way, as far as the game is concerned, its files aren't moved at all. (Though I'm not sure if making the game installation folder a virtual drive is actually valid, as the usual practice seemed to install the game into a folder at the top level of the drive.)

LeXofLeviafan:
> If you install innoextract, this should no longer be a problem.

Unfortunately it only seems to work with games that _don't need to have their install path in system registry_; when I've installed _Soulbringer_ with innoextract, trying to run it only resulted in an error message saying the game is not installed (…and the only way to install it normally as of now is via hacky means such as disabling innoextract usage manually in the application code).

sharkwouter:
Okay, that's bad. There has to be a clean solution to this. Changing the way installations work would be a good start, but the innoextract method should not result in a game that doesn't work either.

LeXofLeviafan:
At the very least it should be made possible for the user to choose installation method (e.g. by offering a global setting to [dis]allow using innoextract when it's set).

…Speaking of which, allowing to choose how the games are run would be good as well (at least by giving `start.sh` higher priority so that Wine games can be wrapped in a custom-made script or something), but that's probably a separate matter :-)

sharkwouter:
[...]

PR #533, opened 2022-09-29
fixed install path when using Wine installer

LeXofLeviafan:
A fix for #512. Includes following changes:

* `extract_by_wine()` now installs the game to `D:\%NAME%\` (instead of `Z:\home\%USER%\.cache\minigalaxy\extract\%ID%\`), ensuring the install location is maintained after game files are moved
* `move_and_overwrite()` now detects installer output dir based on installer type, with options for Linux, Wine and Innoextract (+ removed use of `minigalaxy_game_files/` which is not present anywhere else in the codebase)
* moved installer type decision point for Windows games to `install_game()` (with possibility to override it with an optional argument)
* added detection of unhandled errors during installation process (as without it the respective GUI button is simply left stuck at “installing…” state in such cases), complete with a test
* simplified the use of `remove_installer()` and `postinstaller()`

…Side thought: the code executed in `extract_by_innoextract()` after successful unpack should be changed, as the current version will happily remove any folder named `game/` (or `app/`, but at least that one seems less likely) that happens to be located inside of the game directory.

### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -3,6 +3,7 @@ import shutil
 import subprocess
 import hashlib
 import textwrap
+import traceback
 
 from minigalaxy.game import Game
 from minigalaxy.translation import _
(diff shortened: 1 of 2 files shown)


sharkwouter:
I'm just going to merge this and work with the result. I was going to do this after having gotten further with my refactor, but my personal life has been getting in the way. I'm sorry about it taking so long and thank you for contributing.

PR #539, opened 2022-10-23
debian: add python3-requests to build deps for help2man

stephanlachnit:
This fixes a build error with Minigalaxy: since bin/minigalaxy now imports requests, this package is now a build dependency as help2man needs to run bin/minigalaxy for the man page.


sharkwouter:
Thanks!

PR #541, opened 2022-11-09
d/control: add unrar-free as alternative

stephanlachnit:
unrar is in non-free on Debian, thus Minigalaxy had to move to contrib, which is disabled by default on Debian/Ubuntu systems. However, since unrar-free v0.1.0 it supports RAR5 as well, so it should be pretty much identical (binary is named unrar as well). While unrar-free v0.1.0 is out for some time now, it finally landed in Debian Unstable.


sharkwouter:
Thanks! That's good news!

PR #549, opened 2022-12-02
Hide A Plague Tale Digital Goodies Pack

TotalCaesar659:
## Description

Hide A Plague Tale Digital Goodies Pack from library cause it shows error "A Plague Tale Digital Goodies Pack with id 1581684020 couldn't be installed" when trying to download.

## Checklist
 

sharkwouter:
Thanks!

PR #551, opened 2022-12-28
Link to LGOGDownloader

cweiske:
## Description

Update the README to mention LGOGDownloader, so that people see there are other tools they can use.

## Checklist
 
 -- no, because I do not think that a readme change needs a changelog entry

----
Related: https://github.com/Sude-/lgogdownloader/pull/233


sharkwouter:
Thanks! This seems useful

PR #558, opened 2023-02-19
Move the Gametile menu button alongside the Play button

lmeunier:
Not sure of this pull-request as this is an opinionated change.

The menu button, located at the top right corner of the thumbnail, is barely visible for some games. I think it would be better the have this button alongside of the action button (play/download/...).
Should we keep the previous icon ("applications-system-symbolic") or use the default icon for GtkMenuButton?

Also, the first letter of the action button is now capitalized.

Any comments are welcome (I'm not very confident with Gtk).

Before:
Screenshot from 2023-02-19 16-12-29

After:
Screenshot from 2023-02-19 16-11-17

### DIFF
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -544,7 +544,7 @@ class GameTile(Gtk.Box):
             self.progress_bar.destroy()
 
     def __state_queued(self):
-        self.button.set_label(_("in queue…"))
+        self.button.set_label(_("In queue…"))
         self.button.set_sensitive(False)
         self.image.set_sensitive(False)
         self.menu_button_uninstall.hide()
@@ -553,7 +553,7 @@ class GameTile(Gtk.Box):
         self.__create_progress_bar()
 
     def __state_downloading(self):
-        self.button.set_label(_("downloading…"))
+        self.button.set_label(_("Downloading…"))
         self.button.set_sensitive(False)
         self.image.set_sensitive(False)
         self.menu_button_uninstall.hide()
@@ -564,7 +564,7 @@ class GameTile(Gtk.Box):
         self.progress_bar.show_all()
 
     def __state_installing(self):
-        self.button.set_label(_("installing…"))
+        self.button.set_label(_("Installing…"))
         self.button.set_sensitive(False)
         self.image.set_sensitive(True)
         self.menu_button_uninstall.hide()
@@ -579,7 +579,7 @@ class GameTile(Gtk.Box):
         self.parent.filter_library()
 
     def __state_installed(self):
-        self.button.set_label(_("play"))
+        self.button.set_label(_("Play"))
         self.button.get_style_context().add_class("suggested-action")
         self.button.set_sensitive(True)
         self.image.set_sensitive(True)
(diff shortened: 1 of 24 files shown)


sharkwouter:
Oh nice, that's really cool!

minigalaxy/ui/gametile.py
```diff
@@ -609,7 +614,7 @@ def __state_uninstalling(self):
     def __state_updatable(self):
         self.update_icon.show()
         self.update_icon.set_from_icon_name("emblem-synchronizing", Gtk.IconSize.LARGE_TOOLBAR)
-        self.button.set_label(_("play"))
+        self.button.set_label(_("Play"))
```
sharkwouter, line 617: This change makes it necessary to update all translation files. So we'd either have to leave this change out or update those.

minigalaxy/ui/gametile.py
```diff
@@ -553,7 +556,7 @@ def __state_queued(self):
         self.__create_progress_bar()
 
     def __state_downloading(self):
-        self.button.set_label(_("downloading…"))
+        self.button.set_label(_("Downloading…"))
```
sharkwouter, line 559: This change makes it necessary to update all translation files. So we'd either have to leave this change out or update those.

minigalaxy/ui/gametile.py
```diff
@@ -544,7 +547,7 @@ def __state_installable(self):
             self.progress_bar.destroy()
 
     def __state_queued(self):
-        self.button.set_label(_("in queue…"))
+        self.button.set_label(_("In queue…"))
```
sharkwouter, line 550: This change makes it necessary to update all translation files. So we'd either have to leave this change out or update those.

minigalaxy/ui/gametile.py
```diff
@@ -618,7 +623,7 @@ def __state_updatable(self):
             self.wine_icon.set_margin_left(22)
 
     def __state_updating(self):
-        self.button.set_label(_("updating…"))
+        self.button.set_label(_("Updating…"))
```
sharkwouter, line 626: This change makes it necessary to update all translation files. So we'd either have to leave this change out or update those.

sharkwouter (changes requested):
I really like this, it looks much better. Thanks for doing this!

I had one comment on the changes, which applies in a couple of places, so I made sure to make all of them. Besides that, could you update CHANGELOG.md as well? Then I'll merge it.

lmeunier:
In order to update all translation files, all I have to do is run the script `./scripts/update-translation-files.sh`?

Should I also change the `gametilelist.ui/.py` files? To be honest, I was not aware that the game tiles could be displayed as a list, I just discovered this feature :)

sharkwouter:
I'd say the best way to update the translations would actually be to go into the `.po` files and replace the translations for these words which are already there with the capitalized ones. Here would be an example: https://github.com/sharkwouter/minigalaxy/blob/766cc7c7071fdd4759c38991114a9ffe5da7646d/data/po/cs_CZ.po#L626

Just running the update script will not yield the result we want, as it will make new entries and the old ones will be ignored.

lmeunier:
Translations for `msgid "Play"` have been added and some translations for `msgid "Install"` have been fixed.

sharkwouter:
I've cherry-picked the commits which did not involve the updates to the strings and translation files. Thanks for this change! I don't think I can merge the rest of this right now. Maybe it would be better to have a separate PR for that part.

lmeunier:
> Maybe it would be better to have a separate PR for that part.

I agree with you. I'll prepare a new PR with only the Capitalization of the first letter.
I think we can close this PR now.

sharkwouter:
Cool, thanks again for the contribution!

PR #562, opened 2023-03-07
Fix progress bar position in gametiles

lmeunier:
## Description

Since b8894b0, the download progress bar was positioned at the bottom of gametiles (below buttons).

Screenshot from 2023-03-07 15-59-09

This PR fixes this issue and load CSS data from `style.css` for the whole Gtk Window (previously the CSS was loaded only for gametiles).

Screenshot from 2023-03-07 16-59-55

## Checklist
 
 - [ ] ~~_CHANGELOG.md_ was updated~~ (not needed)


sharkwouter:
Thanks!

PR #565, opened 2023-04-03
updated greek translation

Pyrofanis:
## Description
Updated Greek translation
also i found an issue. i don’t thing the Greek translation is being compiled while being packaged 
since it doesn’t show up in neither the flatpak nor the debian package as shown in the pictures below 
Debian:
εικόνα
Flatpak:
εικόνα

Local build with compiled translations :
εικόνα
εικόνα

System Info:
εικόνα
Locale Gen:
εικόνα
Default Locale:
εικόνα

### DIFF
--- a/data/po/el.po
+++ b/data/po/el.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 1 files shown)


sharkwouter:
Thanks! I'll look into why it's not being picked up. The build script should use it.

PR #571, opened 2023-06-22
Add notification popup on game installed

orende:
## Description

Shows a native notification popup when games have been successfully downloaded and installed. Partially addresses #566 .

## Checklist
 
 - [x] _CHANGELOG.md_ was updated (**format**: - Change made (thanks to github_username))

PR #572, opened 2023-06-23
Remove popup.close() call in gametilelist.py

orende:
## Description

Removes the call to the close method as it causes the popup to immediately close after showing on the Gnome DE. This changes was accidentally left out of #571.

## Checklist
 

sharkwouter:
Thanks for the fix!

Issue #543, opened 2022-11-11
[Feature Request] User Categories for GOG games


dtantono:
With big libraries, especially with the games from GOG, it can be really hard to browse the library and Ability to create and manage categories like Steams collections is very useful. What do you think?

sharkwouter:
That's a good idea, thanks! Will see if I manage to get to this at some point. PRs for this would be welcome too.

orende:
@dtantono I'm interested in implementing something like this as I could use this feature as well. But just to clarify what you mean, did you imagine something like:

1. selecting multiple genres and then getting a filtered list of games _or_
2. creating named groups of games (similar to Steam's collections) and filtering the list of games with them?

dtantono:
I think basic named group collection like Steam will be nice

sharkwouter:
I can elaborate a little bit on this one. Each game has categories already assigned to it in the GOG API. I think being able to search for them might be nice. I'm not sure if Gnome supports search suggestions.

orende:
Right, I'll start off simple with just getting game categories from GOG and adding filtering functionality using that data, and once that's done and merged I can start looking into replicating something like Steam's dynamic collections based on categories.

As for search suggestions, it seems there's a auxiliary object named Gtk.EntryCompletion that can be used in conjunction with the Gtk.Entry component to enable input suggestions.

sharkwouter:
@orende that sounds good to me. I would save the categories in Game objects. Good find on the search suggestions.

I would say something like Steam's dynamic collections should be a separate issue. I'm not exactly sure what that would look like, though.

orende:
@dtantono If you use the master branch version of the client, you can try this feature out now.

PR #573, opened 2023-07-03
Feature: Filter games by category

orende:
## Description

This PR addresses issue #543.

The PR adds a dialog that can be used to filter the games library based on game categories. Enabling one category filters shows only games matching the selected category, while enabling multiple category filters shows games matching any of the selected categories.
The games and associated categories are stored on disk in the user's cache directory to allow for filtering in offline mode. Once the client goes online again, the file will be updated with any new games and categories found.

Note that this solution uses the `category` field as returned by the call to `embed.gog.com/account/getFilteredProducts` and not the `genres` field as returned by the call to `gamesdb.gog.com/platforms/gog/external_releases`. This is because the former requires only a single API call, while the latter initially requires one API call per game in the library. 

Searching for categories, as proposed in the issue, was left out as the total number of categories was low (8 in total).

The PR also replaces all print statements with the logger library for ease of debugging.

## Checklist
 

data/ui/application.ui
```diff
@@ -141,6 +141,27 @@
...
+            <property name="can_focus">True</property>
+            <property name="receives_default">False</property>
+            <property name="tooltip_text" translatable="yes" context="category_filter" comments="Tooltip for category filter button">Category filters</property>
+            <property name="valign">center</property>
+            <signal name="clicked" handler="on_menu_category_filter_clicked" swapped="no"/>
+            <child>
+              <object class="GtkImage">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="icon_name">filter-symbolic</property>
```
sharkwouter, line 156: This icon is not included into Ubuntu 22.04 and Debian 12 by default, so we'll need to use another icon.

minigalaxy/ui/categoryfilters.py
```diff
@@ -0,0 +1,60 @@
...
+from minigalaxy.ui.gtk import Gtk
+
+
+@Gtk.Template.from_file(os.path.join(UI_DIR, "categoryfilters.ui"))
+class CategoryFilters(Gtk.Dialog):
+    __gtype_name__ = "CategoryFilters"
+
+    genre_filtering_grid = Gtk.Template.Child()
+    filter_dict = {}
+    categories = [
```
sharkwouter, line 16: Are these the standard categories for GOG? Do you have a source for these?

sharkwouter (changes requested):

sharkwouter:
Looks good, thanks for this! I've found the icon does not load, otherwise I'd merge it.

data/ui/application.ui
orende: Okay, I'll change it. Is there a list somewhere of available icons?

minigalaxy/ui/categoryfilters.py
orende: The source for these categories is the list in the lower left of this GOG store page: https://www.gog.com/en/games

image

data/ui/application.ui
orende: I've replaced the icon with `view-conceal-symbolic` which I found in the `/usr/share/icons/Adwaita` folder (Adwaita being a Gnome theme, so hopefully that icon should be included in stock Ubuntu and Debian). Let me know if that one works.

sharkwouter:
Thanks! That works!

PR #576, opened 2023-07-23
Use threads to speed up connectivity check

orende:
## Description

Adds a ThreadPoolExecutor to the api.can_connect function in order to send the requests to embed.gog.com and auth.gog.com from separate threads. This lowers the worst case execution time from 10 seconds to 5 seconds, which is noticeable on slow connections.

## Checklist
 

sharkwouter:
Nice, that makes more of a difference than I was expecting. Thanks!

PR #577, opened 2023-07-27
Check disk space in the nearest existing directory

### LINKED ISSUE #448: Download error: Not enough disk space to install game
I have this error when I'm trying to download Witcher III GOTY. It stays in `in queue` state for some time and then throws this error.

I checked the disk and I have around 760 GB of free space and didn't saw this issue with any other game.

I'm using the flatpak version of Minigalaxy.

Unrud:
## Description

Check the available disk space in the nearest existing directory, instead of the parent directory. Fixes #448

## Checklist
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Add notifications on successful download and installation of games (thanks to orende)
 - Add category filtering dialog for game library (thanks to orende)
 - Parallelize api.can_connect function with threads, futures (thanks to orende)
+- Fix available disk space being checked in parent directory (thanks to Unrud)
 
 **1.2.2**
 - Fix progress bar not showing up for downloads
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -244,6 +244,7 @@ class GameTile(Gtk.Box):
 
     def __download(self, download_info, download_type, finish_func, cancel_to_state):  # noqa: C901
         download_success = True
+        self.game.set_install_dir(self.config.install_dir)
         GLib.idle_add(self.update_to_state, State.QUEUED)
 
         # Need to update the config with DownloadType metadata
@@ -295,7 +296,7 @@ class GameTile(Gtk.Box):
             download_files.insert(0, download)
         self.download_list.extend(download_files)
 
-        if check_diskspace(total_file_size, self.config.install_dir):
+        if check_diskspace(total_file_size, self.game.install_dir):
             self.download_manager.download(download_files)
             ds_msg_title = ""
             ds_msg_text = ""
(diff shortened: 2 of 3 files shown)


minigalaxy/installer.py
```diff
@@ -16,7 +16,14 @@ def get_available_disk_space(location):
     symlinks to disks with sufficient space are correctly measured. Note this is
     a linux-specific command."""
     absolute_location = os.path.realpath(location)
-    disk_status = os.statvfs(os.path.dirname(absolute_location))
+    while True:
```
sharkwouter, line 19: I would write this in a slightly different way to make sure we're not going up all the way to `/`, which could be a different filesystem.

Something like this should work:
```
if os.path.isdir(absolute_location):
    disk_status = os.statvfs(absolute_location)
else:
    disk_status = os.statvfs(os.path.dirname(absolute_location))
```

I might be missing something, though. Let me know if that's the case, then I'll merge this as is.

sharkwouter (changes requested):
Thanks a lot for figuring this one out and suggesting a fix, that's really appreciated! I did leave one comment. Thanks for contributing.

minigalaxy/installer.py
Unrud: > I would write this in a slightly different way to make sure we're not going up all the way to `/`, which could be a different filesystem.

I don't understand the concern. We go only up to `/` if no part of the path actually exists. Any part of the path could be on a different filesystem. But we never go there, because we stop at the first existing node.

> I might be missing something, though. Let me know if that's the case, then I'll merge this as is.

I wasn't sure if it's always guaranteed that the first parent directory exists. I guess the current code makes the assumptions and it wouldn't cause any new problems.

sharkwouter: Actually, I did a quick test just now and it does not really matter if os.statvfs is used on a directory or a file, the output should be the same.

So instead of this change, we can just change the original line 19 to not use dirname. That should solve the issue.

Could you do that and confirm that this solves the issue for you?

sharkwouter: To explain my thinking here. We need statvfs information from where we want to write. Right now we get information from one directory higher than that, which can be a different file system (we know for sure it is on Flatpak). I'm worried going up further will just give us wrong information, like right now.

Thanks a lot for investigating this issue and putting work into it :)

Unrud: I have simplified the PR and just always call it in the parent directory. I hope this is more in line with your expectations.

> Actually, I did a quick test just now and it does not really matter if os.statvfs is used on a directory or a file, the output should be the same.

Yes, but it is never called with a file. It is called with either `config.install_dir` or `installer.install_dir`. In the latter case the directory does not exist yet. 

> I'm worried going up further will just give us wrong information, like right now.

But we do not go further. If it is another file system, the path is guaranteed to exist up to this point, otherwise there can be no mount point for another file system.

Unrud:
I approached the issue from another angle and replaced the call of `check_diskspace(..., config.install_dir)` instead.

sharkwouter:
Thanks!

PR #594, opened 2024-01-25
debian/control: add gir1.2-webkit2-4.1 as alternative

stephanlachnit:
Adds `gir1.2-webkit2-4.1` as alternative to `gir1.2-webkit2-4.0` in `debian/control`. This just reflects what is present in [minigalaxy/ui/webkit.py](https://github.com/sharkwouter/minigalaxy/blob/master/minigalaxy/ui/webkit.py).

See also https://salsa.debian.org/games-team/minigalaxy/-/merge_requests/2.


sharkwouter:
Thanks!

jbicha:
@sharkwouter Could you use my version from https://salsa.debian.org/games-team/minigalaxy/-/merge_requests/2 instead?

sharkwouter:
I read through that thread and pushed the change. Thanks @jbicha

PR #598, opened 2024-04-24
Adding Flatpak to README

SuperDuperUser:
## Description

I considered making this an issue, but I figured I could just write the change, and you can decline the pull request if you don't want it. 

I only noticed today that Minigalaxy has what seems to be an [official & maintained release](https://flathub.org/apps/io.github.sharkwouter.Minigalaxy) on Flathub. However, it is not mentioned on the README/website. I added a section under Installation for Flatpak like the other entries. This would help make sure people don't miss this option.

## Checklist
 
### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,14 @@ sudo xbps-install -S minigalaxy
 </pre>
 </details>
 
+<details><summary>Flatpak</summary>
+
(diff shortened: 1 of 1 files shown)


sharkwouter:
Thanks! That's appreciated!

PR #599, opened 2024-04-29
Fix Remove forward slash from desktop filename (Shenzhen I/O)

slowsage:
## Description

Installing Shenzhen I/O throws the following error:
```
  File "/usr/lib/python3.12/site-packages/minigalaxy/installer.py", line 72, in install_game
    error_message = create_applications_file(game)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/minigalaxy/installer.py", line 271, in create_applications_file
    os.remove(path_to_shortcut)
FileNotFoundError: [Errno 2] No such file or directory: '/home/USERNAME/.local/share/applications/SHENZHEN I/O.desktop'

```
`/` is not a valid character for a filename in linux. This PR removes all `/` from the filename part of the path.
## Checklist
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 **1.3.0**
+- Fix Remove forward slash from desktop filename for Shenzhen I/O (thanks to slowsage)
 - Fix race when preparing download location (thanks to viacheslavka)
 - Fix multithreaded downloads of Windows games (thanks to viacheslavka)
 - Fix DLC installation for Windows games (thanks to viacheslavka)
--- a/minigalaxy/game.py
+++ b/minigalaxy/game.py
@@ -19,11 +19,11 @@ class Game:
         self.category = category
         self.status_file_path = self.get_status_file_path()
 
-    def get_stripped_name(self):
-        return self.__strip_string(self.name)
+    def get_stripped_name(self, to_path=False):
+        return self.__strip_string(self.name, to_path)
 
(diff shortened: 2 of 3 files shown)


sharkwouter (changes requested):
Thanks for this PR, that's a good find!

I did find one issue with this change. Could you add the same change to the `uninstall_game` function? There the desktop file is removed.

slowsage:
Can also add a one line helper function `_sanitize_filename` that is reused if thats cleaner. Not sure what you prefer.

sharkwouter:
If you want to use a sanitize method, I would suggest you do the following:
- Update this function to exclude slashes: https://github.com/sharkwouter/minigalaxy/blob/master/minigalaxy/game.py#L50
- Then use this function in your change: https://github.com/sharkwouter/minigalaxy/blob/master/minigalaxy/game.py#L22
- It probably makes sense then to also make this function use the _strip_string method: https://github.com/sharkwouter/minigalaxy/blob/master/minigalaxy/game.py#L25

slowsage:
`__strip_string` excludes slashes already but unfortunately also spaces which messes up the desktop file names. I have changed the code to reuse `get_stripped_name` in `installer.py` and updated `get_install_directory_name` since its basically the same.

sharkwouter:
Thanks for the fix! It's really appreciated!

PR #610, opened 2024-08-28
Create pt.po

GLSWV:
## Checklist
 

sharkwouter:
Thanks for this PR. I'm surprised we did not have Portuguese Portuguese yet. What I did notice that there are some fields in the file which are marked as fuzzy. Could you address this in your editor? We have instructions on how make translating a bit easier here: https://github.com/sharkwouter/minigalaxy/wiki/Translating-Minigalaxy

PR #611, opened 2024-08-28
Portuguese translation

GLSWV:
Portuguese translation. I hope everything is correct now.

### DIFF
--- a/data/po/pt_PT.po
+++ b/data/po/pt_PT.po
@@ -1,0 +1,752 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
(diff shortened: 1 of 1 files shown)


data/po/pt_PT.po
```diff
@@ -0,0 +1,752 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+#, fuzzy
```
sharkwouter, line 6: You can remove this

sharkwouter (changes requested):
This looks pretty good, thanks for the PR. I had one comment and I would like to ask you to add a line to 1.3.1 in CHANGELOG.md which states something like this:

Added Portuguese translation (thanks to GLSWV)

data/po/pt_PT.po
GLSWV: Done.

PR #620, opened 2024-11-26
Fix Windows games with multiple parts not installing

sharkwouter:
## Description

Right now Windows games of which the installer consisting of multiple files are failing to install with 1.3.0. It seems to be caused by the urls on GOGs end having changed to no longer include the token. This PR should fix this in a more future proof way.

## Checklist
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 **1.3.1**
+- Fix Windows games with multiple parts not installing with wine
 - Minor AppData fixes (thanks to tim77)
 - Added Portuguese translation (thanks to GLSWV)
 
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -2,7 +2,6 @@ import shutil
 import locale
 import os
 import threading
-import re
 import time
 import urllib.parse
 
@@ -285,7 +284,7 @@ class GameTile(Gtk.Box):
                 break
             info = self.api.get_download_file_info(file_info["downlink"])
             total_file_size += info.size
-            # Extract the filename from the download url (filename is between %2F and &token)
+            # Extract the filename from the download url
             filename = urllib.parse.unquote(urllib.parse.urlsplit(download_url).path)
             filename = filename.split("/")[-1]
             download_path = os.path.join(self.download_dir, filename)
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -285,11 +285,9 @@ class GameTile(Gtk.Box):
                 break
             info = self.api.get_download_file_info(file_info["downlink"])
             total_file_size += info.size
-            try:
-                # Extract the filename from the download url (filename is between %2F and &token)
-                filename = urllib.parse.unquote(re.search('%2F(((?!%2F).)*)&t', download_url).group(1))
-            except AttributeError:
-                filename = "{}-{}.bin".format(self.game.get_stripped_name(), key)
+            # Extract the filename from the download url (filename is between %2F and &token)
+            filename = urllib.parse.unquote(urllib.parse.urlsplit(download_url).path)
+            filename = filename.split("/")[-1]
             download_path = os.path.join(self.download_dir, filename)
             if info.md5:
                 self.game.md5sum[os.path.basename(download_path)] = info.md5
--- a/minigalaxy/ui/gametilelist.py
+++ b/minigalaxy/ui/gametilelist.py
@@ -2,7 +2,6 @@ import shutil
 import locale
 import os
 import threading
-import re
 import time
 import urllib.parse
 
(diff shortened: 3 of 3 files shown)


sharkwouter:
I don't think the id always matches the file name. I know the checksum file has file names, but that is not always available. It seems to me the url is still the safest, but I would consider a different solution with some proof that it works.

Yeah, I really don't like the duplicate code between gametile.py and gametilelist.py. I probably should not have accepted that code. We have had some gametilelist specific bugs, because I personally don't use it.

GB609:
I understand your concerns about the id, although i personally think it is there exactly because the devs of GogGalaxy had similar issues and didn't want to continue using url magics. Might be mistaken ofc.

Ultimately it's up to you of course.
I'm at work atm, so i can't test your change. But i'll see tonight if i can install a couple games with your fix and then report back.

GB609:
Alright, my report. I tried to install 'Rise Eterna':
eterna

1. With Minigalaxy version 1.3.0 and wine installation method i get the following files downloaded:
downloads
Installation fails.
error

2. Checkout of branch fix-615
Download file names:
download_fixed
Installation fails again, but **not** because of a missing .BIN.
Instead, I now get the following output:
```
2024-11-26 20:48:37,158 - minigalaxy - ERROR - Error installing game Rise Eterna
  File "/home/karsten/git/minigalaxy/minigalaxy/installer.py", line 66, in install_game
    error_message = extract_installer(game, installer, tmp_dir, language, _use_innoextract)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/karsten/git/minigalaxy/minigalaxy/installer.py", line 131, in extract_installer
    err_msg = extract_windows(game, installer, temp_dir, language, use_innoextract)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/karsten/git/minigalaxy/minigalaxy/installer.py", line 150, in extract_windows
    err_msg = extract_by_wine(game, installer, temp_dir)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/karsten/git/minigalaxy/minigalaxy/installer.py", line 205, in extract_by_wine
    os.unlink(drive)
FileNotFoundError: [Errno 2] Datei oder Verzeichnis nicht gefunden: '/home/karsten/GOG Games/Rise Eterna/prefix/dosdevices/d:'
2024-11-26 20:48:37,541 - minigalaxy - ERROR - Unhandled error.
```

... which happens at the end of the install.
After the install, i can see the following dosdevices in the prefix:
```
[karsten@vortex Rise Eterna]$ ls -la prefix/dosdevices/*\:
lrwxrwxrwx 1 karsten karsten 10 26. Nov 20:46 prefix/dosdevices/c: -> ../drive_c
lrwxrwxrwx 1 karsten karsten  8 26. Nov 20:46 prefix/dosdevices/d:: -> /dev/sr0
lrwxrwxrwx 1 karsten karsten  1 26. Nov 20:46 prefix/dosdevices/z: -> /
```

I think there's some work to be done regarding the setup and configuration of wineprefixes. 
But that might also happen for me because i actually still do have an optical drive which wine  tries to configure as d: automatically. I just tried this:
```
[karsten@vortex ~]$ WINEPREFIX=~/cdtest wine wineboot -u
wine: created the configuration directory '/home/karsten/cdtest'
[karsten@vortex ~]$ ls -la cdtest/dosdevices/*\:
lrwxrwxrwx 1 karsten karsten 10 26. Nov 21:05 cdtest/dosdevices/c: -> ../drive_c
lrwxrwxrwx 1 karsten karsten  8 26. Nov 21:05 cdtest/dosdevices/d:: -> /dev/sr0
lrwxrwxrwx 1 karsten karsten  1 26. Nov 21:05 cdtest/dosdevices/z: -> /
```

So, a fresh wineprefix on my machine always has a symlink d:: to optical.
This is wine's way of saying that it maps the drive itself (device) into the wineprefix, not the contents. This might actually prevent or cause wine to force-delete the manual link to d: created before the setup executable is called as there can't be 2 mappings to the same drive letter.

Besides that, i don't really understand all that drive and temp_dir linking tango. Why not just specify C:\gamename as install directory and be done with it?

GB609:
Update: I got the game install to finish successfully by doing one minor little code change:
```python
<<<    drive = os.path.join(prefix_dir, "dosdevices", "d:")
---------------------------------------------------
>>>    drive = os.path.join(prefix_dir, "dosdevices", "x:")
```

I consider this a bit of a hacky quick-fix to reduce likeliness of name collisions. I could start the game from minigalaxy afterwards although it didn't get past the intro, but that's because of my incomplete wine setup on that machine.

sharkwouter:
Okay, it looks like the error you're running into is a different, unrelated error. I would say we have proven this change works now, though. I'll merge this and keep an eye on other issues.

About the installation code, it's is a bit of a hack and it actually has some problems. What it does is it installs the game to a temporary directory and then moves it. The reason it does this is that that is how it worked with Linux games, which is mostly fine because it is just extracting files from a zip and we don't need all files. For Windows games it causes problems, though, because often it will install registry entries pointing to the temp directory, causing some games to not work anymore after moving.

Honestly, the installation code needs work altogether.

@GB609  Should I change the drive letter to a different name or is this hyper specific to your installation?

GB609:
@sharkwouter How likely is it that i'm the only person using minigalaxy with an optical drive installed? I don't think i'm the only one.

The thing with wineprefixes is: you CAN move them no probs. If you want to do the same as for the linux game, then don't use a drive within the prefix as temp dir, but create the prefix itself at a temporary location and move it after installation.

PR #621, opened 2024-11-27
Fix extract_by_wine when PC has optical drive

GB609:
## Description

Changed the temporary secondary drive used in dosdevices during windows wine installation from `d:` to `t:` which is less likely to create name collisions with real drives created by wine itself. Applies whenever wine creates `d::` during prefix creation.

`d:` and `d::` can't exist at the same time, so wine is removing one of the mappings on its own again. When the removed mapping happens to be `d:` created by the installer, the unlink statement after successful installation will fail because `d:` doesn't exist at that point in time anymore. This error will bubble up as unexpected and abort the entire game installation process.

## Checklist
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Fixed windows install issue for PCs with optical drives (thanks to GB609)
+
 **1.3.1**
 - Fix Windows games with multiple parts not installing with wine
 - Minor AppData fixes (thanks to tim77)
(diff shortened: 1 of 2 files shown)


GB609:
The trailing whitespace check is pretty strict.
Don't merge yet, i'm going to do a squash first to get my repeated trials cleaned up.

sharkwouter:
Thanks! I appreciate the contribution :smile:

Issue #618, opened 2024-11-08
Homeworld Emergence - Wine extraction failed


NemoBarbarossa:
I run Fedora 41 and tried to install Homeworld Emergence.

First I had minigalaxy installed as flatpak. When the error happened, I tried if I had to manually install wine as well so I added that via flatpak as well. When the error kept on coming up I removed the flatpaks and installed minigalaxy through dnf which brought wine as a prerequisite already.

I still get the extraction error though. When looking for it I came across issue #357 and was wondering how I could run the mentioned debug mode.
I also saw the discussions about innoextract and went looking for it. Seems that was also installed as a prerequisite and I'm not sure if I can uninstall it.

Is there a way to force one installation method or the other for individual titles?

dnf package information:
```
Name            : innoextract
Epoch           : 0
Version         : 1.9
Release         : 15.fc41
Architecture    : x86_64
Installed size  : 582.5 KiB
Source          : innoextract-1.9-15.fc41.src.rpm
From repository : fedora
Summary         : Tool to extract installers created by Inno Setup
URL             : https://constexpr.org/innoextract/
License         : zlib
Description     : Inno Setup is a tool to create installers for Microsoft Windows
                : applications. innoextract allows to extract such installers under
                : non-windows systems without running the actual installer using wine.
Vendor          : Fedora Project

Name            : minigalaxy
Epoch           : 0
Version         : 1.3.0
Release         : 3.fc41
Architecture    : noarch
Installed size  : 768.9 KiB
Source          : minigalaxy-1.3.0-3.fc41.src.rpm
From repository : fedora
Summary         : GOG client for Linux that lets you download and play your GOG Linux games
URL             : https://sharkwouter.github.io/minigalaxy
License         : GPL-3.0-or-later and CC-BY-3.0
Description     : A simple GOG client for Linux that lets you download and play your GOG Linux
                : games.
Vendor          : Fedora Project

Name            : wine
Epoch           : 0
Version         : 9.15
Release         : 1.fc41
Architecture    : x86_64
Installed size  : 0.0   B
Source          : wine-9.15-1.fc41.src.rpm
From repository : fedora
Summary         : A compatibility layer for windows applications
URL             : https://www.winehq.org/
License         : LGPL-2.1-or-later
Description     : Wine as a compatibility layer for UNIX to run Windows applications. This
                : package includes a program loader, which allows unmodified Windows
[...]

Issue #117, opened 2020-04-06
Uninstalling a wine game doesn't remove menu items


sharkwouter:
Currently when you uninstall a wine game, it just deletes the directory in which the games is located. This means items in the user's operating system menu which were added by the GOG installer will remain.

sharkwouter:
This one is a bit harder to fix than I initially thought. The shortcuts are installed in ``~/.local/share/applications/wine/Programs/GOG.com/``. The issue is that the name of the directory is not decided by Minigalaxy, but by the installer.

makson96:
Can you post a content of such a file? My idea is that we can read all files content in this directory (~/.local/share/applications/wine/Programs/GOG.com/) and if there is reference in such a file to game install directory. Remove that file to.

sharkwouter:
I think the shortcuts shouldn't be created at all to be honest. I'm sure we can make sure of that using some command like what #289 tries to do.

makson96:
This one should be fixed together with #313

sharkwouter:
This will be considered fixed, which Innoextract is a mandory dependency.

sharkwouter:
This is actually fixed in the master branch now. The next release will have a fix for this.

sharkwouter:
What I mean to say is, that we won't care about this anymore, because we no longer create these shortcut files.

GB609:
@sharkwouter 
It is mostly fixed for regular installations.
There is one situation where shortcuts from wine will still appear:
Right now, only the code that runs the installation temporarily disables the winemenubuilder. And prevents shortcuts while the installation runs.

But i've also seen unity games that start some other dependency installation when the game is started the first time. However, the winemenebuilder is not disabled then, because the code does not set WINEDLLOVERRIDE currently when starting a game (users might have configured other overrides to get the game to run correctly and we'd overwrite those when setting the env).
I did experiment with setting a registry key to disable the menubuilder, but wasnt successful up to now.

sharkwouter:
Interesting, lets consider that edge cases for now and wait for specific bug reports.

Issue #605, opened 2024-07-05
Banished: Wine extraction failed


tastytea:
i'm using the minigalaxy flatpak, version 1.3.0.

```shellsession
% MG_DEBUG=1 flatpak run io.github.sharkwouter.Minigalaxy
[…]
2024-07-05 12:58:52,443 - minigalaxy.download_manager.DownloadManager - DEBUG - Returning result from _download_operation: True
2024-07-05 12:58:52,446 - minigalaxy.download_manager.DownloadManager - DEBUG - Download finished, thread 140233517553344
2024-07-05 12:58:52,448 - minigalaxy - INFO - Installing Banished
2024-07-05 12:58:52,450 - minigalaxy.download_manager.DownloadManager - DEBUG - Removing download from active downloads list
2024-07-05 12:58:54,208 - minigalaxy - INFO - setup_banished_64_1.0.7_(14938).exe integrity is preserved. MD5 is: 463b2720c5c88c28f24de9176b8b1ec4
2024-07-05 13:00:13,283 - minigalaxy - ERROR - Wine extraction failed.
```

is there a way to get more debug info, try out different wine versions, …?

sharkwouter:
This is probably an issue with the flatpak package. Can you reproduce it with a natively installed verison?

tastytea:
i tried to install minigalaxy_1.3.0_all.deb from the github release page in a debian 12 VM and minigalaxy won't start:

<details> <summary>traceback</summary>

```
Traceback (most recent call last):
  File "/usr/games/minigalaxy", line 66, in <module>
    main()
  File "/usr/games/minigalaxy", line 46, in main
    from minigalaxy.ui.gtk import Gtk
  File "/usr/lib/python3/dist-packages/minigalaxy/ui/__init__.py", line 3, in <module>
    from minigalaxy.ui.window import Window             # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/minigalaxy/ui/window.py", line 6, in <module>
    from minigalaxy.ui.categoryfilters import CategoryFilters
  File "/usr/lib/python3/dist-packages/minigalaxy/ui/categoryfilters.py", line 6, in <module>
    from minigalaxy.ui.filterswitch import FilterSwitch
  File "/usr/lib/python3/dist-packages/minigalaxy/ui/filterswitch.py", line 4, in <module>
[...]

Issue #602, opened 2024-06-13
Current minigalaxy 1.3.0 seems to somehow corrupt the wine prefixes


ell1e:
Just some months ago, I managed to install and run most games, [other than DOS due to that other unresolved issue](https://github.com/sharkwouter/minigalaxy/issues/592) using minigalaxy just fine. But with the current version 1.3.0, somehow it seems to break the wine prefixes after it managed to run the GOG.com installer just fine, leading to me being unable to launch the game with either no reaction to the play button or strange errors popping up:

![Screenshot_20240613_213310](https://github.com/sharkwouter/minigalaxy/assets/64124388/e8f29349-1c1b-4dba-bf50-6b81793b29b9)

Now if it was just one game affected I would just reinstall it, assuming a weird lightning in a bottle rare glitch, but *all* regular Windows games seem to be affected. Winecfg won't launch either for any of them. On the host system outside of flatpak, wine is installed both as a 64bit and 32bit version. I couldn't think of any obvious reason why this would be happening, so is this a minigalaxy regression of some sorts?

**Affected minigalaxy version:** 1.3.0

**Affected on linux distro:** openSUSE Slowroll

**Install method for minigalaxy:** flathub/flatpak

sharkwouter:
Odd, seems like a flatpak specific issue. Could you report this to the flatpak repo?

ell1e:
Sadly, combined with #592 it seems like in practice, it's not possible to launch any games anymore no matter which type. I reported that other one to the flatpak but there wasn't much of a reaction.

Is there a different distribution independent install method for minigalaxy that works?

ell1e:
This seems to have been fixed somehow, perhaps #615 or something similar was related to this.

PR #622, opened 2024-11-28
Support UMU-Launcher

GB609:
## Description
This code change introduces a dropdown in preference giving a choice between 'Wine' and 'UMU-Launcher' as windows game runner.
I've introduced a helper, wine_utils, to handle everything about wine: environment variables, locating the executable etc. There were quite a few other code changes/refactorings as well, partly for maintenance, partly to actually use the new utils.

I also started revising the entire install strategy for windows games as a lot of issues i can see here stem from the different ways to install and that they both seem to have issues.

I'm in favor of using the conventional install within wine over innoextract. 
The main reason is that innoextract is basically just like unzip, it doesn't perform any post-file-copy steps, nor does it execute the installer scripts. This leaves the 'installations' in a potentially undefined intermediate state where a lot of additional code is needed to 'make it work'. That's why i temporarily disabled innoextract hard-coded in my branch. Therefore the system installation won't matter. The revised installer does not utilize a temporary directory (with file moving around afterwards) anymore. instead it configures the installer to use a well-known pre-defined directory within the wineprefix.

If there is a desire to still keep innoextract even when the changed installation proves to be stable, i also have a few other changes in stock to make this an option, allowing to toggle between innoextract and regular install without having to remove inno from the system.

I'm putting this up as a draft PR here already (although it is in a bit of an early stage with lots of broken tests) because i need some help. I need information about games and 'constellations':

**1. Test of game installations, both with regular wine and umu**
I don't own too many windows games on GOG yet and my wallet is not infinite.
I'm mostly interested in the following things:
1. Does the installation itself work?
2. A file listing of the install directory in case of success, as well as the content of goggame-*.info file (if any). However, please try to only report if it is not the most trivial case imaginable (later more on that)
3. In case of failure, any console output, message dialog error etc is helpful.
4. In all cases: I configured the installer to write its setup configuration to `gamedirectory/prefix/drive_c/setup.inf`. I'm currently considering whether this file contains any reliable input usable to make workingdir and executable detection easier.

**2. Running games**
Installation should basically work for all games, since the working directories and installer executables are well known.
This does not apply when running a game, however. Even before my code changes, minigalaxy had quite some logic to try and detect the game's exe file.
There's some try to parse the goggame-*.info file (these appear to be optional), then a direct search for the first exe to found in the game's install dir (some games place their exe in a subdir like bin).
But all installers generate desktop shortcuts. It might be possible to leverage them instead.

That's why i need input on games that install, but don't launch. These are most likely not simple cases.

**3. Protonfixes**
Some games require winetricks etc.
The folks from UMU started to track this in a database [UMU-DB](https://github.com/Open-Wine-Components/umu-database/blob/main/umu-database.csv)
UMU-Launcher requires a variable named 'GAMEID', which can be arbitrary, but should correspond to value of UMU_ID for a game in that database (if it is there). I've implemented a lookup function and what i need to know is if it actually retrieves the right values for games with protonfixes. This requires a working internet connection. Without internet, the lookup will generate a reasonable default and try again the next time anything within that game's prefix is launched with umu.
I added a tempory dirty log statement to see the full wine launch commands on installation, including GAMEID= when umu is chosen as runner.
So, if anyone owns a game that appears in the database, under store gog, steam or none, please let me know if the value of GAMEID matches what is in that table for UMU_ID.

**4. Simple case**
A 'simple' case either:
* has only one executable (other than the unins000.exe) in it's root directory
* or several, but a parsable goggame-*.info file giving info about which one to use
* and does not require to step into any other working directory. Determining this is a bit harder. When there is a goggame-*.info file, but it does not contain any 'workingDir' property, it should be ok. Without this file, all we can do is try to start and check the output for errors.

**5. Savegames / home pollution**
This is something that i have not touched, but it is related to wine configuration.
Regular Wine normally tries to integrate windows applications as seamlessly as possible into the desktop environment. As such, it will link windows user folders like 'Documents' or 'Downloads' to those in /home/user. Depending on the game, it will start putting stuff there. I think it's arguable whether this is desired and this behavior can be changed. When this integration is disabled, the 'Documents' etc will just be regular folders nested somewhere deep within the game's wineprefix directories. The question here is, shall Minigalaxy provide a default configuration or leave it to the users to configure each and every game installation by themselves regarding that?
This is also roughly related to cloud saves or similar. But to achieve this, a uniform and well known save location is necessary which is very hard to do when running games from stone age vs brand new ones for more recent windows versions. But i'm not going on elaborating this here, since this is going off track.

The sandboxing/pollution issue might not apply the same way in UMU since it uses the isolated proton stuff. 

The changes to the installer potentially impact or fix:
#618, #605, #602, #512, #506
#289 - i did an adjustment to the /VERYSILENT argument in the wine installer. I found not having any visual feedback at all during a long-running installation makes me feel impatient or that there's something wrong. So i changed the flag to /SILENT which only shows the installation progress, but no configuration options dialog. Potential candidate to go into preferences.

The wine_env configuration might fix:
#117, #313 - i have disabled the winemenubuilder in all prefixes managed by minigalaxy, so shortcuts from within wine are not automatically added to the linux desktop. Only applies on new installations. If there already are shortcuts, they need to be removed manually.

The enhancement of preferences to pick a wine variant opens up the possibility to work on:
#288

## Checklist
 
### DIFF
--- a/README.md
+++ b/README.md
@@ -82,14 +82,15 @@ Other Linux distributions may work as well. Minigalaxy requires the following de
 To run windows games, the following optional dependencies are required:
 
 - wine or umu-launcher (as an alternative to standard wine)
-- innoextract: for installations with native linux tools
+- innoextract: For installations with native linux tools
+- zenity: Used by UMU-Launcher to show progress dialogs when updating the proton framework. Not required to use UMU.
 
 **Note:**
 
 Some 'native' Linux games available on GOG are Windows applications prepackaged with Wine in reality.
-Running them with UMU-Launcher can not be guaranteed to work in all situations. 
+Using UMU-Launcher as runner can not be guaranteed to work in all situations.
 Depending on the game directory structure, Minigalaxy might not be able to detect that it is a windows game
-and thus won't be able to configure or change the actual wine invocation.
+and thus won't be able to change the actual wine invocation of the game itself.
 
 Also, Minigalaxy expects all dependencies of Wine to be setup correctly, it can't install missing drivers, plugins or packages.
 
--- a/README.md
+++ b/README.md
@@ -79,6 +79,20 @@ Other Linux distributions may work as well. Minigalaxy requires the following de
 - Python Requests
 - gettext
 
+To run windows games, the following optional dependencies are required:
+
+- wine or umu-launcher (as an alternative to standard wine)
+- innoextract: for installations with native linux tools
+
+**Note:**
+
+Some 'native' Linux games available on GOG are Windows applications prepackaged with Wine in reality.
+Running them with UMU-Launcher can not be guaranteed to work in all situations. 
+Depending on the game directory structure, Minigalaxy might not be able to detect that it is a windows game
+and thus won't be able to configure or change the actual wine invocation.
+
+Also, Minigalaxy expects all dependencies of Wine to be setup correctly, it can't install missing drivers, plugins or packages.
+
 ## Installation
 
 <a href="https://repology.org/project/minigalaxy/versions">
--- a/data/po/de.po
+++ b/data/po/de.po
@@ -446,9 +446,17 @@ msgstr "Ob versteckte Spiele in der Bibliothek angezeigt werden oder nicht"
 #: data/ui/preferences.ui:285
 msgctxt "create_menu_shortcuts_tooltip"
 msgid "Whether shortcuts are created for newly installed games or not"
-msgstr ""
-"Ob Verknüpfungen für neu installierte Spiele erstellt werden oder nicht"
+msgstr "Ob Verknüpfungen für neu installierte Spiele erstellt werden oder nicht"
 
+#: data/ui/preferences.ui:298
+msgctxt "windows_wine_variant_tooltip"
+msgid "Define the wine variant to use as default runner for new windows game installations."
+msgstr "Variante von wine, die standardmäßig für neu installierte Windows-Spiele verwendet werden soll."
+
+#: data/ui/preferences.ui:301
+msgctxt "windows_wine_variant"
+msgid "Default wine variant: "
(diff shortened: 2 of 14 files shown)


GB609:
Ok, i'll start.
I've got the following Windows games which i'll try to install and run:
* Rise Eterna
* Outcast - Second Contact
* Severed Steel
* Star Trek(TM): Voyager - Elite Force
* Trepang² Demo
* Genesis: Alpha One Deluxe Edition
* Lethis - Path of Progress

I will update the following table as i go one (cant test all at once, takes too much time).
All tests are done with regular wine first.
| Game | Install Works? | starts | Simple case? | has parseable gog.info | requires different workdir? |  Notes | diag files |
| - | - | - | - | - | - | - | - |
| Rise Eterna | yes | yes | yes | yes | | has 2 exes, but a simple gog.info |
| Severed Steel | yes | starts, but freezes on intro. havent debugged for the reason yet | yes* | gog.info is there, but primary is of type "launcher" tasks | | Only one *.exe in root which is a launcher, BUT primary info in gog.info defines additional args for this | [ls_severed_steel.txt](https://github.com/user-attachments/files/17965064/ls_severed_steel.txt) [info_severed_steel.txt](https://github.com/user-attachments/files/17965066/info_severed_steel.txt) |
| Star Trek | yes | yes | yes | yes | | multiple exe, but has primary gametask | - |
| Genesis | currently freezes on dxsetup with wine (triggered at end of game install) | starts when setup is force-killed (although my pc is currently missing some dependency for wine | yes | yes | not for primary task | download progress bar toggles back and forth, probably because several larger files are downloaded in parallel. | - |
| Outcast | TBD | - | - | - | - | - |
| Trepang | TBD | - | - | - | - | - |
| Lethis | TBD | - | - | - | - | - | - |

PR #624, opened 2024-12-02
Rework installation of windows games

GB609:
## Description

This PR offers a complete rework of the installer.py:extract_windows installation method for windows games.
1. There is no temporary directory involved anymore.
2. Thus, no linking of an additional dosdevice to temp, paths remain constant throughout the entire lifetime of the game install
3. The game installation path within the wine prefix will be set (and assumed to be) c:\game. Removing the temporary directory and using a fixed path will avoid any issues with registry keys created during installation.
4. The installer tries the windows installation with 2 sets of arguments. First fully unattended. When this fails, it will try to open the manual installation dialog, but preconfigured with c:\game as target dir. Potential TODO: Show an additional dialog in the fallback situation to ask users not to change the directory name. The fallback is necessary as some installer seem to reject the /SILENT argument. I've had this with 'Lethis - Path of progress' and 'Record of Agarest War'. Maybe there is a GOG Galaxy api we could use to get the required args and fix this for good without fallback.
6. During game installation, winemenubuilder is disabled to prevent wine from creating desktop shortcuts outside of Minigalaxy's control.
7. Shortcut creation done by Minigalaxy was changed to correctly include the command prefix 'env ...'. Commands placed in the Exec property are now correctly and fully shell-style quoted by shlex, instead of manual escaping.

The changes to the installer and how wine environment variables are handled made it also necessary to touch the launcher to correctly configure wine prefixes on launch.
Wherever necessary and possible, wine commands now use 'env' as prefix for configuration, Minigalaxy will not modify its own environment any longer.

I have also used shlex for parsing and command generation to improve handling of quotes and spaces. This makes it possible to use env variables or arguments containing spaces in the game properties (if quoted correctly).

installer.py:_exe_cmd was also reworked to use poll() and readline instead of communicate(). The reason for that is that wine can be very verbose with its stdout and i'm trying to prevent process buffers from filling up and thus effectively halting an installation. I think i've had this with a game that also installed directx afterwards  where using communicate would never return because the directx installation was apparently frozen.

The downside of these changes is that they are not entirely free of potentially breaking backwards compatibility. Games that save directory paths to registry and were installed through wine before this PR might break (although it is more likely they didn't work before).
I've tried to keep the effects as small as possible, by adding a few lines of code in launcher.py which try to restore the c:\game link on game launching. Desktop shortcuts are not regenerated and launching over shortcuts won't apply the fix.

But i've also added a note in the readme about this. If that is not the right place, just tell me where and what i shall add instead.

The initial change to the installation is also included in #622, but with some enhancements to also handle umu. The fallback solution is new to this branch.

## Checklist
 
Edit: Adding TODOs from discussion here for easier tracking:

[2516779478](https://github.com/sharkwouter/minigalaxy/pull/624#issuecomment-2516779478):
- [ ] install game with inno and test for working c:\game link 

[2516801895](https://github.com/sharkwouter/minigalaxy/pull/624#issuecomment-2516801895):
- [ ] manual test installation with inno success for temp_dir cleanup and NOT wine fallback
- [ ] manual test installation with inno extract ok but total failure, working wine fallback: temp_dir cleanup, usage of files from temp?

[2520299969](https://github.com/sharkwouter/minigalaxy/pull/624#issuecomment-2520299969)
- [ ] double-check wine installer language
- [ ] configure language in game launched in wine

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,15 +26,17 @@ In addition to that, Minigalaxy also allows you to:
 - Install Windows games using Wine
 
 ### Backwards compatibility
-Minigalaxy version 1.3.2 or higher changes some aspects of windows game installations through wine.
+Minigalaxy version 1.3.2 and higher change some aspects of windows game installations through wine.
 It will try to adapt already installed games to the new concept when launched through Minigalaxy.
+
 However, this will *likely break games* that save some directory paths in the *windows registry*.
 In that case, only a reinstallation will repair the game. 
+
 **Please make sure to backup any save games you might have within the game folder**
 
 The windows installer in wine uses a 2-step attempt to install games. 
 1. An unattended installer.
-2. In case this fails, the regular installation wizard will open. It is mandatory not to change the 
+2. In case this fails, the regular installation wizard will open. **Please do not change** the  
 install directory 'c:\game' given in the wizard as this an elementary part of the wine fix!
 
 ## Supported languages
--- a/README.md
+++ b/README.md
@@ -25,6 +25,18 @@ In addition to that, Minigalaxy also allows you to:
 - Use the system's ScummVM or DOSBox installation
 - Install Windows games using Wine
 
+### Backwards compatibility
+Minigalaxy version 1.3.2 or higher changes some aspects of windows game installations through wine.
+It will try to adapt already installed games to the new concept when launched through Minigalaxy.
+However, this will *likely break games* that save some directory paths in the *windows registry*.
+In that case, only a reinstallation will repair the game. 
+**Please make sure to backup any save games you might have within the game folder**
+
+The windows installer in wine uses a 2-step attempt to install games. 
+1. An unattended installer.
+2. In case this fails, the regular installation wizard will open. It is mandatory not to change the 
+install directory 'c:\game' given in the wizard as this an elementary part of the wine fix!
+
 ## Supported languages
 
 Currently, Minigalaxy can be displayed in the following languages:
--- a/README.md
+++ b/README.md
@@ -29,15 +29,10 @@ In addition to that, Minigalaxy also allows you to:
 Minigalaxy version 1.3.2 and higher change some aspects of windows game installations through wine.
 It will try to adapt already installed games to the new concept when launched through Minigalaxy.
 
-However, this will *likely break games* that save some directory paths in the *windows registry*.
-In that case, only a reinstallation will repair the game. 
-
-**Please make sure to backup any save games you might have within the game folder**
-
-The windows installer in wine uses a 2-step attempt to install games. 
+The windows installer in wine now uses a 2-step attempt to install games. 
 1. An unattended installer.
 2. In case this fails, the regular installation wizard will open. **Please do not change** the  
-install directory 'c:\game' given in the wizard as this an elementary part of the wine fix!
+install directory 'c:\game' given in the wizard as this an elementary part of the wine fix.
 
 ## Supported languages
 
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -221,10 +221,11 @@ def extract_by_wine(game, installer, temp_dir, config=Config()):
         '/SILENT'
     ]
 
-    #first try full, unattended install
+    # first, try full unattended install
     success = try_wine_command(installer_cmd_basic + installer_args_full)
     if not success:
-        #some games will reject the /SILENT flag. Open normal installer as fallback and hope for the best
+        # some games will reject the /SILENT flag.
+        # Open normal installer as fallback and hope for the best
         print('Unattended install failed. Try install with wizard dialog.', file=sys.stderr)
         success = try_wine_command(installer_cmd_basic)
 
@@ -422,11 +423,9 @@ def _mv(source_dir, target_dir):
 def lang_install(installer: str, language: str):
     languages = []
     arg = ""
-    process = subprocess.Popen(["innoextract", installer, "--list-languages"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    stdout, stderr = process.communicate()
-    output = stdout.decode("utf-8")
+    stdout, stderr, ret_code = _exe_cmd(["innoextract", installer, "--list-languages"])
 
-    for line in output.split('\n'):
+    for line in stdout.split('\n'):
         if not line.startswith(' -'):
             continue
         languages.append(line[3:])
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -221,8 +221,10 @@ def extract_by_wine(game, installer, temp_dir, config=Config()):
         '/SILENT'
     ]
 
+    #first try full, unattended install
     success = try_wine_command(installer_cmd_basic + installer_args_full)
     if not success:
+        #some games will reject the /SILENT flag. Open normal installer as fallback and hope for the best
         print('Unattended install failed. Try install with wizard dialog.', file=sys.stderr)
         success = try_wine_command(installer_cmd_basic)
 
@@ -272,20 +274,13 @@ def copy_thumbnail(game):
     return error_message
 
 
-def get_exec_line(game):
-    exe_cmd_list = get_execute_command(game)
-    for i in range(len(exe_cmd_list)):
-        exe_cmd_list[i] = exe_cmd_list[i].replace(" ", "\\ ")
-    return " ".join(exe_cmd_list)
-
-
 def create_applications_file(game):
     error_message = ""
     path_to_shortcut = os.path.join(APPLICATIONS_DIR, "{}.desktop".format(game.get_stripped_name(to_path=True)))
-    exe_cmd = get_exec_line(game)
+    exe_cmd = shlex.join(get_execute_command(game))
     # Create desktop file definition
     desktop_context = {
-        "game_bin_path": os.path.join('"{}"'.format(game.install_dir.replace('"', '\\"')), exe_cmd),
+        "game_bin_path": exe_cmd,
         "game_name": game.name,
         "game_install_dir": game.install_dir,
         "game_icon_path": os.path.join(game.install_dir, 'support/icon.png')
@@ -298,7 +293,8 @@ def create_applications_file(game):
         Exec={game_bin_path}
         Path={game_install_dir}
         Name={game_name}
-        Icon={game_icon_path}""".format(**desktop_context)
+        Icon={game_icon_path}
+        Category=Game""".format(**desktop_context)
     if not os.path.isfile(path_to_shortcut):
         try:
             with open(path_to_shortcut, 'w+') as desktop_file:
@@ -378,7 +374,9 @@ def uninstall_game(game):
 
 
 def _exe_cmd(cmd, capture_output=True, print_output=False):
-    print(f'executing command: {shlex.join(cmd)}')
+    """Wine commands are very verbose.
+    We should consume the output in regularly to prevent buffers (and minigalaxy memory) from filling up"""
+    print(f'executing wine command: {shlex.join(cmd)}')
     std_out = []
     process = subprocess.Popen(cmd,
                                stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -186,7 +186,7 @@ def extract_by_wine(game, installer, temp_dir, config=Config()):
     prefix_dir = os.path.join(game.install_dir, "prefix")
     game_dir = os.path.join(prefix_dir, "dosdevices", 'c:', 'game')
     wine_env = [
-        "WINEPREFIX={}".format(prefix_dir),
+        f"WINEPREFIX={prefix_dir}",
         "WINEDLLOVERRIDES=winemenubuilder.exe=d"
     ]
     wine_bin = shutil.which('wine')
@@ -224,7 +224,8 @@ def extract_by_wine(game, installer, temp_dir, config=Config()):
     success = try_wine_command(installer_cmd_basic + installer_args_full)
     if not success:
         print('Unattended install failed. Try install with wizard dialog.', file=sys.stderr)
-        try_wine_command(installer_cmd_basic)
+        success = try_wine_command(installer_cmd_basic)
+
     if not success:
         return _("Wine extraction failed.")
 
@@ -233,7 +234,7 @@ def extract_by_wine(game, installer, temp_dir, config=Config()):
 
 def try_wine_command(command_arr):
     print('trying to run wine command:', shlex.join(command_arr))
-    stdout, stderr, exitcode = _exe_cmd(command_arr)
+    stdout, stderr, exitcode = _exe_cmd(command_arr, False, True)
     print(stdout)
     if exitcode not in [0]:
         print(stderr, file=sys.stderr)
@@ -376,12 +377,32 @@ def uninstall_game(game):
         os.remove(path_to_shortcut)
 
 
-def _exe_cmd(cmd):
-    process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    stdout, stderr = process.communicate()
-    stdout = stdout.decode("utf-8")
-    stderr = stderr.decode("utf-8")
-    return stdout, stderr, process.returncode
+def _exe_cmd(cmd, capture_output=True, print_output=False):
+    print(f'executing command: {shlex.join(cmd)}')
+    std_out = []
+    process = subprocess.Popen(cmd,
+                               stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
+                               bufsize=1, universal_newlines=True, encoding="utf-8")
+    rc = process.poll()
+    while rc is None:
+        out_line = process.stdout.readline()
+        if capture_output and out_line != '':
+            std_out.append(out_line)
+
+        if print_output:
+            print(out_line, end='')
+
+        rc = process.poll()
+
+    print('command finished, read remaining output (if any)')
+    for line in process.stdout.readlines():
+        std_out.append(line)
+    print('done')
+
+    process.stdout.close()
+
+    output = ''.join(std_out)
+    return output, output, rc
 
 
 def _mv(source_dir, target_dir):
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -1,10 +1,12 @@
 import sys
 import os
 import shutil
+import shlex
 import subprocess
 import hashlib
 import textwrap
 
+from minigalaxy.config import Config
 from minigalaxy.game import Game
 from minigalaxy.logger import logger
 from minigalaxy.translation import _
@@ -179,40 +181,65 @@ def extract_by_innoextract(installer: str, temp_dir: str, language: str, use_inn
     return err_msg
 
 
-def extract_by_wine(game, installer, temp_dir):
-    err_msg = ""
+def extract_by_wine(game, installer, temp_dir, config=Config()):
     # Set the prefix for Windows games
     prefix_dir = os.path.join(game.install_dir, "prefix")
-    """pick a letter that is unlikely to create collisions with the actual mount/hw setup:
-    wine creates links for mounted media and optical drives
-    this might lead to errors because wine knows 2 names for these - d: and d::
-    (difference: : exposes directory, :: exposes the block device itself)
-    But they can't exist at the same time within a prefix.
-    Changing this letter is a temporary fix, the entire install method requires an overhaul in the long run"""
-    drive = os.path.join(prefix_dir, "dosdevices", "t:")
+    game_dir = os.path.join(prefix_dir, "dosdevices", 'c:', 'game')
+    wine_env = [
+        "WINEPREFIX={}".format(prefix_dir),
+        "WINEDLLOVERRIDES=winemenubuilder.exe=d"
+    ]
+    wine_bin = shutil.which('wine')
+
     if not os.path.exists(prefix_dir):
         os.makedirs(prefix_dir, mode=0o755)
         # Creating the prefix before modifying dosdevices
-        command = ["env", "WINEPREFIX={}".format(prefix_dir), "wine", "start", "/B", "cmd", "/C", "exit"]
-        stdout, stderr, exitcode = _exe_cmd(command)
-        if exitcode not in [0]:
-            print(stderr, file=sys.stderr)
+        command = ["env", *wine_env, wine_bin, "wineboot", "-u"]
+        if not try_wine_command(command):
             return _("Wineprefix creation failed.")
-    if os.path.exists(drive):
-        os.unlink(drive)
-    os.symlink(temp_dir, drive)
-    _dir = os.path.join(temp_dir, os.path.basename(game.install_dir))  # can't install to drive root
+
+    # calculate relative link from prefix-internal folder to game.install_dir
+    # keeping it relative makes sure that the game can be moved around without stuff breaking
+    if not os.path.exists(game_dir):
+        # 'game' directory itself does not count
+        canonical_prefix = os.path.realpath(os.path.join(game_dir, '..'))
+        relative = os.path.relpath(game.install_dir, canonical_prefix)
+        os.symlink(relative, game_dir)
     # It's possible to set install dir as argument before installation
-    command = ["env", "WINEPREFIX={}".format(prefix_dir), "wine", installer, "/dir={}".format(_dir), "/VERYSILENT"]
-    stdout, stderr, exitcode = _exe_cmd(command)
+    installer_cmd_basic = [
+        'env', *wine_env, wine_bin, installer,
+        # use hard-coded directory name within wine, its just a backlink to game.install_dir
+        # this avoids issues with varying path and spaces
+        "/DIR=c:\\game",
+        # capture information for debugging during install
+        "/LOG=c:\\install.log",
+    ]
+    installer_args_full = [
+        f"/LANG={config.lang}",
+        "/SAVEINF=c:\\setup.inf",
+        # installers can run very long, give at least a bit of visual feedback
+        '/SILENT'
+    ]
+
+    success = try_wine_command(installer_cmd_basic + installer_args_full)
+    if not success:
+        print('Unattended install failed. Try install with wizard dialog.', file=sys.stderr)
+        try_wine_command(installer_cmd_basic)
+    if not success:
+        return _("Wine extraction failed.")
+
+    return ""
+
+
+def try_wine_command(command_arr):
+    print('trying to run wine command:', shlex.join(command_arr))
+    stdout, stderr, exitcode = _exe_cmd(command_arr)
+    print(stdout)
     if exitcode not in [0]:
-        err_msg = _("Wine extraction failed.")
-    elif os.path.exists(drive):
-        """check for existence as a pure safety-measure in case
-        some power-user has pre-configured the letter we picked with double colon"""
-        os.unlink(drive)
-        os.symlink("../../..", drive)
-    return err_msg
+        print(stderr, file=sys.stderr)
+        return False
+
+    return True
 
 
 def move_and_overwrite(game, temp_dir, use_innoextract):
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -5,6 +5,7 @@ import shlex
 import subprocess
 import hashlib
 import textwrap
+import time
 
 from minigalaxy.config import Config
 from minigalaxy.game import Game
@@ -237,7 +238,7 @@ def extract_by_wine(game, installer, temp_dir, config=Config()):
 
 def try_wine_command(command_arr):
     print('trying to run wine command:', shlex.join(command_arr))
-    stdout, stderr, exitcode = _exe_cmd(command_arr, False, True)
+    stdout, stderr, exitcode = _exe_cmd(command_arr, True)
     print(stdout)
     if exitcode not in [0]:
         print(stderr, file=sys.stderr)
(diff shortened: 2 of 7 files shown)


sharkwouter:
Thanks! I really appreciate this work. I'll review it this week, hopefully tonight.

GB609:
@sharkwouter There's more to come.
Just didn't want to put it all into one PR because that would become pretty large.
When this installer proofs to be sufficiently stable, i wanted to make the choice inno vs. wine configurable so that minigalaxy won't automatically prefer inno. This makes testing and using both much more convenient because the system packages won't need to be removed.

minigalaxy/installer.py
```diff
@@ -179,40 +181,69 @@ def extract_by_innoextract(installer: str, temp_dir: str, language: str, use_inn
...
+    if not success:
+        return _("Wine extraction failed.")
+
+    return ""
+
+
+def try_wine_command(command_arr):
+    print('trying to run wine command:', shlex.join(command_arr))
+    stdout, stderr, exitcode = _exe_cmd(command_arr, False, True)
+    print(stdout)
```
GB609, line 241: I think i can remove this print command again, as _exe_cmd(..., False, True) already prints by itself and would return an empty string anyway.

minigalaxy/installer.py
```diff
@@ -349,12 +374,34 @@ def uninstall_game(game):
...
+            std_out.append(out_line)
+
+        if print_output:
+            print(out_line, end='')
+
+        rc = process.poll()
+
+    print('command finished, read remaining output (if any)')
+    for line in process.stdout.readlines():
+        std_out.append(line)
```
sharkwouter, line 398: You should convert to string here using `line.decode("utf-8")`.

minigalaxy/installer.py
```diff
@@ -376,11 +423,9 @@ def _mv(source_dir, target_dir):
 def lang_install(installer: str, language: str):
     languages = []
     arg = ""
-    process = subprocess.Popen(["innoextract", installer, "--list-languages"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    stdout, stderr = process.communicate()
-    output = stdout.decode("utf-8")
+    stdout, stderr, ret_code = _exe_cmd(["innoextract", installer, "--list-languages"])
```
sharkwouter, line 426: This is not a comment about your code, but it is bad that this code will fail silently if there is an error. We should address this at some point, but not in this PR.

README.md
```diff
@@ -25,6 +25,20 @@ In addition to that, Minigalaxy also allows you to:
 - Use the system's ScummVM or DOSBox installation
 - Install Windows games using Wine
 
+### Backwards compatibility
+Minigalaxy version 1.3.2 and higher change some aspects of windows game installations through wine.
+It will try to adapt already installed games to the new concept when launched through Minigalaxy.
+
+However, this will *likely break games* that save some directory paths in the *windows registry*.
+In that case, only a reinstallation will repair the game.
```
sharkwouter, line 33: I don't think want this. We should just use the old behavior for old game installations.

minigalaxy/installer.py
```diff
@@ -179,40 +181,69 @@ def extract_by_innoextract(installer: str, temp_dir: str, language: str, use_inn
...
-    os.symlink(temp_dir, drive)
-    _dir = os.path.join(temp_dir, os.path.basename(game.install_dir))  # can't install to drive root
+
+    # calculate relative link from prefix-internal folder to game.install_dir
+    # keeping it relative makes sure that the game can be moved around without stuff breaking
+    if not os.path.exists(game_dir):
+        # 'game' directory itself does not count
+        canonical_prefix = os.path.realpath(os.path.join(game_dir, '..'))
+        relative = os.path.relpath(game.install_dir, canonical_prefix)
+        os.symlink(relative, game_dir)
```
sharkwouter, line 207: What does this do?

sharkwouter (changes requested):
Hey @GB609, thanks again for this PR. I found some small things, but this is looking pretty good. I added one little note for myself and some for you. This does look like a big improvement, though.

sharkwouter:
I think I found an issue with the shortcut fix in here:
```
trying to run wine command: env 'WINEPREFIX=/home/wouter/GOG Games/Fallout Classic/prefix' WINEDLLOVERRIDES=winemenubuilder.exe=d /usr/bin/wine '/home/wouter/.cache/minigalaxy/download/Fallout Classic/setup_fallout_2.1.0.18.exe' '/DIR=c:\game' '/LOG=c:\install.log' /LANG=en '/SAVEINF=c:\setup.inf' /SILENT
executing wine command: env 'WINEPREFIX=/home/wouter/GOG Games/Fallout Classic/prefix' WINEDLLOVERRIDES=winemenubuilder.exe=d /usr/bin/wine '/home/wouter/.cache/minigalaxy/download/Fallout Classic/setup_fallout_2.1.0.18.exe' '/DIR=c:\game' '/LOG=c:\install.log' /LANG=en '/SAVEINF=c:\setup.inf' /SILENT
002c:err:wineboot:process_run_key Error running cmd L"C:\\windows\\system32\\winemenubuilder.exe -a -r" (126).
007c:fixme:hid:handle_IRP_MN_QUERY_ID Unhandled type 00000005
007c:fixme:hid:handle_IRP_MN_QUERY_ID Unhandled type 00000005
007c:fixme:hid:handle_IRP_MN_QUERY_ID Unhandled type 00000005
007c:fixme:hid:handle_IRP_MN_QUERY_ID Unhandled type 00000005
0024:fixme:thread:GetThreadUILanguage : stub, returning default language.
0024:fixme:nls:RtlGetThreadPreferredUILanguages 00000038, 0054D998, 00000000 0054D9C0
0024:fixme:nls:get_dummy_preferred_ui_language (0x38 0x409 0054D998 00000000 0054D9C0) returning a dummy value (current locale)
0024:fixme:nls:RtlGetThreadPreferredUILanguages 00000038, 0054D998, 015519B0 0054D9C0
0024:fixme:nls:get_dummy_preferred_ui_language (0x38 0x409 0054D998 015519B0 0054D9C0) returning a dummy value (current locale)
0108:fixme:win:DisableProcessWindowsGhosting : stub
0108:fixme:graphics:ShutdownBlockReasonDestroy (00010094): stub
0108:fixme:graphics:ShutdownBlockReasonCreate (00010094, L"Installing"): stub
0108:fixme:rstrtmgr:RmStartSession 0050D024, 0, 0050D028 stub!
0108:fixme:graphics:ShutdownBlockReasonDestroy (00010094): stub
0108:fixme:graphics:ShutdownBlockReasonCreate (00010094, L"Installing Fallout."): stub
0108:fixme:msg:ChangeWindowMessageFilterEx 000200A2 c04e 1 00000000
0108:fixme:msg:ChangeWindowMessageFilterEx 000300A2 c04e 1 00000000
0108:fixme:msg:ChangeWindowMessageFilterEx 000100AC c04e 1 00000000
0108:fixme:msg:ChangeWindowMessageFilterEx 000200AC c04e 1 00000000
0108:fixme:shell:SHAutoComplete stub
0108:fixme:msg:ChangeWindowMessageFilterEx 000300B2 c04e 1 00000000
0108:fixme:wincodecs:jpeg_decoder_get_metadata_blocks stub
0108:fixme:gdiplus:resample_bitmap_pixel Unimplemented interpolation 6
0108:fixme:rstrtmgr:RmEndSession 3735928559 stub!
0108:fixme:graphics:ShutdownBlockReasonDestroy (00010094): stub
command finished, read remaining output (if any)
done


Unattended install failed. Try install with wizard dialog.
```

sharkwouter:
I would suggest leaving out the menu shortcuts fix for now. We can figure that out later.

sharkwouter:
This change does actually fix 1nsane, which is really cool to see.

sharkwouter:
You might want to prevent Innoextract from being used honestly. It looks like with this change innoextract is being executed and then fully ignored.

sharkwouter:
I found another bug. If the innoextract installation works, it is not possible to launch the game anymore with this change, since the game files are not inside the prefix and the new code does not account for that.

minigalaxy/installer.py
GB609: I had some of them as logger.info before. but it didnt show on the console by default where i wanted it to. that's why i used print. Easy enough to change

minigalaxy/launcher.py
```diff
@@ -167,8 +172,8 @@ def get_scummvm_exe_cmd(game, files):
     return ["scummvm", "-c", scummvm_config]
 
 
-def get_start_script_exe_cmd():
-    return ["./start.sh"]
+def get_start_script_exe_cmd(game):
+    return [os.path.join(game.install_dir, "start.sh")]
```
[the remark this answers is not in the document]
GB609: When launching through Minigalaxy. 
But what happens when a .desktop file is created for the game during installation? It's exec command would then be './start.sh'
Under the assumption that the Path entry is set and honored by the respective launcher, it will also work the ./start.sh. I wasn't too sure about the support and behavior there.

minigalaxy/installer.py
GB609: I'm already setting up POpen in text mode and encoding 'utf-8'. Is that not enough?

minigalaxy/installer.py
GB609: As the temporary directory before, the location of the game within the wineprefix is just a link to the real installation dir.
The problem with that is that 'c:' itself ALSO is a link to a linux path placed somewhere also created by wine, maybe some path up or down. Appears to be ../drive_c currently. I didn't want to take this for granted. 
So i try to fully resolve the real path install_dir/prefix/c:/game and then convert this to the true relative path that 'game' must be to correctly point to install_dir

minigalaxy/installer.py
GB609: It shouldnt. I've setup the Default arguments of _exe_cmd to be captureOutput=true and print=False which will lead to both stdout and stderr returned here containing all output.
I can also add a test case to verify that works if you like to

minigalaxy/launcher.py
```diff
@@ -130,10 +126,11 @@ def get_windows_exe_cmd(game, files):
                         if "category" in task and task["category"] == "game" and "path" in task:
                             working_dir = task["workingDir"] if "workingDir" in task else "."
                             path = task["path"]
-                            exe_cmd = [get_wine_path(game), "start", "/b", "/wait", "/d", working_dir,
-                                       path]
+                            exe_cmd = [get_wine_path(game), "start", "/b", "/wait",
+                                       "/d", f'c:\\game\\{working_dir}',
```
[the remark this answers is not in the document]
GB609: I've had a look into some of the gog*info files used in the test. It appears as if the workingDir property as well as the executable path always are relative paths.
The documentation for start is pretty sparse.
start runs within wine. It's normally expecting windows type paths. If you don't specify an absolute path, it will resolve with the current working directory.
Anything coming from linux is normally translated to z:\, e.g. z:\home\user\GOG Games\somegame. While it may be the same directory in the linux workspace, for wine they are different.
That's why i made it absolute here. If you happen to know or own a game that requires a different working directory, please let me know and i see if can also test and verify that.

minigalaxy/launcher.py
```diff
@@ -142,7 +139,15 @@ def get_windows_exe_cmd(game, files):
         filename = os.path.splitext(os.path.basename(executables[0]))[0] + '.exe'
         exe_cmd = [get_wine_path(game), filename]
 
-    return exe_cmd
+    # Backwards compatibility with windows games installed before installer fixes.
+    # Will not fix games requiring registry keys, since the paths will already
+    # be borked through the old installer.
+    gamelink = os.path.join(prefix, 'dosdevices', 'c:', 'game')
+    if not os.path.exists(gamelink):
+        os.makedirs(os.path.join(prefix, 'dosdevices', 'c:'))
```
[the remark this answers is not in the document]
GB609: Because this quickly becomes a maintenance and testing hell. There's also the issue of cyclomatic complexity. I always try to avoid deep nesting if simple ifs can instead be used to make sure a certain requirement is given before actually using a common piece of code.

minigalaxy/installer.py
```diff
@@ -349,12 +374,34 @@ def uninstall_game(game):
...
 
 
-def _exe_cmd(cmd):
-    process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    stdout, stderr = process.communicate()
-    stdout = stdout.decode("utf-8")
-    stderr = stderr.decode("utf-8")
-    return stdout, stderr, process.returncode
+def _exe_cmd(cmd, capture_output=True, print_output=False):
+    """Wine commands are very verbose.
```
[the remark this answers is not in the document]
GB609: Seems to be a leftover is my own merge and cherrypicking tango between my branches.
Will do.

minigalaxy/installer.py
sharkwouter: Okay, if the link is within the game's installation directory that is fine. I was worried that it was linking to where the game was extracted.

GB609:
> I think I found an issue with the shortcut fix in here:
> 
> ```
> trying to run wine command: env 'WINEPREFIX=/home/wouter/GOG Games/Fallout Classic/prefix' WINEDLLOVERRIDES=winemenubuilder.exe=d /usr/bin/wine '/home/wouter/.cache/minigalaxy/download/Fallout Classic/setup_fallout_2.1.0.18.exe' '/DIR=c:\game' '/LOG=c:\install.log' /LANG=en '/SAVEINF=c:\setup.inf' /SILENT
> executing wine command: env 'WINEPREFIX=/home/wouter/GOG Games/Fallout Classic/prefix' WINEDLLOVERRIDES=winemenubuilder.exe=d /usr/bin/wine '/home/wouter/.cache/minigalaxy/download/Fallout Classic/setup_fallout_2.1.0.18.exe' '/DIR=c:\game' '/LOG=c:\install.log' /LANG=en '/SAVEINF=c:\setup.inf' /SILENT
> 002c:err:wineboot:process_run_key Error running cmd L"C:\\windows\\system32\\winemenubuilder.exe -a -r" (126).
> 007c:fixme:hid:handle_IRP_MN_QUERY_ID Unhandled type 00000005
> 007c:fixme:hid:handle_IRP_MN_QUERY_ID Unhandled type 00000005
> 007c:fixme:hid:handle_IRP_MN_QUERY_ID Unhandled type 00000005
> 007c:fixme:hid:handle_IRP_MN_QUERY_ID Unhandled type 00000005
> 0024:fixme:thread:GetThreadUILanguage : stub, returning default language.
> 0024:fixme:nls:RtlGetThreadPreferredUILanguages 00000038, 0054D998, 00000000 0054D9C0
> 0024:fixme:nls:get_dummy_preferred_ui_language (0x38 0x409 0054D998 00000000 0054D9C0) returning a dummy value (current locale)
> 0024:fixme:nls:RtlGetThreadPreferredUILanguages 00000038, 0054D998, 015519B0 0054D9C0
> 0024:fixme:nls:get_dummy_preferred_ui_language (0x38 0x409 0054D998 015519B0 0054D9C0) returning a dummy value (current locale)
> 0108:fixme:win:DisableProcessWindowsGhosting : stub
> 0108:fixme:graphics:ShutdownBlockReasonDestroy (00010094): stub
> 0108:fixme:graphics:ShutdownBlockReasonCreate (00010094, L"Installing"): stub
> 0108:fixme:rstrtmgr:RmStartSession 0050D024, 0, 0050D028 stub!
> 0108:fixme:graphics:ShutdownBlockReasonDestroy (00010094): stub
> 0108:fixme:graphics:ShutdownBlockReasonCreate (00010094, L"Installing Fallout."): stub
> 0108:fixme:msg:ChangeWindowMessageFilterEx 000200A2 c04e 1 00000000
> 0108:fixme:msg:ChangeWindowMessageFilterEx 000300A2 c04e 1 00000000
> 0108:fixme:msg:ChangeWindowMessageFilterEx 000100AC c04e 1 00000000
> 0108:fixme:msg:ChangeWindowMessageFilterEx 000200AC c04e 1 00000000
> 0108:fixme:shell:SHAutoComplete stub
> 0108:fixme:msg:ChangeWindowMessageFilterEx 000300B2 c04e 1 00000000
> 0108:fixme:wincodecs:jpeg_decoder_get_metadata_blocks stub
> 0108:fixme:gdiplus:resample_bitmap_pixel Unimplemented interpolation 6
> 0108:fixme:rstrtmgr:RmEndSession 3735928559 stub!
> 0108:fixme:graphics:ShutdownBlockReasonDestroy (00010094): stub
> command finished, read remaining output (if any)
> done
> 
> 
> Unattended install failed. Try install with wizard dialog.
> ```

Where is the issue with the shortcut here? This is the installer you are running. 
Shortcuts come much later.
This is just the output of the first wine installation command. As commented somewhere else, wine itself is prety verbose.

minigalaxy/installer.py
sharkwouter: Oh, I forgot about that. It seems that the default log level is error. If you set the environment variable MG_DEBUG it will be set to debug. Honestly, stdout should use print. Everything else should be logs.

We might want to consider changing the default log level.

GB609:
> I found this: https://documentation.help/Inno-Setup/topic_setupcmdline.htm
> 
> We can actually use the `/NOICONS` option to make sure the installers do not create menu shortcuts.

I saw this as well. But i wasn't sure how many of the installers actually support that switch (or most of the others). After i got some issues with the /SILENT switch for some games i tried to use only as few setup switches as possible. And the same effect of the 'exporting' the windows shortcuts can be achieved with winemenubuilder=d, so why bother with uncertain windows apis?

minigalaxy/installer.py
```diff
@@ -349,12 +374,34 @@ def uninstall_game(game):
...
+    rc = process.poll()
+    while rc is None:
+        out_line = process.stdout.readline()
+        if capture_output and out_line != '':
+            std_out.append(out_line)
+
+        if print_output:
+            print(out_line, end='')
+
+        rc = process.poll()
```
sharkwouter, line 394: Can you add a sleep to this loop? Minigalaxy is freezing up.

GB609: I was actually thinking about this as well. The question is if that is necessary because readline blocks.
The freezing possibly happens because i'm not forking out another thread here and install might come from (or run in) the ui thread.

minigalaxy/launcher.py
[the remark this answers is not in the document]
sharkwouter: We do set the Path entry in desktop shortcuts actually.

minigalaxy/installer.py
```diff
@@ -349,12 +374,34 @@ def uninstall_game(game):
...
+
+    print('command finished, read remaining output (if any)')
+    for line in process.stdout.readlines():
+        std_out.append(line)
+    print('done')
+
+    process.stdout.close()
+
+    output = ''.join(std_out)
+    return output, output, rc
```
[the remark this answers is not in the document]
sharkwouter: Was the old _exe_cmd function causing crashes?

GB609:
> I found another bug. If the innoextract installation works, it is not possible to launch the game anymore with this change, since the game files are not inside the prefix and the new code does not account for that.

It does.
That is what actually happens in launcher.py on lines 144 - 147 which you commented above. Here, any prefix that is run will be checked if the c:\game exists. If it does not, it created before the actual launch command is executed.
But i can see if there's a way to test this.

GB609:
> You might want to prevent Innoextract from being used for extraction honestly. It looks like with this change innoextract is being executed to extract the game and then the extracted files are fully ignored.

I've got another code addition that makes usage of inno configurable. Just didn't include it here to keep the changes as small as possible. 

But why do you think the extracted files are ignored? I haven't changed to entry point to windows installation in installer.py
```python
def extract_windows(game: Game, installer: str, temp_dir: str, language: str, use_innoextract: bool):
    err_msg = extract_by_innoextract(installer, temp_dir, language, use_innoextract)
    if err_msg:
        err_msg = extract_by_wine(game, installer, temp_dir)
    return err_msg

```

And i didn't really touch the innoextract method. If it does not create an error, it will just skip over wine.
But i'll try to run a manual test for that to verify that.

GB609:
> This change does actually fix 1nsane, which is really cool to see.

I tried the changes to the installer with as many games as i could without going bankrupt for now. If you followed the game test table in #622 you could see what i've tested with. Even asked for help in terms of install test reports.
I'm doing this exactly because i'm aware how impactful of a change this will be.

GB609:
> I would suggest leaving out the menu shortcuts fix for now. We can figure that out later.

If i leave it out, shortcuts created for wine games will break and not work anymore unless i let the code generate a wrapper start.sh script to use instead. I've some code for this lying around and thought about whether that actually is worth it and a good idea. But that would make this PR again more complicated.

minigalaxy/installer.py
```diff
@@ -179,40 +181,69 @@ def extract_by_innoextract(installer: str, temp_dir: str, language: str, use_inn
...
-    (difference: : exposes directory, :: exposes the block device itself)
-    But they can't exist at the same time within a prefix.
-    Changing this letter is a temporary fix, the entire install method requires an overhaul in the long run"""
-    drive = os.path.join(prefix_dir, "dosdevices", "t:")
+    game_dir = os.path.join(prefix_dir, "dosdevices", 'c:', 'game')
+    wine_env = [
+        f"WINEPREFIX={prefix_dir}",
+        "WINEDLLOVERRIDES=winemenubuilder.exe=d"
+    ]
+    wine_bin = shutil.which('wine')
```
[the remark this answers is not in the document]
GB609: True for now. Because there only is one wine. Will do.

minigalaxy/installer.py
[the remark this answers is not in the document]
GB609: Not a crash, but a freeze or deadlock.
I tried to install 'lethis - path of progress' i think and it just didn't proceed in the post-install step that installs directx.
The problem with the old variant is that it just collected/buffered the entire process output first. But if the output of the process is bigger than the buffer, the process is blocked until it can write again and running directx install in wine generates a lot of output.
After i changed the _exe_cmd to consume buffers, the installation that failed before ran through without a problem.

minigalaxy/launcher.py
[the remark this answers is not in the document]
GB609: Yeah. i wasn't doubting that. Did see it.

What i meant is that Path is *optional* and depending on the application that actually evaluates these shortcuts, Path may not be parsed and used.
I had stuff like es-de in mind here which do a manual parsing of desktop files. And since there's more than a dozen different desktop environments there's bound to be one that ignores Path, either deliberately or because of a bug.

minigalaxy/installer.py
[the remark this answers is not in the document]
GB609: There's a bit more to do here. The installer could also run for patches, updates or dlc on pre-installed games. Which means that the game might have a custom_wine path property value. In that case we must use it here too. Changing wine executables in an already configured prefix does not always work and can on the contrary completely break the prefix.

GB609:
> There was one game which I tested for which innoextract did not fail and then the installer was used afterwards to do the installation. The innoextract files were not deleted. I'm sorry about not remembering which game that was.

No, it's fine. This part needs a rework or addition. I mostly focused on wine and don't even have inno installed.

sharkwouter:
You should install innoextract. It is also used for determining what languages are supported by the installer. If you don't have it, you'll probably always get English, not 100% sure. It has been ages since that code was added.

sharkwouter:
I don't think we need to support both if using the installer works fine honestly, but we do use Innoextract for language stuff here: https://github.com/sharkwouter/minigalaxy/blob/a2c04d30b7027546bffe2b91a20ff3d79cd3ff76/minigalaxy/installer.py#L376

GB609:
> I don't think we need to support both if using the installer works fine honestly, but we do use Innoextract for language stuff here:
> 
> https://github.com/sharkwouter/minigalaxy/blob/a2c04d30b7027546bffe2b91a20ff3d79cd3ff76/minigalaxy/installer.py#L376-L395

Exactly. But the method lang_install you point out is only called from extract_by_innoextract itself. There is no other occurrence yet. Maybe i can leverage that could to help search for the game language.

I'm fine with dropping innoextract for installation otherwise. Windows games must run within wine anyway. If the install already fails, successfully running a manually extracted game will always a be a bit of a gamble.

sharkwouter:
I just found that /NOICONS does not work unfortunately. We might have to remove the shortcut files manually.

GB609:
> I just found that /NOICONS does not work unfortunately. We might have to remove the shortcut files manually. That is unrelated to this PR, though.

That's why i set winemenubuilder=d for the wine command running the install. The windows shortcuts will be generated by the install and exists in the wineprefix directory, but they will NOT be exported to the linux application menu.

sharkwouter:
> > I just found that /NOICONS does not work unfortunately. We might have to remove the shortcut files manually. That is unrelated to this PR, though.
> 
> That's why i set winemenubuilder=d for the wine command running the install. The windows shortcuts will be generated by the install and exist in the wineprefix directory, but they will NOT be exported to the linux application menu.

In my tests this causes wine to crash when it tries to create shortcuts, though. Does wine have an option that prevents this?

sharkwouter:
Okay, my bad, seems the crash was unrelated to the error about creating shortcuts.

minigalaxy/installer.py
sharkwouter: Oh, I didn't notice. This is fine then.

minigalaxy/installer.py
sharkwouter, line 394: What about something like this instead?
```suggestion
    while (rc := process.poll()) is None:
        if data := process.stdout.read():
            stdout.append(data)
            print(data, end='')
        if data := process.stderr.read():
            stdout.append(data)
            print(data, end='')
```

minigalaxy/installer.py
```diff
@@ -349,12 +374,34 @@ def uninstall_game(game):
...
-    stderr = stderr.decode("utf-8")
-    return stdout, stderr, process.returncode
+def _exe_cmd(cmd, capture_output=True, print_output=False):
+    """Wine commands are very verbose.
+    We should consume the output in regularly to prevent buffers (and minigalaxy memory) from filling up"""
+    print(f'executing wine command: {shlex.join(cmd)}')
+    std_out = []
+    process = subprocess.Popen(cmd,
+                               stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
+                               bufsize=1, universal_newlines=True, encoding="utf-8")
```
sharkwouter, line 384: With the suggestion below, I would direct stderr to its own pipe:
```suggestion
                               stdout=subprocess.PIPE, stderr=subprocess.PIPE,
                               bufsize=1, universal_newlines=True, encoding="utf-8")
```

minigalaxy/installer.py
```diff
@@ -349,12 +374,34 @@ def uninstall_game(game):
...
-    process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    stdout, stderr = process.communicate()
-    stdout = stdout.decode("utf-8")
-    stderr = stderr.decode("utf-8")
-    return stdout, stderr, process.returncode
+def _exe_cmd(cmd, capture_output=True, print_output=False):
+    """Wine commands are very verbose.
+    We should consume the output in regularly to prevent buffers (and minigalaxy memory) from filling up"""
+    print(f'executing wine command: {shlex.join(cmd)}')
+    std_out = []
```
sharkwouter, line 381:
```suggestion
    std_out = []
    std_err = []
```

minigalaxy/installer.py
sharkwouter: Let me do a bit of testing with this.

sharkwouter: I made a PR to your branch which has an alternative solution for this based on your work. Maybe that's the solution.

tests/test_installer.py
```diff
@@ -281,29 +280,12 @@ def test_get_game_size_from_unzip(self, mock_subprocess):
...
-        mock_subprocess().communicate.return_value = [stdout, b"stderr"]
+        mock_subprocess().communicate.return_value = [stdout, "stderr"]
         installer_path = "/home/i/.cache/minigalaxy/download/Beneath a Steel Sky/beneath_a_steel_sky_en_gog_2_20150.sh"
         exp = 159236636
         obs = installer.get_game_size_from_unzip(installer_path)
         self.assertEqual(exp, obs)
 
-    @mock.patch('shutil.which')
-    @mock.patch('os.listdir')
-    def test_get_exec_line(self, mock_list_dir, mock_which):
```
sharkwouter, line 292: Why was this test removed?

minigalaxy/launcher.py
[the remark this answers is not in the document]
sharkwouter: Okay, I guess there is no problem with that.

minigalaxy/launcher.py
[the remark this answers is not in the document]
sharkwouter: Makes sense, thanks for clarifying.

minigalaxy/installer.py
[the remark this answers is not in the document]
sharkwouter: This is addressed by my PR: https://github.com/GB609/minigalaxy/pull/2

minigalaxy/installer.py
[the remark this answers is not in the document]
sharkwouter: This sounds like a separate bug. Good to know, though. I'll update the dlc install issue. Maybe this is related.

README.md
sharkwouter: Lets just make sure the linking works in general by testing it well.

I'll ask you to remove this second from the README, since Minigalaxy has never worked with registry key installation through the installer, since it would move the files afterwards.

minigalaxy/installer.py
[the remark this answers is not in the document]
sharkwouter: Also addressed in my PR: https://github.com/GB609/minigalaxy/pull/2

minigalaxy/installer.py
sharkwouter: I've created an issue for this. I'll close this one.

sharkwouter (changes requested):
Okay, I looked into it once more and did some more testing. What I found is that most of my comments were either unrelated or you clarified them.

I also created a merge request for your branch to resolve some issues with the exe_cmd function. I hope that is welcomed and helps you out.

If the left over comments are resolved, I will merge this. Then we can look into left over wine related issues. I don't think this PR is creating new ones at the moment apart from the ones by the _exe_cmd change.

sharkwouter:
I found that the reason for a successful innoextract being seen as a failure is that the check to see if it failed only looks at if the stderr contains something. If it does it fails. This is addressed in the change I made in the PR I linked. Otherwise this seems to be a big improvement. I just wanted to clarify that.

I would also like to thank you for pushing me to take another look at this code. There are some issues in it which we are finding now because of that.

tests/test_installer.py
GB609: Because the method it tests doesnt exist anymore. If replaced the self-coded shell masking with shlex.join(cmd) in the caller of get_exec_line

GB609:
> I found that the reason for a successful innoextract being seen as a failure is that the check to see if it failed only looks at if the stderr contains something. If it does it fails. This is addressed in the change I made in the PR I linked. Otherwise this seems to be a big improvement. I just wanted to clarify that.
> 
> I would also like to thank you for pushing me to take another look at this code. There are some issues in it which we are finding now because of that.

Thanks for the feedback. Glad it's useful.
About the stderr separation. Can i suggest a simpler way?
How about i keep the line buffer and readline loop of stdout and use stderr.readlines() after the process returned to consume all of stderr at once?
Buffer size should be less of a problem for error output since it usually is smaller.

sharkwouter:
I was thinking about that, since stderr output will be smaller. We could just do stderr.read at the end. I'll update my PR to your branch.

tests/test_installer.py
sharkwouter: Makes sense, will close this one then

sharkwouter (approved):

sharkwouter:
Thanks!

PR #629, opened 2024-12-11
critical bug fix in installer

GB609:
## Description

My last pullrequest with the reworked wine installer contains a critical bug that prevents wine installations from working:

_exe_cmd might not consume all of the subprocess stdout because the loop exits as soon as the subprocess has a return code, but only 1 line per poll() check is consumed, so there might still be more lines in the buffer. These must also be consumed, especially when the output is required, as is the case for lang_install.

The output of innoextract --list-languages is very short and it returns quickly, so the first poll() might already return 0 and just 1 line will be consumed, which will be 'Inspecting "Game" ...', the actual lines listing languages come afterwards.

As lang_install is now used in extract_wine, there are situations where the return of lang_install will not be in the expected format '--language=langname', therefore the split operation fails with an error and the install aborts.

This happened because of my last-minute changes to lang_install and innoextract installation  introduced when fixing review comments.

This PR fixes this in 2 places:
1. Make _exe_cmd consume everything
2. make sure there will always be a default return from lang_install

## Checklist
 
### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -402,6 +402,17 @@ def _exe_cmd(cmd, print_output=False):
                 print(data, end='')
         time.sleep(0.01)
 
+    # there might some lines left in the buffer, get them all
+    data = process.stdout.readlines()
+    std_out += ''.join(data)
+    if print_output:
+        print(data, end='')
+
+    data = process.stderr.readlines()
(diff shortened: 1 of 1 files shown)


sharkwouter:
Good catch, I forgot about it reading only one line after changing it to readline.

sharkwouter (approved):

PR #632, opened 2024-12-25
Desktop file creation

GB609:
## Description

While testing, I discovered that I've introduced a new error with my previous PR #624:
The change in how the value for the 'Exec' key in .desktop files is generated. 

I assumed that Exec uses quoting rules identical to shells. However, it does not. Only " can be used for quotes and there are less special characters. shlex.join quotes differently, it uses single quotes ' exclusively which is not compatible with how the Exec key is supposed to be quoted.

I restored the get_exec_line method in installer.py which i previously deleted, and enhanced it instead to conform to the quoting rules as described in [The Exec Key](https://specifications.freedesktop.org/desktop-entry-spec/latest/exec-variables.html).

Sorry about this oversight on my part.

## Checklist
 
### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -6,6 +6,7 @@ import subprocess
 import hashlib
 import textwrap
 import time
+import re
 
 from minigalaxy.config import Config
 from minigalaxy.game import Game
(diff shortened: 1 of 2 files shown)


sharkwouter (approved):

sharkwouter:
Thanks! This looks good! For the other PR I'll still need some time to review it, but I have seen it and I really appreciate the work.

GB609:
Sure, take all the time you need. It's the end of the year and holiday time after all.

PR #636, opened 2025-01-05
Fix Unreal Gold being unable to launch

sharkwouter:
## Description

Right now if the executable is one level deeper, the creation of shortcuts and launching the game will fail. This resolves this at least for Unreal Gold, but probably also for some other games

## Checklist
 
### DIFF
--- a/minigalaxy/launcher.py
+++ b/minigalaxy/launcher.py
@@ -149,7 +149,7 @@ def get_windows_exe_cmd(game, files):
         executables.remove(os.path.join(game.install_dir, "unins000.exe"))
         if not executables:
             # Look one directory level deeper
-            executables = glob.glob(game.install_dir + '/**/*.exe')
+            executables = glob.glob(game.install_dir + '/*/*.exe')
         filename = os.path.relpath(executables[0], game.install_dir)
         exe_cmd = [get_wine_path(game), filename]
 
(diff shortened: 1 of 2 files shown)


minigalaxy/launcher.py
```diff
@@ -147,7 +147,10 @@ def get_windows_exe_cmd(game, files):
         # in case no goggame info file was found
         executables = glob.glob(game.install_dir + '/*.exe')
         executables.remove(os.path.join(game.install_dir, "unins000.exe"))
-        filename = os.path.splitext(os.path.basename(executables[0]))[0] + '.exe'
+        if not executables:
+            # Look one directory level deeper
+            executables = glob.glob(game.install_dir + '/**/*.exe')
```
GB609, line 152: Python glob ** does not just search one level deeper, it traverses as far as it can.
This will become a problem here because of
`install_dir/prefix/dosdevices`. That directory also contains a link named z: pointing to the linux root fs /.
You effectively be searching tnrough the entire mounted file system here which will take a long time and produce a very large list of files.

sharkwouter: I'll change it, because I do think it is less clear like this.

sharkwouter: The issue with finding more than one exe we already have at the moment. We'll have to do a change for it to work better at some point, but I would say that that is independent of this change.

PR #644, opened 2025-02-11
remove duplicate code step 3 - introduce barebones LibraryEntry

GB609:
Step 3 of the refactoring i started with #642.

This more or less marks the beginning of the big changes for real. There likely won't be coming many more small PRs, but I'll try to see what i can do. And the last one for now (this morning).

@sharkwouter Sorry if i'm flooding you with too many PRs in quick sucession. Just take your time, i don't want to rush you. 

Changes:
1. A skeleton superclass for GameTile(List). It already contains 2 methods from the subclasses that didn't need any of the private instance attributes. And it also shows how i plan to manage interaction between GTK -> subclasses -> common code in places where annotations are involved.
2. Reduced constructor arguments - there is no need to pass around stuff like config and api as individual arguments if they are part of the parent object that also gets passed in anyway.
3. Above changes also lead to some moving around of imports.

I've also had another look at the sort. There are small misunderstandings:
1. on the way it is used in general in the current code
2. and my interpretation of the impact on performance

The sort isn't manually retrigged all the time, instead the GTK api that configures a sort function for the list is used. The rest is completely up to GTK everytime the list is manipulated. It does this automatically and the to only way to interfere would be to temporarily disable sort or unset/change the sort function.
And it's currently only done when adding games, there are no preferences or options to change the sort order.
Therefore, it is enough to call 'sort_library' once during Library construction and drop all repeated calls otherwise. This still means that some unnecessary items on the queue can be avoided.
However, the impact on performance likely is much less severe as i thought initially, at least for the sorting aspect.

Sorry about the confusion.

### DIFF
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -4,10 +4,8 @@ import threading
 import time
 import urllib.parse
 
-from minigalaxy.api import NoDownloadLinkFound, Api
-from minigalaxy.config import Config
+from minigalaxy.api import NoDownloadLinkFound
 from minigalaxy.download import Download, DownloadType
-from minigalaxy.download_manager import DownloadManager
 from minigalaxy.entity.state import State
 from minigalaxy.game import Game
 from minigalaxy.installer import uninstall_game, install_game, check_diskspace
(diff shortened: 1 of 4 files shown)


minigalaxy/ui/gametile.py
```diff
@@ -16,12 +14,11 @@
 from minigalaxy.paths import CACHE_DIR, THUMBNAIL_DIR, ICON_DIR, UI_DIR, DOWNLOAD_DIR, ICON_WINE_PATH
 from minigalaxy.translation import _
 from minigalaxy.ui.gtk import Gtk, GLib, Notify
-from minigalaxy.ui.information import Information
-from minigalaxy.ui.properties import Properties
+from minigalaxy.ui.library_entry import LibraryEntry
 
 
 @Gtk.Template.from_file(os.path.join(UI_DIR, "gametile.ui"))
-class GameTile(Gtk.Box):
```
sharkwouter, line 24: This could be a bit risky, but it does make sense. Lets monitor if the double inheritance causes any issues later.

sharkwouter (approved):

sharkwouter:
No worries, I appreciate the work. Putting it in smaller pull requests makes it much easier to review as well. Good work so far.

minigalaxy/ui/gametile.py
GB609: As long as we don't also put some GTK class as superclass of LibraryEntry, it should be fine.
An alternative would be to let LibraryEntry inherit Gtk.Box.
But i deliberately chose not to to keep the future possibility that another view implementation to be done might need another container class for games.

PR #645, opened 2025-02-12
remove duplicate code step 4 - eliminate differences in GameTile(List)

GB609:
Step 4 of the refactoring i started with #642.

1. Common variable setup is moved to `LibraryEntry.__init__`
2. Prepared a method `init_ui_elements` in subclasses. This already contains a lot of common code for both. Will move the common part up in one of the next steps
3. progress_bar in GameTileList is now created once only. With that, the repeated 'if progress_bar: destroy or create' lines can be replaced by a simple `progress_bar.hide()` or `progress_bar.show()` which is identical to GameTile.
4. Use the same ui labels for GameTileList as they are used by GameTile ('updating...' -> 'Updating...')
5. Small change to how the library calls `show` on newly added game widget. `show_all` overrules the `progress_bar.hide()` statements from GameTileList, so it can't be used.

### DIFF
--- a/minigalaxy/ui/gametile.py
+++ b/minigalaxy/ui/gametile.py
@@ -11,7 +11,7 @@ from minigalaxy.game import Game
 from minigalaxy.installer import uninstall_game, install_game, check_diskspace
 from minigalaxy.launcher import start_game
 from minigalaxy.logger import logger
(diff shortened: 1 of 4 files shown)


sharkwouter (approved):

sharkwouter:
Thanks! This looks like it works well

PR #648, opened 2025-02-14
remove duplicate code step 6 - state handling

GB609:
Step 6 of the refactoring i started with #642.

If there are no unforeseen problems, it'll be just one (maybe 2) more, then GameTile and GameTileList won't be much more than carriers for GTK template annotations with slight differences in ui behavior. There's still a bit more i would want to do afterwards, but that chain of PRs linked to each other will come to an end. 

Once all are done, i'll again do a full manual test of the ui flow in both views (of course i'm always also doing quick tests)

It's been quite a tough series of PRs and i'm admittedly moving onwards at quite a pace. Thanks for working with me on this.

Changes this time:
The state handling is now common behavior and was moved to LibraryEntry.
It's basically an application of template pattern. LibraryEntry defines the hook names and some defaults. Subclasses can use them as-is by not defining them, enhance or replace them.

I've had to pull 2 more private functions up into LibraryEntry, because those are called from a state change directly. They are related to checking for updates or dlc.
Any other private will follow in the next PR.


minigalaxy/ui/library_entry.py
```diff
@@ -89,6 +101,34 @@ def get_download_info(self, platform="linux"):
             result = False
         return result, download_info
 
+    def __check_for_update_dlc(self):
```
sharkwouter, line 104: These functions cannot be private anymore, otherwise the child class won't be able to access them. I'd suggest renaming them to have only a single underscore in front of them.

I found the following error after installing a game:
```
Exception in thread Thread-201 (finish):
  File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.11/threading.py", line 975, in run
  File "/home/wouter/Personal/minigalaxy/minigalaxy/download.py", line 59, in finish
    self.__finish_func(self.save_location)
  File "/home/wouter/Personal/minigalaxy/minigalaxy/ui/gametile.py", line 177, in wrapper
    finish_func(self.download_list[-1].save_location)
  File "/home/wouter/Personal/minigalaxy/minigalaxy/ui/gametile.py", line 236, in __install_game
    self.__check_for_dlc(self.api.get_info(self.game))
    ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'GameTile' object has no attribute '_GameTile__check_for_dlc'

```

sharkwouter:
Cool, this one makes a big difference. I've reviewed it and found one issue.

minigalaxy/ui/library_entry.py
GB609: Ah. I see the issue here.
These functions can, in theory, still be private, as long as there are no references to it in the subclasses. 
I didn't see that the download function also uses them, only thought it is the start up.
I can rename them, but the other solution would be to keep pulling up other privates, but then i'd have to do almost all of them at once

GB609: I'll rename them, should be the smaller change.
On the long run, it might be better to change the semantics of those methods and split them into one part that executes the update and a generic utility returning some kind of update check result.

sharkwouter (approved):

sharkwouter:
Looks good now!

PR #652, opened 2025-02-21
Provide a UI for active downloads

GB609:
## Description

I'm currently working on several issues and features related to downloads. These are:
- Some refactoring/code shifting in LibraryEntry to prepare for improvements of download handling and progress calculation.
- investigate need/current capability of minigalaxy to smartly pause/resume partial downloads. From what i can tell, manually cancelling a download will unconditionally delete the downloaded parts. The only way to keep parts is to just close minigalaxy without cancelling.
- The logic that determines which files to delete. It deletes too much right now.
- Some central place in the ui that makes downloads and their progress more visible.
- Making multi-part downloads 'first-class objects' known to download_manager. I've already code something hacked down in LibraryEntry in another branch locally, but i'd like to integrate that to download_manager where it belongs instead.

This (draft) PR is for a new UI: Showing all downloads and their current state.
As a picture often is more powerful than thousands of words, here's a little screenshot of how it looks currently:

screenshot_download_manager

It's still rough and there's a bit of work to do, but i've got most of the basic concept down:
1. download_manager provides an event distribution mechanism to which listeners can register themselves freely. All relevant changes to the download state will be published there. Start, queueing, (manual) cancelling, cancel caused by error, download progress, regular completion.
2. The UI attaches to these events and reacts accordingly
I've chosen that approach because it is more flexible than going to some modded callback hooks in the Download instances themselves, as passing them around to the right places would be an absolute nightmare. Moreover, it's extensible and basically decoupled from the UI.

What's missing:
- [x] The game icon on the left. The api provides game icon files, maybe optional, but it's there. Could be leveraged here.
- [x] Localisation keys
- [x] Hide empty categories
- [x] Change icon in menu bar to `folder-download-symbolic`
- [x] the action buttons/icons could use a bit of highlighting when targeted by the mouse cursor
- [x] The last pieces of event handler code for the download actions pause, ~~resume~~, ~~cancel~~, ~~delete (file)~~
- [x] ~~actions for queued downloads? pause, stop, priority change by dragging?~~ nothing special for now. too complicated because access to the priorised queue is difficult.
- [x] A way in download_manager to differentiate between 'cancel=abort' and 'cancel=pause'. Both are effectively a type of 'cancel', because the download is not actively being continued on a worker thread. Difference between both would be whether the partial file is deleted or not. See #655 
- [x] download_manager should keep track of files that were manually paused/stopped and don't automatically restart them when queued on startup. Might require an addition list of paused downloads and an addition to config to keep track of file names, not just gameids. See #655
- [x] Maybe a bit of spacing corrections in the ui
- [x] ~~Fancy bonus: do something with the button/icon while there are active downloads, e.g. some slow blinking to show that there are some downloads running~~ Menu button changes to `suggested_action` while there is any active or queued download. BUG: not undone correctly
- [x] handling of split downloads (games with several bin files). i'm going to integrate support for this into download_manager itself by using a new subclass of Download and shifting around a bit of the starting/queueing logic to make the difference invisible to download_manager. - **Will follow in another PR**
- [x] Some input, ideas, feedback - ongoing process
- [x] some visual indicator of the different types of 'completion states' (completed, stopped, failure, cancel) which all go into the 'Done' section (aside from the different button options they might have). Maybe background colors or label color, maybe another small icon in front of the name, like a green checkmark or a diagonal red cross, an exlamation mark etc..
- [x] The 'Manage' button just opens a file manager, but the library is not automatically refreshed after closing the manager again. This impacts the case that installers are deleted. In that case the GameTile button which is labeled 'install' won't change back to 'download'. Providing a synchronized subprocess is more than a quick shot here. Properly opening a file manager requires either dbus or xdg-open, both appear to fork, i can't wait for the subprocess to end to refresh the library programmatically. Something to deal with later in a dedicated PR.

How the actions and states differ:

| Action | stream remote > local? | keep partial file | prevent auto-restart | callback behavior | 
| --- | --- | --- | --- | --- |
| Pause  | ✖️ | ✔️ | (✔️) technically GameTile must try to resume them as normal. Necessary to 'transform' gameid file list to compare with list of paused files | paused download never goes active after restart, so no callbacks invoked until resumed |
| Cancel | ✖️ | ✖️ | ✔️ | stopping fires cancel. basically like failure, but different feedback regarding cancel reason. Like cancel on the game tiles. |
| Start/Resume | ✔️ | re-use partial | - | as always. resume from pause requeues download |

## Checklist
 
### DIFF
--- a/minigalaxy/ui/download_list.py
+++ b/minigalaxy/ui/download_list.py
@@ -14,8 +14,13 @@ from minigalaxy.ui.gtk import GLib, Gtk
 class DownloadManagerList(Gtk.Viewport):
     __gtype_name__ = "DownloadList"
 
+    label_active = Gtk.Template.Child()
     flowbox_active = Gtk.Template.Child()
+
+    label_queued = Gtk.Template.Child()
     flowbox_queued = Gtk.Template.Child()
+
+    label_done = Gtk.Template.Child()
     flowbox_done = Gtk.Template.Child()
 
     listener_download_types = [DownloadType.GAME, DownloadType.GAME_UPDATE, DownloadType.GAME_DLC]
@@ -37,8 +42,14 @@ class DownloadManagerList(Gtk.Viewport):
          #   DownloadState.PAUSED : self.download_paused
         }
 
+        self.flowbow_labels = {
+            self.flowbox_active: self.label_active,
+            self.flowbox_queued: self.label_queued,
+            self.flowbox_done: self.label_done
+        }
+
         self.download_manager.add_active_downloads_listener(self.download_manager_listener)
-        self.show_all()
+        self.show()
 
     def download_manager_listener(self, change: DownloadState, download: Download):
         self.logger.debug('Received %s for Download[save_location=%s, progress=%d]',
@@ -76,16 +87,28 @@ class DownloadManagerList(Gtk.Viewport):
 
     def __move_to_section(self, flowbox, entry, new_state: DownloadState):
         if entry.flowbox:
+            old_flowbox = entry.flowbox
             '''the entry needs to be removed from its parent FlowBoxChild
             or there will be a memory access error hard crashing the application'''
             box_child = entry.get_parent()
             box_child.remove(entry)
-            entry.flowbox.remove(box_child)
+            old_flowbox.remove(box_child)
             box_child.destroy()
+            self.__change_group_visibility(old_flowbox)
+
         flowbox.add(entry)
         entry.flowbox = flowbox
         entry.show()
+        self.__change_group_visibility(flowbox)
         entry.update_buttons(new_state)
+
+    def __change_group_visibility(self, group_flowbox):
+        if group_flowbox.get_children():
+            self.flowbow_labels[group_flowbox].show()
+            group_flowbox.show()
+        else:
+            group_flowbox.hide()
+            self.flowbow_labels[group_flowbox].hide()
 
 
 @Gtk.Template.from_file(os.path.join(UI_DIR, "download_list_entry.ui"))
--- a/minigalaxy/ui/download_list.py
+++ b/minigalaxy/ui/download_list.py
@@ -10,6 +10,7 @@ from minigalaxy.download_manager import DownloadManager, DownloadState
 from minigalaxy.paths import UI_DIR
 from minigalaxy.translation import _
 from minigalaxy.ui.gtk import GLib, Gtk
+from gi.overrides.GdkPixbuf import GdkPixbuf
 
 
 @Gtk.Template.from_file(os.path.join(UI_DIR, "download_list.ui"))
@@ -158,6 +159,11 @@ class OngoingDownloadListEntry(Gtk.Box):
                                              remove_panel_action=self.remove_from_current_box,
                                              logger=self.manager.logger)
         self.update_buttons(initial_state)
+
+        self.manager.logger.debug("trying to pull icon from: %s", download.download_icon)
+        if download.download_icon and os.path.exists(download.download_icon):
+            pixbuf = GdkPixbuf.Pixbuf.new_from_file_at_size(download.download_icon, 48, 48)
+            self.icon.set_from_pixbuf(pixbuf)
         self.pack_start(self.buttons, False, False, 0)
 
     def update_progress(self, percentage):
--- a/minigalaxy/ui/download_list.py
+++ b/minigalaxy/ui/download_list.py
@@ -46,7 +46,7 @@ class DownloadManagerList(Gtk.Viewport):
             DownloadState.COMPLETED: self.download_stopped,
             DownloadState.QUEUED: self.download_queued,
             DownloadState.PROGRESS: self.download_progress,
-            DownloadState.FAILED: self.download_stopped,
+            DownloadState.FAILED: self.download_failed,
             DownloadState.CANCELED: self.download_stopped,
             DownloadState.STOPPED: self.download_stopped,
             DownloadState.PAUSED: self.download_paused
@@ -62,7 +62,7 @@ class DownloadManagerList(Gtk.Viewport):
         self.download_manager.add_active_downloads_listener(self.download_manager_listener)
         self.show()
 
-    def download_manager_listener(self, change: DownloadState, download: Download):
+    def download_manager_listener(self, change: DownloadState, download: Download, *additional_params):
         self.logger.debug('Received %s for Download[save_location=%s, progress=%d]',
                           change, download.filename(), download.current_progress)
         if download.download_type not in self.listener_download_types:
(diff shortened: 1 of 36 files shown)


sharkwouter:
This is a great idea. I have actually tried to implement something like this before, but back then it was a bit hard to make happen without separation of gametile and gamelist.

GB609:
> This is a great idea. I have actually tried to implement something like this before, but back then it was a bit hard to make happen without separation of gametile and gamelist.

If you have any suggestions about the UI position, style, capabilities or whatever, just tell me.

Refactoring GameTile and GameTileList before really helped with doing more general ui work. And i'm going to do a few more things there, the get a better handle on the download callbacks.

sharkwouter:
I should use a download icon for the button and hide categories that are not in use.

GB609:
I tried to find an icon that is a bit similar to the firefox download icon, but didn't find one in glade with my gtk theme, so is used `go-down`. Do you happen to know the identifier for the download icon?
Good call about the hiding aspect. I thought about this as well, but delayed it as beautifying work for now as i wanted to get the features down first.

sharkwouter:
This is what the icon in that PR looks like, which is `folder-download-symbolic`.

image

GB609:
Thanks for the compliment.
I spent half the night to debug what i have right now because i had hard crashes caused by memory errors at first. Took a lot of trying and googling to find out that i didn't really remove all necessary items when moving between categories.

sharkwouter:
It's a bit hard to deal with threads and Gtk. When you make a call from a different thread than the main thread to functions that interact with Gtk components you'll see crashes some of the time. That is why `GLib.idle_add` is used so much in library_entry.py.

GB609:
> It's a bit hard to deal with threads and Gtk. When you make a call from a different thread than the main thread to functions that interact with Gtk components you'll see crashes some of the time. That is why `GLib.idle_add` is used so much in library_entry.py.

I'm aware. And i thought at first that this was the issue. But it wasn't.
Right at the beginning of the callback into download_list, im moving to the gtk main thread.

The issue was that i'm using flowboxes and removed my custom gtk boxes from them, but forgot to remove them from the automatically created FlowBoxChildren. So, the second a download changed category, it crashed.

minigalaxy/ui/download_list.py
```diff
@@ -0,0 +1,382 @@
...
+            'actions': [NOOP, trigger_remove],
+            'icons': [None, 'list-remove']
+        },
+        DownloadState.PAUSED: {
+            'actions': [unpause, delete_download],
+            'icons': ['media-playback-start', 'edit-delete']
+        },
+        DownloadState.STOPPED: {
+            'actions': [restart, delete_download],
+            'icons': ['media-playback-start', 'edit-delete']
```
sharkwouter, line 372: Could you use delete or cancel here instead of edit-delete? It looks like this on Debian now:

Screenshot from 2025-03-03 17-34-35

sharkwouter (changes requested):
Thanks for this PR, this is a great feature to have. I did some testing with it and I found 2 things. For one I made a comment, the other one is the following:

When you hover over the name of the current download, it shows a tooltip with the name of the current state in all caps, which is a bit weird. I don't know where this comes from, though.

sharkwouter:
Yeah, I had to change the access to the master branch to Weblate can push to it. Please don't push to master directly, though.

sharkwouter:
Could you run the script to update the translations once? Then it should add the new fields to the translation files.

sharkwouter:
I don't know if manage is the right word for the button. Did you change the download location for installers?

GB609:
> Thanks for this PR, this is a great feature to have. I did some testing with it and I found 2 things. For one I made a comment, the other one is the following:
> 
> When you hover over the name of the current download, it shows a tooltip with the name of the current state in all caps, which is a bit weird. I don't know where this comes from, though.

The caps: 
The origin of this is line 119:
```python
        entry.update_tooltip(_(new_state.name))
```

This is one of the places were localisation keys are needed. For now, i just used the name of the DownloadState enum constant as additional hint of the type of 'completion'. This requires a proper string behind it.

sharkwouter:
Honestly, I don't think the label for the file name needs a tooltip. You can leave that out.

minigalaxy/ui/download_list.py
GB609: Sure, i can change that.

GB609:
> Could you run the script to update the translations once? Then it should add the new fields to the translation files.

Will do.

GB609:
> Yeah, I had to change the access to the master branch to Weblate can push to it. Please don't push to master directly, though.

As i've said, i'm not doing any change to master without a PR & review.
I was just wondering. Does this change only apply to contributors? If not, anyone could theoretically merge whatever. Thats why i was asking

sharkwouter:
>  I was just wondering. Does this change only apply to collaborators? If not, anyone could theoretically merge whatever. Thats why i was asking

Yeah, only people who have been invited to the project can make changes. That is now 3 contributors I trust (including you), Weblate and me.

sharkwouter:
> Didn't change any location. The button is tied to the option 'keep_installers' 

I see, that is clever and really cool. I see what you mean. Maybe it should say 'Manage installers'.

minigalaxy/ui/download_list.py
GB609: One thing however:
The icons for active download aren't those in the line you commented.
For active downloads its `media-playback-pause` and `dialog-cancel`. So `dialog-cancel` is missing for you on debian.

Edit-delete appears when a download was stopped.
Its pretty difficult to find an icon to use when they are so theme related. I tried to stick to the standards only. But even the icon for the download button isnt listed in the standards in glade.

sharkwouter: Ah okay, then edit-delete is working.

GB609: I changed the stop button to `media-playback-stop`. That way, it is more in line with the pause button.

GB609:
I think i've made all the changes you wanted, including the translations.
I did 2 small changes to update-translations itself:
1. Disable fuzzy matching - from what i can tell, it generated nonsense we'd have to manually review either way.
2. added --add-comments to the xgettext invocation for the python code. This allows to provide additional context information for translators.

GB609:
Guess i need to re-create the translation files thanks to an update from weblate.

GB609:
Alright, rebased onto latest from master branch.

sharkwouter:
I'll merge this now before weblate comes in and makes it unmergableagain.

GB609:
> I'll merge this now before weblate comes in and makes it unmergableagain.

I'd just update it again then. Maybe next time i'd not update by rebase, but by merging instead  although i normally try to avoid this.
I find a linear history better for running tests, makes stuff more predictable.

sharkwouter:
What we could do is do translation updates on master instead. Makes things easier to merge.

GB609:
Or just split it out from the ui itself just to separate it from the functional commits

PR #653, opened 2025-02-21
Icons for (windows) desktop shortcuts

GB609:
## Description

This is an alternative to #380 which doesn't require to add more dependencies to minigalaxy.
If `api.get_info()` provides a data entry `['images']['icon']`, the corresponding file is downloaded as `game.install_dir/support/icon.png` which matches the location linux games uses.

This means that i'm effectively ignoring all *.ico files and just download a (hopefully) compatible format.

I've added a few early returns and ifs to protect existing files and prevent error when offline etc, but thats the basic gist of it.
The icon download is currently triggered in 2 places: 
1. after a successful game install
2. when checking for updates
Why both: Game install might also happen while offline from local installer. There'd be no icon then, and no access to download it either.

## Checklist
 
### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -303,6 +303,7 @@ def create_applications_file(game, override=False):
 
         with open(path_to_shortcut, 'w+') as desktop_file:
             desktop_file.writelines(textwrap.dedent(desktop_definition))
+
     except Exception as e:
         os.remove(path_to_shortcut)
         error_message = e
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -303,7 +303,6 @@ def create_applications_file(game, override=False):
 
         with open(path_to_shortcut, 'w+') as desktop_file:
             desktop_file.writelines(textwrap.dedent(desktop_definition))
-
     except Exception as e:
         os.remove(path_to_shortcut)
         error_message = e
--- a/minigalaxy/ui/library_entry.py
+++ b/minigalaxy/ui/library_entry.py
@@ -351,7 +351,7 @@ class LibraryEntry:
     def __check_for_update_dlc(self):
         if self.game.is_installed() and self.game.id and not self.offline:
             game_info = self.api.get_info(self.game)
-            self.__download_icon(False, game_info)
+            self.__download_icon(game_info=game_info)
             if self.game.get_info("check_for_updates") == "":
                 self.game.set_info("check_for_updates", True)
             if self.game.get_info("check_for_updates"):
--- a/minigalaxy/ui/library_entry.py
+++ b/minigalaxy/ui/library_entry.py
@@ -329,6 +329,9 @@ class LibraryEntry:
             GLib.idle_add(self.update_to_state, cancel_to_state)
 
     def __download_icon(self, force=False, game_info=None):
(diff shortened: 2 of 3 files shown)


minigalaxy/installer.py
```diff
@@ -303,6 +303,7 @@ def create_applications_file(game, override=False):
 
         with open(path_to_shortcut, 'w+') as desktop_file:
             desktop_file.writelines(textwrap.dedent(desktop_definition))
+
```
sharkwouter, line 306:
```suggestion
```
This extra newline is not needed.

minigalaxy/ui/library_entry.py
```diff
@@ -325,9 +328,30 @@ def __download_update(self) -> None:
...
+        '''game_info images dict does not contain fully valid urls.
+        The entries there appear to start with //'''
+        icon_url = re.sub('.*?//', 'https://', icon)
+        download = Download(url=icon_url, save_location=local_name)
+        self.download_manager.download(download)
+
     def __check_for_update_dlc(self):
         if self.game.is_installed() and self.game.id and not self.offline:
             game_info = self.api.get_info(self.game)
+            self.__download_icon(False, game_info)
```
sharkwouter, line 354:
```suggestion
            self.__download_icon(game_info=game_info)
```

sharkwouter (changes requested):
This looks fine to me, I just had 2 style nitpicks. Thanks for the contribution!

GB609:
> This looks fine to me, I just had 2 style nitpicks. Thanks for the contribution!

There's one more thing i'd possible like to add: Check if config.create_applications_file actually is enabled. no need to download stuff when not enabled

sharkwouter:
Oh that's a good one, I had not thought of that.

GB609:
There, rebased, enhanced and comments included.

minigalaxy/ui/library_entry.py
```diff
@@ -325,9 +328,33 @@ def __download_update(self) -> None:
...
+
+        if not game_info:
+            game_info = self.api.get_info()
+        icon = game_info.get('images', {}).get('icon', None)
+        if not icon:
+            return
+
+        '''game_info images dict does not contain fully valid urls.
+        The entries there appear to start with //'''
+        icon_url = re.sub('.*?//', 'https://', icon)
```
sharkwouter, line 350: Maybe this should be `"https:" + icon`

sharkwouter: We could also replace `$//` instead.

GB609: The value in that data field currently actually is: `//path/icon`,
without protocol. But i wanted to guard against it being changed later on.
So there is no hard coded substring stuff and string concatenation. i'm ignoring and re-writing the protocol part

sharkwouter: Then place use `count=1` as argument for `re.sub`. This matches any double slash, which can be in a url without it causing issues. For example https://github.com////////////sharkwouter//minigalaxy is a perfectly valid url.

GB609: you are right. i've been a bit too careless with that replacer.
I've updated it with your suggestion.

sharkwouter (approved):

PR #655, opened 2025-02-22
Enhance download manager (required for download ui)

GB609:
## Description

I'm afraid that this is a bigger pull request than my previous ones. But i've figured that it wouldn't make much sense to split this up any further.

This PR contains changes i've implemented during my draft of the new download ui. 
It consists of the following parts:
1. a small set of improvements / bugfixes in the original download manager code i found while implementing new features
2. i've tried to encapsulate access to several internal state objects to helper methods to abstract away the logic itself from code required for correct data maintenance. This mostly concerns `__active_downloads` and `__cancel`. I've done this to avoid errors and code duplication when handling the various state changes
3. introduction of a small lifecycle for downloads:
   `QUEUED -> STARTED -> PROGRESS*{3,100} -> [COMPLETED | FAILED | CANCELED | STOPPED]; PAUSED`. That is the expected flow of a regular download without intervention. However, basically any transition is possible, with a little exception: going from pause to started is guarded by one additional none-default parameter to `download`
4. Some helper methods to encapsulate checking for states and correct bookkeeping of meta data
5. I've changed bigger portions of `__download_file` and `__download_operation` to leverage the new states and helpers, also removed some unnecessary branching
6. Listeners: This is where it all started. I need a way to get information about downloads for my download manager UI. For that i've implemented a listener pattern, as is explained in #652.
The listener/callback notification runs in a dedicated thread (using a ThreadPoolExecutor to avoid frequent construction and destruction of thread objects). And while i was doing that, i've changed how callbacks in Download objects themselves are called to also run over the listener thread and be called from the same method (if the state to notify about is supported by callbacks)

I spent half the day to try and split all those changes into several commits to make the review more manageable.
I'm going to have to rebase my ui branch onto this one, as i also changed the state names and i want to remove the changes to download_manager from the ui branch.

## Checklist
 
### DIFF
--- a/minigalaxy/download_manager.py
+++ b/minigalaxy/download_manager.py
@@ -20,10 +20,12 @@ import os
 import shutil
 import time
 import threading
+import traceback
 import queue
 
 import minigalaxy.logger  # noqa: F401
 
+from concurrent.futures.thread import ThreadPoolExecutor
 from enum import Enum
 from minigalaxy.config import Config
 from minigalaxy.constants import DOWNLOAD_CHUNK_SIZE, MINIMUM_RESUME_SIZE, GAME_DOWNLOAD_THREADS, UI_DOWNLOAD_THREADS
@@ -79,6 +81,17 @@ class DownloadManager:
     First, you need to create a Download object, then pass the Download object to the
     DownloadManager download or download_now method to download it.
     """
+
+    CANCEL_STATES = [DownloadState.PAUSED, DownloadState.STOPPED, DownloadState.FAILED, DownloadState.CANCELED]
+
+    STATE_DOWNLOAD_CALLBACKS = {
+        DownloadState.COMPLETED: lambda d, *a: d.finish(),
+        DownloadState.CANCELED: lambda d, *a: d.cancel(),
+        DownloadState.FAILED: lambda d, *a: d.cancel(),
+        DownloadState.STOPPED: lambda d, *a: d.cancel(),
+        DownloadState.PROGRESS: lambda d, percent: d.set_progress(percent)
+    }
+
     def __init__(self, session: Session, config: Config):
         """
         Create a new DownloadManager Object
@@ -104,7 +117,12 @@ class DownloadManager:
         # The list of currently active downloads
         # These are items not in the queue, but currently being downloaded
         self.active_downloads = {}
-        self.active_downloads_lock = threading.Lock()
+        self.active_downloads_lock = threading.RLock()
+
+        # members related to download change listener
+        self.download_list_change_listener = []
+        self.listener_thread = ThreadPoolExecutor(max_workers=1, thread_name_prefix="download_listener")
+        self.fork_listener = True
 
         self.logger = logging.getLogger("minigalaxy.download_manager.DownloadManager")
 
@@ -114,6 +132,48 @@ class DownloadManager:
                 download_thread.daemon = True
                 download_thread.start()
                 self.workers.append(download_thread)
+
+    def add_active_downloads_listener(self, listener):
+        if listener not in self.download_list_change_listener:
+            self.download_list_change_listener.append(listener)
+            self.listener_thread.submit(self.__init_new_listener, listener)
+
+    def remove_active_downloads_listener(self, listener):
+        if listener in self.download_list_change_listener:
+            self.download_list_change_listener.remove(listener)
+
+    def __init_new_listener(self, listener):
+        with self.active_downloads_lock:
+            for download in self.active_downloads:
+                self.__call_listener_failsafe(listener, DownloadState.STARTED, download)
+
+            for download_queue, limit in self.queues:  # noqa: F401
+                while not download_queue.empty():
+                    queued_download = download_queue.get()
+                    self.__call_listener_failsafe(listener, DownloadState.QUEUED, queued_download.item)
+
+    def __notify_listeners(self, change: DownloadState, download, download_params=[], forked=None):
+        '''helper function to notify listeners of changes to the active download list
+        Will be used for each atomic add/remove action in the list of active downloads'''
+        if not change:
+            return
+
+        # must differentiate from False to get default
+        if forked is None:
+            forked = not self.fork_listener
+
+        if forked:
+            for listener in self.download_list_change_listener:
+                self.__call_listener_failsafe(listener, change, download)
+            self.__download_callback(download, change, *download_params, forked=forked)
+        else:
+            self.listener_thread.submit(self.__notify_listeners, change, download, download_params=download_params, forked=True)
+
+    def __call_listener_failsafe(self, listener, *parameters):
+        try:
+            listener(*parameters)
+        except BaseException as exception:
+            self.logger.error(f"Error while trying to notify listener: {traceback.format_exception(exception)}")
 
     def download(self, download, restart_paused=False):
         """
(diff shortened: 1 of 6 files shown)


GB609:
It seems there's still a few bug in that related to event notifications. I'm debugging them right now

GB609:
I've got one very strange thing happening right now. It seems the various file size don't align.
```json
"downloads": {
    "installers": [
      {
        "id": "installer_windows_en",
        "name": "Genesis Alpha One Deluxe Edition",
        "os": "windows",
        "language": "en",
        "language_full": "English",
        "version": "147.8763",
        "total_size": 6165626880,
        "files": [
          {
            "id": "en1installer0",
            "size": 1048576,
            "downlink": "https:\/\/api.gog.com\/products\/1541701876\/downlink\/installer\/en1installer0"
          },
          {
            "id": "en1installer1",
            "size": 4293918720,
            "downlink": "https:\/\/api.gog.com\/products\/1541701876\/downlink\/installer\/en1installer1"
          },
          {
            "id": "en1installer2",
            "size": 1870659584,
            "downlink": "https:\/\/api.gog.com\/products\/1541701876\/downlink\/installer\/en1installer2"
          }
        ]
      }
    ]
```

The last one is bin-2. My logs for progress calculation state:
```
2025-02-23 09:46:10,422 - minigalaxy.download_manager.DownloadManager - DEBUG - setup_genesis_147.8763_(42309)-2.bin: 1300234240 / 1237759380 * 100 = 105 (previous=104)
```

The calculation uses `1237759380` as file size, which is the response from `request.headers.get('content-length').

Interestingly enough the downloaded data appears to be bigger than this (the stream hasn't ended).

I'm not sure if that is an issue in my code. It rather looks like GOG messed something up. 
Or there is a mismatch in encoding involved somehow.

GB609:
Update about the download size:
I found it. This a pretty nasty bug, although logical

```python
resume_header = {'Range': 'bytes={}-'.format(start_point)}
        download_request = self.session.get(download.url, headers=resume_header, stream=True, timeout=30)
        # ...

        downloaded_size = start_point
        self.__notify_listeners(DownloadState.PROGRESS, download, download_params=[0])
        file_size = None
        try:
            file_size = int(download_request.headers.get('content-length'))
        # ...
```
When starting at a position > 0, this part is NOT included in the response stream.
Thus, content-length will be smaller by exactly that amount!
Just taking the content-length header alone as file size won't be correct in that case, it would have to be start_point + file_size

GB609:
Another comment, this time it's more related to the PR itself than the 2 above.
I'm unsure about 2 of the new states i'm introducing.
To be more precise, i am not sure which 'name' makes more sense for which of the effects: PAUSE and STOP.

Both have in common that they 'cancel' an active download and keep any partial file (this is the main difference to CANCEL which should also remove any partial file).
What sets them apart is that one of them shall 'persist' even after restart and thus prevent automatic resume - the download MUST manually be continued.
The other one is just temporary. It can be continued manually or by restarting.

At the moment i set `PAUSE` to stick across restarts and `STOP` to be temporary.
Does that make sense, or would it feel more logical/natural the other way around?

sharkwouter:
I don't know if it makes sense to keep a distinction between stopped and cancelled. I think it is a bit confusing. Maybe you could only keep 1 of them and let the behaviour depend on if the "keep installers" option is enabled. We could also consider a separate option for keeping the cache for cancelled downloads.

GB609:
Thanks for the feedback.

I still have to rebase onto master and solve some conflicts.

The idea with `keep installers` (and `keep cache`) is interesting but i'm not sure if it is the right thing to do. I was thinking about another change here, however: if 'keep' is active, why would there be a need to first download the installer to `.cache/minigalaxy`, then move it over to `~/GOG Games/installers` _after_ successful installation. This just seems to be duplicate file IO and produces a lot of stress for the disk.

If i don't differentiate between stop and pause and link that to 'keep installers' instead, it would not be possible to temporarily halt a download (until restart) WITHOUT deleting partial success when 'keep installers' is off unless we really add a `keep cache` option. However, when i'm looking at the issues related to downloads, users expect a more clever behaviour there. Exposing a `keep partial cache` option to users appears kinda pointless to me. Why would anyone ever disable that? As long as the downloaded part is not corrupted, why redownload it again?

The differentiation is mostly used in the code to decide what to do with a download object and its data. Housekeeping basically.
I've set up the UI to only show 'pause' and 'stop' for anything active or queued. 'cancel' is a state thats linked to the 'trash-bin' or 'delete' icon in the list of (unsuccessfully) finished downloads, so i personally think it should be pretty clear.

Moreover, this would formalize (and featurize) something which currently only happens implicitely.
Shutting down minigalaxy has exactly this effect right now, but very undeterministic, as far as i can tell.
Because it just goes down, while the daemonized download threads continue until they feel like they should stop.
Adding in STOP separate from a full 'abort/cancel' allows to really put a graceful shutdown into code without relying on daemon threads.

sharkwouter:
That is a good point. Okay, that makes sense. I guess linking to keep installers is not really a good idea. I'll let you do your thing.

GB609:
There is one more thing i thought about:
The constant `DOWNLOAD_CHUNK_SIZE`. It's set to 1MB and the amount of time download threads eat up in general.

As far as i've understood the documentation about multithreading in python, there is almost no true parallelism of python code itself because of the GIL (unless the code somehow specially overrules standard locking behavior).
What can run in parallel is IO, thus streaming download data from HTTP or the resulting chunk into the target file.
We might achieve a better parallel throughput in downloads when we increase the chunk size, to reduce the amount of times that a download thread returns to python.
A simple raise to 2MB shouldn't be too much, but it would decrease the amount of times that a download thread locks the GIL by half. 

Plus, there are 4 threads. And they poll their queues with a sleep loop of 0.01 when not actively downloading something.
The download threads don't need to be real time, i think even putting a sleep of 0.5 would be plenty.

sharkwouter:
It depends on what internet speed people have. 2 MB can be a lot on slow connection. Python will stretch the 0.01 to end when it has time. During sleep Python will do other things. I'm not sure if we should change this, we'll have to test this.

GB609:
> It depends on what internet speed people have. 2 MB can be a lot on slow connection. Python will stretch the 0.01 to end when it has time. During sleep Python will do other things. I'm not sure if we should change this, we'll have to test this.

Yeah, this needs some performance measurements.
There are 8 worker threads in downloadmanager. Each wakes up every 0.01 seconds, trying to acquire gil to poll if there is an item on the download queue, then goes back to sleep for exactly that time. That is my entire point here. The downloads threads try to interrupt 'other things' too often and sleep to little, taking resources from the ui. At least thats my gut feeling. Definitely needs some data crunching and profiling.

Changing the chunk size shouldnt change physical download speed, it controls how much is fetched before writing to disk again. I can imagine that this would influence how much progress is lost on shutdown in the middle.
But apart from that it should rather be an issue with ram.

Just thinking about it. Nothing we need to do.

GB609:
Alright. I think that's all for the basics.
There are a few intermediate commits where the file_size issue i fixed for 1.3.2 before is gone again (or not fully functional). That happened because i didnt want to re-organize all commits and re-solve the same conflict every time i touched the same line. I've placed it as the last commit on top again.

sharkwouter:
Thanks for working on this. I hope to be able to review this today or tomorrow. Sorry for the delay, last days have been a bit crazy.

GB609:
> Thanks for working on this. I hope to be able to review this today or tomorrow. Sorry for the delay, last days have been a bit crazy.

No hurry. Take your time. It's quite a lot to review all at once. Shall i try to split it up into smaller PRs?

sharkwouter:
It's okay, I don't think it makes sense to split this one up. That would just add extra work.

minigalaxy/download_manager.py
```diff
@@ -159,32 +262,32 @@ def cancel_download(self, downloads):
...
         """
+
+        if not download_dict:
+            download_dict = self.active_downloads
+
         with self.active_downloads_lock:
             for download in self.active_downloads:
                 if download in download_dict:
-                    self.logger.debug("Found download")
+                    self.logger.debug("Canceling download: " + download.save_location)
```
sharkwouter, line 286:
```suggestion
                    self.logger.debug("Canceling download: %s", download.save_location)
```

sharkwouter (commented):
The code looks pretty good and it works well, I just had one very minor nitpick.

sharkwouter (approved):

GB609:
I've included your change.
So, do you think it's ready to merge? Anything else missing?

sharkwouter:
You can merge it whenever. This seems good to me.

Issue #630, opened 2024-12-14
"BIN file was corrupted, please download it again" should smartly resume and perhaps not reload entire game


ell1e:
Thanks to the awesome #615 I got to try to download a few multi part games again, but I ran into another curious corner case: I got a "BIN file was corrupted, please download it again" error for Fallout 3 GOTY. This seems a bit strange since even with an extremely shaky connection, TLS should usually ensure integrity. But assuming that this really was caused by a connection hiccup anyway, rather than some bug in the downloader, e.g. the downloader not transparently doing a HTTPS resume when the download fell short, there seem to be two things that are a bit odd about what happens after:

1. The message appearing in the first place. Usually, for a smart downloader e.g. like Steam, the established behavior would be to simply quietly resume redownloading the affected part. The user interaction doesn't really seem to be required here.

2. The way the recovery is done when redownloading. When I prompt it to restart, minigalaxy seems to then proceed to start over the entire game from scratch, rather than just the affected BIN file, at least from how the progress bar looks like. Especially for bigger games, which are now not uncommon on GOG, this seems like it doesn't really scale well if that's how errors are handled.

**Affected minigalaxy version:** 1.3.1 (via Flathub)

GB609:
The progress bar is a matter on its own.
But i just found a pretty serious bug in the download_manager related to file resume while reworking the code.
See https://github.com/sharkwouter/minigalaxy/pull/655#issuecomment-2676735067

The way the download_manager checks if a download stream has to be consumed effectively checks:
`downloaded_part < request.header('content-length')`

When resuming a partial download from disc, the to download will include a startpoint.
Naturally this means that the response will NOT include any bytes up until the start point, so the returned content length will also be smaller by that amount.
This ultimately means that any download that has progressed up to exactly half will not be finished when restarting minigalaxy in-between. The response stream is just never consumed in that case.

ell1e:
There seems to be also at least one or possibly more hidden conditions where the resume fails, rather than completing as corrupted.

GB609:
Maybe. But it's hard to debug if there's multiple reasons at once.

GB609:
> 
>     1. The message appearing in the first place. Usually, for a smart downloader e.g. like Steam, the established behavior would be to simply quietly resume redownloading the affected part. The user interaction doesn't really seem to be required here.
> 
> **Affected minigalaxy version:** 1.3.1 (via Flathub)

The problem here is the point in time when corruption is checked currently.
[...]

PR #656, opened 2025-02-23
Bugfix/630

GB609:
## Description

See #630 
and https://github.com/sharkwouter/minigalaxy/pull/655#issuecomment-2676735067

## Checklist
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Desktop shortcuts created by Minigalaxy will now be updated with environment and launch argument changes from the game's property dialog when OK is clicked (thanks to GB609)
 - Clean-up and fixes of the code used for grid and list style library views (thanks to GB609)
 - Desktop shortcuts for windows games can now have icons, if gog provides them in the game info api 'images' section (thanks to GB609)
+- Fixed a bug in download_manager that prevents correct resume of downloads with 50%+ progress after restart (thanks to GB609)
 
 **1.3.1**
 - Fix Windows games with multiple parts not installing with wine
--- a/minigalaxy/download_manager.py
+++ b/minigalaxy/download_manager.py
@@ -363,6 +363,10 @@ class DownloadManager:
             file_size = int(download_request.headers.get('content-length'))
         except (ValueError, TypeError):
             self.logger.error(f"Couldn't get file size for {download.save_location}. No progress will be shown.")
+        if file_size and downloaded_size > 0:
+            # we are resuming a partial file. file_size from content-length
+            # will not include what we requested to skip over
+            file_size += downloaded_size
         result = True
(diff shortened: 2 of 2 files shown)


sharkwouter (approved):

sharkwouter:
Wow, thanks for finally finding this! This has been an issue for so long.

GB609:
Sorry about posting in an already closed issue again, but i've got a general question.
Since the last release 1.3.1, we've already had a pretty big list of changes and bug fixes.
Do you think it might be useful to think about releasing a 1.3.2 version? 
Maybe sort out a milestone for things you'd absolutely want fixed first for that?

I think this would mostly help users of the flatpak version and the debian package, because everyone else could also just use the git master branch if desired.

sharkwouter:
I agree that a new release is sorely needed, but I don't know if now it the time yet with the current changes in the pipeline and what to call the version. I was thinking we might want to release a new version after the download overview and call it 2.0.0. We could also do a release now and have a nice bugfix release. What is your opinion? @GB609

GB609:
Since you've merged the last 2 bug fixes i think we have a relatively stable intermediate version.
I'll try to go over all basic install/download cases tonight to see if i've missed something.

However, i think the fix to download manager regarding the corruptions alone almost warrants a new release. Add in the changes/improvements to windows installations i started with, it would definitely be useful. In fact, i personally am not using 1.3.1 anymore since quit a long time now but run a commit which includes the wine installer fixes at least on my "productive" game rig.
Might also help with re-testing and closing all those 'game does not run/install' issues of which a large amount is likely outdated.
Although i refactored a lot of stuff already, from a users point of view it's mostly been bug fixes. So i'd go with 1.3.2.

When my tests turn out ok, i think we could go with releasing what we have right now as 1.3.2, because that is what the changelog has prepared already. If i there's some all to obvious bug, it needs to be squashed first to prevent regression (i'm not producing entirely error-free code all the time).
For the enhancements to download manager and its ui, we could either go to 1.4.0 or 2.0.0, but here i personally think the amount of changes does not yet warrant a full jump of the major version (2.0.0). But thats something we can decide or change later i guess.

sharkwouter:
> However, i think the fix to download manager regarding the corruptions by itself warrants a new release already.

Yes, I agree. Let me know if you find any more issues tonight. If you give me the go, I'll create the 1.3.2 release tomorrow.

GB609:
But if you want to wait for the download ui, that's also fine by me. It _is_ a pretty large block to add all at once and quite a lot to review, though.

sharkwouter:
I think it makes sense to wait and put the download ui in a major release later, so we still have time to find potential issues with it before we give it to the public.

GB609:
What about the issue in setup.py? Wouldn't this impact the release as well?

PR #661, opened 2025-02-26
minor improvements in wine install

GB609:
This is something i was thinking about and testing since i changed the wine install.
There are 2 little issues i have with how i changed prefix creation.

1. Speed up initial prefix creation: When a new prefix is created during install, it already is fully updated.
The additional command 'wineboot -u' forced running the same upgrade procedure a second time which can take a long time, depending on the hardware.
Since the main point of this command was just to run anything in wine to let it create the prefix, it can be changed to something that works faster. I personally am using a relatively old (yet still powerful) pc, so the difference is really noticable.
2. In addition to using an env variable to disable winemenubuilder, this commit tries to disable it more permanently in the registry to prevent additional installations triggered by the game itself from adding more shortcuts (some game engines install dependencies they need afterwards). It took me a while to find the right file format for the registry file to work.
Combined with 1., so the first action that is done to create the wineprefix is running a regedit import. 

## Checklist
 
 - [ ] _CHANGELOG.md_ was updated - will add it after the other PR is merged to prevent another merge conflict on this file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 **1.4.0**
-- various improvements to the download manager, including a pause function (not usable from the UI yet) (thanks to GB609)
+- Various improvements to the download manager, including a pause function (not usable from the UI yet) (thanks to GB609)
+- Speed up creation of wine prefixes during installations (thanks to GB609)
+- Use regedit to more permanently disable useless shortcut creation by wine from within a prefix (thanks to GB609)
 
 **1.3.2**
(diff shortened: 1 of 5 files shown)


sharkwouter (changes requested):
Thanks for the PR. Can you update the changelog to include the speedup? Otherwise this looks good to me.

GB609:
> Thanks for the PR. Can you update the changelog to include the speedup? Otherwise this looks good to me.

Yeah, i was planning to. Just didnt want to provoke merge conflicts with the download manager.

sharkwouter:
That makes sense

sharkwouter (approved):

Issue #667, opened 2025-03-05
When many downloads are happening, not all can be shown anymore


sharkwouter:
I noticed that it is not possible to scroll if not all downloads fit in the download list fit on the screen. We should probably fix this before we do the next release.

GB609:
Sure. Good catch. Seems like i forgot to add a scrollable parent container

GB609:
Done with #668

Issue #635, opened 2025-01-05
[Feature request] Give more control on downloads


thegreyshadow:
Thanks for Minigalaxy. It's a really good way to manage our GOG games under Linux.

Right now we can download games on Minigalaxy but the control is quite basic: after configuring a game folder, all we can do is press a "Download" button.

Now, we're talking about games that are several GB in size and this introduces some constraints on bandwith- and storage- challenged systems.

I request, if possible:
1. To be able to set up a download folder other than the games folder.
2. To be able to cancel a download, with the option to resume later or delete the partially downloaded file.
3. To get an estimate of the space needed for both the installer and the installed file.

I'd understand if this is not possible, but thanks anyway!

GB609:
It is possible with operation system means to separate installers from games.
There's 2 directories where downloads are stored:
* `~/.cache/minigalaxy/download`
* `<your/game/folder>/installer`

You could bind mount or symlink these directories to point to another drive.
I agree this is not ideal, but until there's an improvement being made, minigalaxy is usable like this even on system with little storage. I actually do the same, i've moved (and linked) the installer directory to a (labeled) usb disc.
Also, game directories should normally be self-sufficient (with the exception of windows games requiring registry keys currently). So you can basically do the same with games after installation as long as you name the link exactly like the original game directory.

Cancelling a single file download directly is not possible. But a game which is currently downloaded can be stopped, although it is rather hard to see: In grid view, there's a small x in the top right corner when the game is actively downloaded. If you click this, the entire game download is aborted. What i don't know right now is if this will keep files that have been downloaded already.
I've had a few situations where minigalaxy immediately started a download after i opened it again, so i think it is picking up and resuming downloads automatically when quitting/restarting.

I agree that this area needs some improvements in terms of ui capabilities and feedback.

sharkwouter:
To give an addition to what GB609 said, we do actually try to make sure the user has enough disk space when downloading and installing a game, but it is hard to be precise with it. The user will get an error message if there is not enough space.

ell1e:
Related problem #630 , it seems like downloads in general are fickle with larger files and often won't resume properly. With Steam I can download 100GB+ games, with minigalaxy I'm lucky if I manage a 5GB game since it might complain it is corrupted and just delete everything and start over.

PR #668, opened 2025-03-06
Bugfix/download UI

GB609:
## Description

Now it's my turn again :-)

This PR deals with 2 small issues:
- #667: I've added a scrollable container and let it calculate the maximum height to use dynamically
- #635: adding an estimated size info to the download list. I think there's not much more we can do for that feature request otherwise. the only thing left would be to change the download location on a per-game basis and i think that is not feasible.

I've used the following test code snippet in the constructor of DownloadList to generate the screenshot:
```python
        for i in range(20):
            game = Game('Name:' + str(i))
            size = 1024**(2+min(int(i / 10), 1)) + (1024**2 * i * 72)
            download = Download('None', str(i), DownloadType.GAME, game=game, expected_size=size)
            self.download_manager_listener(DownloadState.QUEUED, download)
            self.download_manager_listener(DownloadState.PROGRESS, download)
```

size_adjust

## Checklist
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Use regedit to more permanently disable useless shortcut creation by wine from within a prefix (thanks to GB609)
 - Introduce a UI to show all active, stopped or failed downloads. Also allows to permanently pause a download (thanks to GB609)
 - Make more Windows games able to launch through Minigalaxy
+- The download UI also shows a rough size estimate if possible (thanks to GB609)
 
 **1.3.2**
 - Completely reworked windows wine installation. This should solve a lot of problems with failing game installs. Innoextract (if installed) is only used to detect and configure the installation language. (thanks to GB609)
--- a/data/po/cs_CZ.po
+++ b/data/po/cs_CZ.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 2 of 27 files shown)


sharkwouter (approved):

sharkwouter:
Thanks! That is a nice improvement.

sharkwouter:
One thing I am missing is a way to make pending downloads start downloading now. The pause button could be replaced with an up arrow button which makes that happen. If it is a later part of something already downloading, we should probably not allow this, though. I was just thinking of it while looking at this.

GB609:
Yeah, i was also thinking about this at one point in time.
It's a bit tricky to do, however.
When pending downloads contains anything, that means that all worker threads are busy.
When i just stop one download, the next in the queue would be picked.

To get this to work, i'd first have to re-order the queue, then either stop one of the downloads or add one more worker.

sharkwouter:
Yeah, it will be a bit hard to implement. I think what users would expect is that the active download at the bottom would move to the top of the pending list.

sharkwouter:
The way we deal with multi-part downloads in this is what makes it hard.

sharkwouter:
I think having a download now button would be good enough, that is what Steam does as well. Adding dragging support sounds a bit overkill.

GB609:
In that case i'd probably just add another worker and leave it to the user to manually stop (or requeue) low-prio downloads afterwards.
With an additional flag for that situation we can even make it so that the previous, now stopped worker, dies and doesn't take another item from the queue to prevent the amount of workers from growing.

Issue #664, opened 2025-03-02
Downloads: disable parallel downloads unless connection is fast


ell1e:
I've noticed minigalaxy seems to always download 3+ parts at once. This isn't a good idea on a slow connection and will only make everything slower and cause more dropouts. Therefore, I suggest this should only be done if the connection speed exceeds a certain threshold. Not sure where that ideally would be, but I'm guessing something like 10mbit/s as an absolute minimum.

**Affected minigalaxy version:** 1.3.2 (latest flatpak stable)

**Affected operating system:** postmarketOS 24.12 based on Alpine 3.21.3

sharkwouter:
Thanks for this report. Maybe making the amount of download threads configurable would be a good solution for this.

ell1e:
Most users probably won't think to configure it, so the default behavior might still be relevant.

sharkwouter:
Fair point. I do think making it based on download speed would be harder to implement, but not impossible.

GB609:
I'd just give an option and introduce a way to dynamically change it in the downloads ui.
Maybe just a small numbers dropdown in the 'active downloads' section.

PR #670, opened 2025-03-07
Feature/configurable download workers

GB609:
## Description

1. Another assortment of bugs i stumbled across while working on the new feature
2. A UI option to reduce/increase the amount of parallel workers. Covers #664 

Preview of the ui:

download_workers_option

The worker number is saved in config and thus persists across restarts.
But i also implemented a method to increase/reduce the worker number on the fly. So the new setting applies immediately without restart, even when there are active downloads already.
The stopping function comes with 2 variants:
1. Do not actively stop running downloads. Worker will finish their current task and terminate if there are more workers than configured.
2. As an addition to 1, stop and requeue downloads with the least progress. This is the current default behavior.

The 2 variants are controlled by a boolean method parameter, which is hard coded at the moment. I'm not sure which of both variants is the better one and whether we should just open both variants to let the user decide, maybe through an option or a yes/no question.

~~Edit 2: I just saw that i'm missing one thing: the workers arent removed from the list of workers when stopped. Need to fix/add this.~~ - FIXED

## Checklist
 
### DIFF
--- a/minigalaxy/config.py
+++ b/minigalaxy/config.py
@@ -9,7 +9,7 @@ from minigalaxy.paths import CONFIG_FILE_PATH, DEFAULT_INSTALL_DIR
 # UI download threads are for UI assets like thumbnails or icons
 UI_DOWNLOAD_THREADS = 4
 # Game download threads are for long-running downloads like games, DLC or updates
-GAME_DOWNLOAD_THREADS = 4
+DEFAULT_DOWNLOAD_THREAD_COUNT = 4
 
 
 class Config:
@@ -168,7 +168,7 @@ class Config:
 
     @property
     def max_parallel_game_downloads(self) -> int:
-        return self.__config.get("max_download_workers", GAME_DOWNLOAD_THREADS)
+        return self.__config.get("max_download_workers", DEFAULT_DOWNLOAD_THREAD_COUNT)
 
     @max_parallel_game_downloads.setter
     def max_parallel_game_downloads(self, new_value: int) -> None:
--- a/minigalaxy/config.py
+++ b/minigalaxy/config.py
@@ -5,6 +5,12 @@ from typing import List
 from minigalaxy.logger import logger
 from minigalaxy.paths import CONFIG_FILE_PATH, DEFAULT_INSTALL_DIR
 
+# Moved from constants.py to here because of circular import between translations, config and constants
+# UI download threads are for UI assets like thumbnails or icons
+UI_DOWNLOAD_THREADS = 4
+# Game download threads are for long-running downloads like games, DLC or updates
+GAME_DOWNLOAD_THREADS = 4
+
 
 class Config:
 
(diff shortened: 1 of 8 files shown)


GB609:
> Thanks for the PR. I hope I'll be able to look at this soon. It probably won't be today, though. I've been a bit swamped with work and moving lately.

Take your time.

sharkwouter:
I did a quick test and I really like this change. I do think it would be nicer if the change would be instant when lowering the amount of threads, but that could be a separate PR. I'll go through the code, see if I see any issues.

minigalaxy/config.py
```diff
@@ -5,6 +5,12 @@
 from minigalaxy.logger import logger
 from minigalaxy.paths import CONFIG_FILE_PATH, DEFAULT_INSTALL_DIR
 
+# Moved from constants.py to here because of circular import between translations, config and constants
+# UI download threads are for UI assets like thumbnails or icons
+UI_DOWNLOAD_THREADS = 4
+# Game download threads are for long-running downloads like games, DLC or updates
+GAME_DOWNLOAD_THREADS = 4
```
sharkwouter, line 12: You might want to rename this one. I also don't know if we should have these separate from regular downloads, they kind of are the same thing. That last part should not be addressed in this PR I think.

GB609:
> I did a quick test and I really like this change. I do think it would be nicer if the change would be instant when lowering the amount of threads, but that could be a separate PR. I'll go through the code, see if I see any issues.

The change _is_ instant currently.
But i've prepared the option for both.

minigalaxy/download_manager.py
```diff
@@ -568,7 +623,7 @@ def __is_same_download_as_before(self, download):
         # Check first part of the file
         resume_header = {'Range': 'bytes=0-{}'.format(size_to_check - 1)}  # range header is index-0-based
         with self.session.get(download.url, headers=resume_header, stream=True, timeout=30) as download_request:
-            if not download_request.status_code == codes.ok:
+            if not 200 <= download_request.status_code < 300:
```
sharkwouter, line 626: This can be `if not download_request.ok:`: https://docs.python-requests.org/en/latest/api/#requests.Response.ok

sharkwouter (changes requested):
Looks good, but I did have 2 minor things. Thanks for the work on this and sorry for taking a while to get back to you.

minigalaxy/config.py
GB609: 4 threads for ui alone always seemed kind of like overkill to me. But i think its a good idea to keep one or 2 threads idle for the ui. Otherwise opening game info will not show images while games are downloaded and all workers are busy.

We could rename it to something like `DEFAULT_POOL_SIZE` or `DEFAULT_NUM_WORKERS`

sharkwouter: Yeah, we could lower the ui threads to 2, but lets not do that in this PR. I think changing `GAME_DOWNLOAD_THREADS` to `DEFAULT_DOWNLOAD_THREAD_COUNT`  could work. I think we should probably load it from the config instead, though.

GB609: We do. 
Download mananger uses the new property config.max_parallel_game_downloads.
The constant is only there so that this default can be changed quickly without having to search the property in the code.

sharkwouter (approved):

Issue #675, opened 2025-03-28
The DLC list seems to stop working once there are too many entries


ell1e:
It seems like the DLC list will no longer work once there are too many entries:

![Image](https://github.com/user-attachments/assets/37377fd2-1402-4e26-be24-3e7062ad7609)

Or perhaps I missed some way to navigate this, but it seems like many of the entries aren't reachable and it's therefore not possible to access them.

**Affected minigalaxy version:** 1.3.2 (flathub)

sharkwouter:
Oh, good catch. I had not considered this could happen when this was implemented.

GB609:
I think thats's a similar issue to what we had with the new download ui list before, there is no scrollable parent element.
That is an absurdly long list, though.

ell1e:
I swear I just bought one addon, but apparently the game dev thought it had to be a couple dozen separate items in practice. 😆

GB609:
I'm not blaming you.

I think the issue here is that the meaning of 'DLC' has shifted from the 'traditional' meaning of '(relatively) large game add-ons' to also include literally every little microtransaction separately.

Maybe we should rework or redesign the DLC-UI in its entirety on the long run.

sharkwouter:
In the long run, it should probably be made better. For now we can just make it able to scroll.

ell1e:
I noticed that all or most of the DLCs seem to cause separate game entries after install as well. Is that intended? I don't think these are items that can actually be launched separately:

![Image](https://github.com/user-attachments/assets/0ff60f79-51c7-42ad-a4fa-a53560da16fe)

GB609:
Seems like the DLC installation is broken here. That likely happens because the DLC install generates the wrong target directory for the installation. 
Might also be related to the GOG product info about the DLCs.

I've touched the DLC UI code during some of my reworks (but tested it for installation with some linux games). I'm gonna have a look at that tonight, this seems to be rather serious.

ell1e:
**Update:** It seems like I also can't launch the game via minigalaxy anymore, I get this error:

![Image](https://github.com/user-attachments/assets/6d695c60-3212-4e65-824f-67c3e87e97a9)

I can still launch it if I open up a terminal and start wine from there. It seems like the DLC being installed cause the code that finds out what executable to launch to be confused. 🤔  Fascinating!

ell1e:
If you want me to gift this game to you for testing, let me know.

GB609:
[...]

Issue #619, opened 2024-11-26
Launching Unreal Gold leads to minigalaxy error with backtrace


ell1e:
Trying to launch Unreal Gold leads to minigalaxy error with backtrace:
```
~ $ flatpak run io.github.sharkwouter.Minigalaxy
Traceback (most recent call last):
  File "/app/lib/python3.11/site-packages/minigalaxy/ui/gametile.py", line 121, in on_button_click
    err_msg = start_game(self.game)
              ^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/minigalaxy/launcher.py", line 48, in start_game
    error_message, process = run_game_subprocess(game)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/minigalaxy/launcher.py", line 205, in run_game_subprocess
    get_execute_command(game),
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/minigalaxy/launcher.py", line 65, in get_execute_command
    exe_cmd = get_windows_exe_cmd(game, files)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/minigalaxy/launcher.py", line 142, in get_windows_exe_cmd
    filename = os.path.splitext(os.path.basename(executables[0]))[0] + '.exe'
                                                 ~~~~~~~~~~~^^^
IndexError: list index out of range
```
It seems like the install has worked fine though, so I think the wine prefix setup and everything works and only launching it once it's ready to go doesn't work.

**Affected minigalaxy version:** 1.3.0 (from flathub)

sharkwouter:
Interesting, my guess is that there is no binary in the root directory of Unreal Gold. It is probably in a bin directory. I do have this game, maybe I can figure this one out.

sharkwouter:
Turns out I was right, Unreal Gold has no exe files in the main directory and no config file which shows where to find the executable.

ell1e:
minigalaxy 1.3.2 seems to try to launch some random foxit reader install inside the Unreal Gold folder now. 😆 It doesn't seem to guess the game executable correctly.

GB609:
Have you tried a full delete + uninstall?

Could you provide an ls of the game dir?
And also the content of the file goggame*.info?

sharkwouter:
[...]

PR #676, opened 2025-03-29
DLC handling fixes, improve logic searching for executable command

### LINKED ISSUE #674: Workers  Resources Soviet Republic starts modelviewer
I have deleted my installation and re-installed Workers  Resources Soviet Republic through minigalaxy. In the hopes that it would resolve the issue I'm facing. Minigalaxy launches the wrong program and I don't understand why.  The file opened with Minigalaxy is "Modelviewer.exe", but that is not in the goggame.info file:
```
{
    "languages": [
        "en-US"
    ],
    "version": 1,
    "name": "Workers \u0026 Resources: Soviet Republic",
    "language": "English",
    "gameId": "1860281269",
    "rootGameId": "1860281269",
    "playTasks": [
        {
            "languages": [
                "en-US"
            ],
            "name": "Workers \u0026 Resources: Soviet Republic",
            "type": "FileTask",
            "isPrimary": true,
            "path": "SETUPAPPLICATION SOVIET_g.exe",
            "category": "launcher",
            "url": ""
        },
        {
            "languages": [
                "en-US"
            ],
            "name": "Workers \u0026 Resources: Soviet Republic - launcher process SOVIET64_g_exe",
            "type": "FileTask",
            "path": "SOVIET64_g.exe",
            "isHidden": true,
            "category": "game",
            "url": ""
        },
        {
            "languages": [
                "en-US"
            ],
            "name": "Workers \u0026 Resources: Soviet Republic - launcher process SOVIET_g_exe",
            "type": "FileTask",
            "path": "SOVIET_g.exe",
            "isHidden": true,
            "category": "game",
            "url": ""
        }
    ],
    "buildId": "58215007719320020"
}
```
The .lnk file points to SETUPAPPLICATION SOVIET_g.exe, but that is not started.
Where can I find the reason why minigalaxy does not start the correct application, either the launcher or the SOVIET64_g.exe?

GB609:
1. Only recognize files with none-empty `playTasks` for finding installed games

2. launcher.py:get_windows_exe_cmd: Don't iterate over all `goggame*.info` files in a game directory to search the launch command, only use the one holding the current game's id.  
This becomes important when multiple files contain playTasks, e.g. when there is something like a stand-alone extension for a game that requires another executable. The extension's game id could be sorted alphabetically after the base game. In that case, launching the stand-alone extension would still pull out the executable from the base game's `goggame*.info` file, which would be wrong.  
Additional fallback: If there is a goggame.info file, but it's not valid, then first check if there is an `*.lnk` file. This is only a slight addition to the previous logic which used  
`if goggame_file: ... elif link_file else find_first_exec` where an invalid, but existing goggame.info file would prevent the fallback logic from being used and result in returning just `env WINEPREFIX="prefix-path"`
Generally simplified the if logic by using `dict.get('key', defaultValue)` patterns for access to potentially empty json values.

3. launcher.py:get_windows_exe_cmd_from_goggame_info: Simplified branch nesting and ifs. Removed unnecessary check for `category=game` - we should also use `category=launcher` if it is marked as `isPrimary=true`

4. small bugfix in LibraryEntry.__download_dlc: the inner function finish_func must be defined after dlc_title variable or it won't be visible/set.

Fixes #674
Fixes for #675: The part were the DLCs are listed as their own games. And DLC list can now scroll and change layout dynamically as required.
Fixes some of the games mentioned in #619

**Edit:**
I've added some code to handle large dlc lists as well. The 'list' is now a FlowBox. I'm dynamically adjusting the size and max items per column and have it scroll horizontally.
The size calculations are not perfect, there are some border cases with **very** small windows and large DLC lists where a part of the DLC Popover is cropped. After a few hours of trial and error and different, much more complicated size and position calculations, i figured it's just not worth it.

I'm going to add a few screenshots how that effectively looks:
dlc_list_default
dlc_list_1
dlc_list_2
dlc_list_3

 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 - The download UI also shows a rough size estimate if possible (thanks to GB609)
 - fix connection resource leak in DownloadManager (thanks to GB609)
 - The number of parallel downloads can now be adjusted dynamically. Previous setting is saved. (thanks to GB609)
+- Fixed a bug that would lead to DLCs being recognized as stand-alone games (thanks to GB609)
+- Allow more games to be launched via information from goggame.info (thanks to GB609)
 
 **1.3.2**
 - Completely reworked windows wine installation. This should solve a lot of problems with failing game installs. Innoextract (if installed) is only used to detect and configure the installation language. (thanks to GB609)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - The number of parallel downloads can now be adjusted dynamically. Previous setting is saved. (thanks to GB609)
 - Fixed a bug that would lead to DLCs being recognized as stand-alone games (thanks to GB609)
 - Allow more games to be launched via information from goggame.info (thanks to GB609)
+- The DLC list will now show a table-like multi-column view where necessary. This fixes issues with games having a high number of DLCs. (thanks to GB609)
 
 **1.3.2**
 - Completely reworked windows wine installation. This should solve a lot of problems with failing game installs. Innoextract (if installed) is only used to detect and configure the installation language. (thanks to GB609)
--- a/data/ui/gametile.ui
+++ b/data/ui/gametile.ui
@@ -1,7 +1,126 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.38.2 -->
+<!-- Generated with glade 3.40.0 -->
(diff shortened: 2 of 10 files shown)


sharkwouter (approved):
Looks great! Thanks for making this change and sorry for getting back to you late. I have now finally set up my computer again. I'll have proper internet again somewhere this week.

GB609:
Don't worry, it's fine.
Thanks for the approval.

PR #687, opened 2025-05-09
Bugfix: Maintaining downloaded installers (cache and keep_installers option)

### LINKED ISSUE #513: Erases all of the installers when downloading new ones.
When Mini begins to download an update or a DLC it deletes the existing installers that are related to the game.

### LINKED ISSUE #545: (BUG) it cannot download the game
Hi
I tried to download the game and it does not worked
image

### LINKED ISSUE #579: DLC Queuing Produces Install Error
When I clicked multiple DLCs to install, *all* will fail with errors either immediately or after the download completes. In addition the download bar will jump rapidly around.

There are two errors that pop up: "Failed to install" DLC and DLC "Corrupted"

Screenshot from 2023-08-26 09-44-53

Original DLC queuing PR: https://github.com/sharkwouter/minigalaxy/pull/395

### LINKED ISSUE #677: After installing a game and some DLC while "Keep installers" is enabled, uninstalling the base game and reinstalling may launch wrong installer
It seems like after installing a game and some DLC while "Keep installers" is enabled, uninstalling the base game and reinstalling may launch wrong installer.

Steps to reproduce: 1. I installed "Potion Permit" and a bunch of the DLCs, "Keep installers" was enabled before I began downloading it. 2. But then uninstalled it due to likely some DLC install issue, since I tried to install multiple at once and apparently that currently can lead to corruption problems. 3. I quit minigalaxy after uninstalling the game. 4. I restarted minigalaxy, and 5. I clicked "Install". Instead of the base game installer however, I now get a DLC installer that complains that the base game can't be detected.

**Affected minigalaxy version:** 1.3.2 (flatpak)

### LINKED ISSUE #679: Crusader Kings Complete installs to folder "~/GOG Games/Crusader Kings Complete /" but game is launched via "~/GOG Games/Crusader Kings Complete/"
The game [Crusader Kings Complete](https://www.gog.com/en/game/crusader_kings_complete) installs to folder `~/GOG Games/Crusader Kings Complete /` with a trailing space, but game is launched via `~/GOG Games/Crusader Kings Complete/` without a trailing space, causing this error:

Image

```
$ cat goggame-1423233026.info

{
	"version" : 1,
	"gameId" : "1423233026",
	"rootGameId" : "1423233026",
	"standalone" : true,
	"dependencyGameId" : "",
	"language"         : "english",
	"name"             : "Crusader Kings Complete",
	"playTasks"        : [
		{
			"isPrimary" : true,
			"type"      : "FileTask",
			"path"      : "Crusaders.exe",
			"workingDir" : ""
		}
	],
	"supportTasks"     : [
		{
			"name" : "Support",
			"type" : "URLTask",
			"link" : "http://www.gog.com/en/support/crusader_kings_complete"
		},
		{
			"name" : "Manual",
			"type" : "FileTask",
			"path" : "Manual.pdf",
			"workingDir" : ""
		},
		{
			"name" : "Readme",
			"type" : "FileTask",
			"path" : "Readme.txt",
			"workingDir" : ""
		}
	]
```

Perhaps this is an error on GOG's end with the data, but I thought in case it wasn't, I'd bring it up here first.

**Affected minigalaxy version:** 1.3.2 (flatpak)

GB609:
This PR fixes a list of long-standing bugs related to the keep_installers option, but also related to the issues with parallel DLC installations interfering with each other by file deletions.

This might not appear to be related to my current main work - the download issues - at first glance, yet it is very much related. I can't really fully fix parallel dlc downloads when the installations keep interfering with each other. There's no point if the downloads succeed and can be done in parallel but they can't be applied because of misbehaving installs.

## General assumption/approach of the new behaviour:
* Downloads will go to the final target directory directly - no post-install moving from cache anymore.
* A fixed directory structure. Basically everything that has its own id on gog gets its own directory, dlcs and base game files will not be in the same directory anymore, not even in cache. Dlcs installers will go into subdirs of the base game installers dir 
  * Base games: `[download_target_dir]/[game_name]/[installer_name][.extension=(.exe, .sh, -n.bin...)]`
  * DLCs: `[download_target_dir]/[game_name]/[dlc_name]/[installer_name][.extension=(.exe, .sh, -n.bin...)]`
  * `[download_target_dir]`: either `config.install_dir/installer` or `.cache/minigalaxy/download`
  * `[game_name]` and `[dlc_name]`: as returned from gog api, stripped of problematic characters. While i added this, i also fixed #679
  * `[installer_name]`: currently seems to be something like `setup_gamename_version.extension`. As long as that is given, it's theoretically possibly to hold several versions in parallel in the same directory (which isn't supported yet). When there are name collisions because the names are generic, new version downloads will overwrite old, leading to less deletions of old files afterwards (basically only when the previous version had more files than the more recent version)
 * A game download consisting of multiple files will place all of its files in the same directory, not litter them across multiple dirs.

## With the assumptions above, the following was implemented:
1. Game downloads go to real target directly: when `keep_installers=True`, `LibraryEntry.__download` will set up save_locations to be under `config.install_dir/installers`, according to the fixed directory structure described above. This has the added benefit, that already downloaded files do not have to be moved around after installation which saves a lot of IO, especially for large games. 
2. Improved directory structure: When downloading a DLC, the save_location will additionally be placed under a directory named after the DLC. This prevents deletion of too many files when a dlc is downloaded and installed.
3. All download finish callbacks get a list of file name, which is passed to install_game
4. `installer.py:remove_installer` builds a list of files to keep depending on the value of keep_installers and the passed in file list:
 * if keep_installers=False, `keep_list=[]` - all direct child files (**not** directories) in the current directory are removed.
 * if keep_installers=True and `keep_list != None`, use that list and delete other not in that list
 * if keep_installers=True and `keep_list == None or empty`, use `installer_name` without extension as filter for files to keep. This case is necessary when the install is started from an already downloaded file.
 * The clean up will not recurse down into subfolders because subfolders would be dlcs. And only files, not directories, will be deleted when not on the keep_list. Directories are only deleted when they are empty after they were processed by the filtered file deletion.
5. Additional maintenance code: remove_installer will walk the directory tree upwards after deleting all files it should delete and remove parent directories in the path of game download files that have become empty. This is mostly used to wipe sub-directories under `.cache/minigalaxy/downloads` when installing a DLC and then remove the now-empty parent dir named after the game.
Walking upwards stops at `config.install_dir/installer` and `.cache/minigalaxy/downloads`, so these root directories are not deleted.
Constructed example: `.cache/download/Rayman/Rayman-inAppPurchase/dlc_installer.exe`. With `keep_installers=False`, `dlc_installer.exe` will be deleted, leaving `Rayman-inAppPurchase` as an empty dir, which is deleted afterwards. Then the parent of `Rayman-inAppPurchase`=`Rayman` is checked, if it is empty and will be deleted if it is. The next parent is `downloads`, which is in the list of iteration stopping dir names. Therefore the check and recursion will stop there.
6. When installing from pre-downloaded files in `LibraryEntry:get_keep_executable_path()`, local executables in the keep directory are sorted after modification time and the most recent is picked for now (in case there are several due to some previous error with deletions).

## Open points:
1. The UI does not yet permit to install pre-downloaded DLCs. The download would be restarted either way, BUT - if DownloadManager determines the files to be identical to what is in the `installer` dir - would complete almost instantly. So we can then still avoid a full redownload of something that's already there.
2. This fix and approach can basically support keeping several old versions in parallel. For this, another setting, like `keep_installer_history` would be required to prevent deletions of old versions in `remove_installer`, and some kind of UI support would have to be added.
3. The current assumption for pre-downloaded installers that all related files share a common name prefix might not necessarily hold in the future. To fix this we could write simple inventory json files when downloads are finished. This would also make maintaining multiple version easier.

## Issues
- fixes #677
- fixes #579
- fixes #513

## Checklist
 
### DIFF
--- a/minigalaxy/config.py
+++ b/minigalaxy/config.py
@@ -104,15 +104,6 @@ class Config:
         self.__write()
 
     @property
-    def keep_installer_history(self) -> bool:
-        return self.__config.get("keep_installer_history", False)
-
-    @keep_installer_history.setter
-    def keep_installer_history(self, new_value: bool) -> None:
-        self.__config["keep_installer_history"] = new_value
-        self.__write()
-
-    @property
     def stay_logged_in(self) -> bool:
         return self.__config.get("stay_logged_in", True)
 
--- a/minigalaxy/config.py
+++ b/minigalaxy/config.py
@@ -104,6 +104,15 @@ class Config:
         self.__write()
 
     @property
+    def keep_installer_history(self) -> bool:
+        return self.__config.get("keep_installer_history", False)
+
+    @keep_installer_history.setter
+    def keep_installer_history(self, new_value: bool) -> None:
+        self.__config["keep_installer_history"] = new_value
+        self.__write()
+
+    @property
     def stay_logged_in(self) -> bool:
         return self.__config.get("stay_logged_in", True)
 
--- a/minigalaxy/game.py
+++ b/minigalaxy/game.py
@@ -21,10 +21,10 @@ class Game:
         self.status_file_path = self.get_status_file_path()
 
     def get_stripped_name(self, to_path=False):
-        return self.__strip_string(self.name, to_path)
+        return self.strip_string(self.name, to_path)
 
     def get_install_directory_name(self):
-        return self.__strip_string(self.name, to_path=True)
+        return self.strip_string(self.name, to_path=True)
 
     def get_cached_icon_path(self, dlc_id=None):
         if dlc_id:
@@ -54,8 +54,9 @@ class Game:
             json.dump(json_dict, status_file)
 
     @staticmethod
-    def __strip_string(string, to_path=False):
-        return re.sub('[^A-Za-z0-9]+', '', string) if not to_path else re.sub('[^A-Za-z0-9 ]+', '', string)
+    def strip_string(string, to_path=False):
+        cleaned_string = re.sub('[^A-Za-z0-9]+', '', string) if not to_path else re.sub('[^A-Za-z0-9 ]+', '', string)
+        return cleaned_string.strip()  # make sure the directory does not start or end with any whitespace
 
     def is_installed(self, dlc_title="") -> bool:
         installed = False
--- a/minigalaxy/game.py
+++ b/minigalaxy/game.py
@@ -21,7 +21,7 @@ class Game:
         self.status_file_path = self.get_status_file_path()
 
     def get_stripped_name(self, to_path=False):
-        return Game.strip_string(self.name, to_path)
+        return Game.strip_string(self.name, to_path=to_path)
 
     def get_install_directory_name(self):
         return Game.strip_string(self.name, to_path=True)
--- a/minigalaxy/game.py
+++ b/minigalaxy/game.py
@@ -21,10 +21,10 @@ class Game:
         self.status_file_path = self.get_status_file_path()
 
     def get_stripped_name(self, to_path=False):
-        return self.strip_string(self.name, to_path)
+        return Game.strip_string(self.name, to_path)
 
     def get_install_directory_name(self):
-        return self.strip_string(self.name, to_path=True)
+        return Game.strip_string(self.name, to_path=True)
 
     def get_cached_icon_path(self, dlc_id=None):
         if dlc_id:
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -14,7 +14,7 @@ from minigalaxy.game import Game
 from minigalaxy.logger import logger
 from minigalaxy.translation import _
 from minigalaxy.launcher import get_execute_command, get_wine_path, wine_restore_game_link
-from minigalaxy.paths import CACHE_DIR, THUMBNAIL_DIR, APPLICATIONS_DIR, WINE_RES_PATH
+from minigalaxy.paths import CACHE_DIR, THUMBNAIL_DIR, APPLICATIONS_DIR, WINE_RES_PATH, DOWNLOAD_DIR
 
 
 def get_available_disk_space(location):
@@ -53,7 +53,8 @@ def install_game(  # noqa: C901
         language: str,
         install_dir: str,
         keep_installers: bool,
-        create_desktop_file: bool
+        create_desktop_file: bool,
+        file_list=None
 ):
     error_message = ""
     tmp_dir = ""
@@ -76,7 +77,7 @@ def install_game(  # noqa: C901
     except Exception:
         logger.error("Error installing game %s", game.name, exc_info=1)
         error_message = _("Unhandled error.")
-    _removal_error = remove_installer(game, installer, install_dir, keep_installers)
+    _removal_error = remove_installer(game, installer, install_dir, keep_installers, file_list)
     error_message = error_message or _removal_error or postinstaller(game)
     if error_message:
         logger.error(error_message)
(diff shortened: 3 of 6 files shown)


minigalaxy/game.py
```diff
@@ -21,10 +21,10 @@ def __init__(self, name: str, url: str = "", md5sum=None, game_id: int = 0, inst
         self.status_file_path = self.get_status_file_path()
 
     def get_stripped_name(self, to_path=False):
-        return self.__strip_string(self.name, to_path)
+        return self.strip_string(self.name, to_path)
```
sharkwouter, line 24: Does this work? I would expect this to fail with a TypeError, because using self provides an additional argument. If that is the case, you could try Game.strip_string().

minigalaxy/installer.py
```diff
@@ -341,31 +342,61 @@ def compare_directories(dir1, dir2):
...
+        for file in os.listdir(installer_dir):
+            file = os.path.join(installer_dir, file)
+            if os.path.isfile(file) and file not in keep_files:
+                logger.info("Deleting file [%s] - not in keep_files", file)
+                os.remove(file)
+
+        # walk up and delete empty directories, but stop if the parent is one of the roots used by MG
+        # this is just maintenance to prevent aggregating empty directories in cache
+        remove_empty_dirs_upwards(installer_dir, installer_root_dirs)
+    except BaseException as e:
```
sharkwouter, line 381:
```suggestion
    except Exception as e:
```
BaseException is probably a bit too broad.

minigalaxy/config.py
sharkwouter, line 1: It seems these properties are not used.

sharkwouter (changes requested):
I wasn't able to test this yet, since I don' have access to my pc at the moment, but I did take a look and found some things that looked odd to me. Otherwise this looks like a nice improvement. Thanks!

minigalaxy/game.py
GB609: I only renamed the method as you can see. So, __strip_string being a static method called with self from within Game is as it was before. But i also tested this with the python interpreter.

But sure, i can change it to use the static call syntax `Game.strip_string` to make it clearer.

minigalaxy/config.py
GB609: You are right. I planned to include the multi-version support from the get-go initially but the code grew too large, so i removed that again. But forgot the additions to config. I'll delete them again.

minigalaxy/game.py
GB609: Done

sharkwouter:
I found an error while testing this:
```
2025-05-13 13:24:54,861 - minigalaxy - ERROR - Error installing game AI War: Fleet Command
  File "/home/wouter/Personal/minigalaxy/minigalaxy/installer.py", line 64, in install_game
    error_message = verify_installer_integrity(game, installer)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wouter/Personal/minigalaxy/minigalaxy/installer.py", line 94, in verify_installer_integrity
    with open(os.path.join(os.path.dirname(installer), installer_file_name), "rb") as installer_file:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
IsADirectoryError: [Errno 21] Is a directory: '/home/wouter/GOG Games/installer/AI War Fleet Command/AI War Fleet Command'
2025-05-13 13:24:54,880 - minigalaxy - ERROR - Unhandled error.
```
I don't know if it is related to this change, though.

sharkwouter:
Sorry I did not elaborate, I just did a quick test. I think what is going on is that I had keep installers enabled while installing the game for the first time and then deleted it. When installing it again, I got this error. Here is some output showing the directory structure:
```
wouter@wouter-laptop:~/GOG Games/installer$ ls -Rl AI\ War\ Fleet\ Command/
'AI War Fleet Command/':
total 312736
drwxr-xr-x 2 wouter wouter      4096 Feb 21 10:38 'AI War Fleet Command'
-rw-r--r-- 1 wouter wouter 320232534 May 13 13:25  ai_war_fleet_command_en_8_024_14437.sh

'AI War Fleet Command/AI War Fleet Command':
total 645572
-rw-r--r-- 1 wouter wouter 661057920 Feb 21 10:40 ai_war_dlc_pack_en_1_0_14437.sh
```

GB609:
No it's fine. Sometimes even quick tests can help find things.
I can see one thing directly: The DLC pack directory name is incorrect. Accoring to the [API](https://api.gog.com/products/1207665333?locale=de&expand=downloads,expanded_dlcs,description,screenshots,videos,related_products,changelog), it should be `AI War: DLC Pack`.

The second issue is that i didn't that the method `verify_installer_integrity` at all. It could be possible that it just doesn't expect sub-directories. I might have to expand the installer files list logic onto that method as well. Or change it a bit to just skip files not in the list of checksums.

sharkwouter:
I'll delete the files and do another quick test.

sharkwouter (approved):
I did another tests and this seems to work well. The one thing I found does seem like a bug, though, but I'm not sure about the nature of it.

GB609:
`verify_installer_integrity` definitely needs an update. It just assumes the content of the installer location dir to be regular files only. When you reinstall the game while there are still some dlcs in sub-directories locally, it will fail.

`LibraryEntry.__download` as well. The problem is that any file that is downloaded will be added to the dict `Game.md5sum` which uses the simple file name as key. This might work in most situations, but is prone to collisions due to name conflicts in some situations.

GB609:
I found another bug in my changes. In `LibraryEntry.get_keep_executable_path`:

```python
for dir_content in os.listdir(self.keep_path):
# ...
    kept_file = os.path.join(self.keep_path, dir_content)
    if os.access(kept_file, os.X_OK) or os.path.splitext(kept_file)[-1] in [".exe", ".sh"]:
        exes_by_creation_date[int(os.path.getmtime(kept_file))] = kept_file
# ...
```

This can't handle directories correctly and possible passes the directory name of a DLC to the installer.

GB609:
I've added another commit to fix the bugs we found right now during the discussion and testing. Feel free to have another look at it.

Issue #681, opened 2025-04-19
"Setup" window should have a setting to disable grabbing focus


TylerCode:
When playing a fullscreen game, this setup window can pop up and continuously grab focus, minimizing the fullscreen game. 

There should be a setting to disable this feature.

sharkwouter:
It would be possible for us to hide the installation window. We could make this an option. Thanks for thinking with us. This is worth looking into, as there are more people who will likely be bothered by this.

GB609:
I'm not sure how we could do this reliably.

I think the setup in question are windows setups. We dropped the full background option `/VERYSILENT` there because it didn't work for a lot of games which require input in the setups, like accepting the Terms and Conditions. If we now go ahead and hide the window, i fear we'll start getting complaints about invisible installs which go on for hours (because the hidden windows actually **require** user input first).

Grabbing and managing focus is the responsibility of the window manager and wine. Bringing windows to the front is handled by the window manager as well. I don't think it is feasible to let minigalaxy fiddle with the configurations of the sheer endless amount of window managers and/or desktop environments. Then there's the difference in behaviour between X and Wayland.

Also, depending on how the actually running game was started, minigalaxy doesn't even have much of a way to detect that it's actually running. Much less so if it is in full screen. Wayland doesn't give the required privileges to other apps and X becomes more and more obsolete.

Possible actions I see are:
1. Don't automatically start an installation after download, but let users trigger them manually.
2. Focus grabbing by wine could be stopped with a registry key (`UseTakeFocus=N`). The problem with that is that when playing a game you often want the exactly contrary behavior because there's an issue with unreal engine permanently losing focus when tabbing out of a game (unless `UseTakeFocus=Y` is set)
3. Maybe it's possible to run the installation in a sub-display like gamescope or Xephyr. But that would add another dependency to minigalaxy.

sharkwouter:
Okay, I wasn't aware. Does it get stuck if we set verysilent or does it fail? If it fails, we could just restart it with silent instead of verysilent. Then we could allow the user to make configure it.

The registry key could work, but is a bit hacky.

GB609:
It just aborts silently. I tested this extensively when i reworked the windows installation. Verysilent just fails with an error code. That lead to a lot of cancellations before. Thats why i introduced the two-step approach with fallback in the first place.

[...]

PR #690, opened 2025-05-17
More improvements to download and install

### LINKED ISSUE #625: Download progress is not being shown correctly
The way it is updated right now doesn't really work since we're no longer doing sequential downloads like before.

### LINKED ISSUE #630: "BIN file was corrupted, please download it again" should smartly resume and perhaps not reload entire game

### LINKED ISSUE #662: Sometimes the download progress bar jumps back and forth between two different values
Sometimes the download progress bar jumps back and forth between two different values:

https://github.com/user-attachments/assets/7ec18543-b7bd-4870-a0aa-3d75134bec10

I'm using the latest minigalaxy version on flathub, which currently seems to be 1.3.2. But I've seen this with older versions too, I don't think it's a recent regression.

### LINKED ISSUE #675: The DLC list seems to stop working once there are too many entries

### LINKED ISSUE #685: Installation of games should be made async to not block download state notification thread
> I've had this yesterday too, but it was not the download itself that was the problem, as you said.
> There are situations where further notifications are only queued and propagated.
> 
> The reason is the finish notification for downloads which runs on the listener thread.
> If that was called and runs in the installation in `installer.py:install_game`, the callback thread is blocked until this returns, which leads to all progress notifications being queued.
> Installations do spawn sub-processes, but use a poll-wait loop for io exchange.
> 
> I think the best way to handle this on the long run is to change how `download.finish` interacts with `LibraryEntry` and `installer.py:install_game`.
> We could introduce an installation queue and make feedback to LibraryEntry async notification based as is done for downloads already, e.g by adding a function `installer.py:queue_installation`:
> 
> ```python
> def queue_install(result_callback, *args): # same args as install_game
>     # save install configuration as json data object to survive early quit,
>     # or let it be re-queued after restart?
> 
>     install_queue.add(InstallTask(result_callback, *args))
> 
>     if not install_thread:
>         install_thread = Thread(target=run_install_queue)
> 
> def run_install_queue():
>     while not install_queue.is_empty():
>         result = install_game(*task.args)
>         result_callback(result)
>         task.task_done()
>         # optional sleep
>     # let the thread die when it's not needed anymore to prevent more obstruction to GIL than necessary
> ```
> 
> The long running parts of install_game are the sub-processes doing the actual installation. These do use a sleep/poll loop already, so they won't fully block the GIL/UI thread.
> 
> There's also the time before the real installation starts when the checksum is being checked. There's no visual feedback for this step at all and it can take a long time for large games. 

 _Originally posted by @GB609 in [#682](https://github.com/sharkwouter/minigalaxy/issues/682#issuecomment-2861996717)_

About the checksums: This step reads all files during installation without any sleep in between consuming chunks.
I think it might be better to use os-level tools for that instead of the custom made implementation. We can then use the regular `_exe_cmd` to allow busy sleep here as well.

### LINKED ISSUE #689: Downloads of large DLC can not be resumed automatically
I planning to also fix this when taking care of parallel DLC downloads, but here's a general rundown of what i think is a pretty inconvenient issue:

`LibraryEntry.__download` marks the game as active download by adding `self.game.id` to the list of active downloads.

However, the `__download` method is also used by (and called from) `__download_dlc`.
That means we get the following combination:
* Active downloads id is the ID of the base game
* The files themselves come from a dlc installer.

After MG restart, in method `LibraryEntry.resume_download_if_expected()`:
* The game id is in active_downloads, but the state of the entry is **NOT** `DOWNLOADABLE`, since it has to be installed already to download DLCs.
* The DLC download is not resumed automatically, it has to be restarted manually by selecting it in the DLC list again.

Why does this matter?
DLCs are not guaranteed to be just small 'key' files which simply unlock pre-packaged items (that's often the case for games with a multiplayer component).
Look at Fallout 4 as an example. It has a DLC `High Resolution Texture pack` which is humongous. Taken from [Fallout 4 API](https://api.gog.com/products/1998527297?locale=de&expand=downloads,expanded_dlcs,description,screenshots,videos,related_products,changelog):

```javascript
id:	"installer_windows_en"
name:	"Fallout 4 - High Resolution Texture Pack"
os:	"windows"
language:	"en"
language_full:	"English"
version:	"1.10.163.0"
total_size:	62522392576
```

It has a total of 14 4GB sized binary files, plus the installer and another binary of around 2GB in size.

To fix this, we should probably pass the id to use into `__download`. Couple this with a bit of code moving from `LibraryEntry` to `DlcListEntry`, then we can locate the right DLC installers data on restart and resume it more easily.

GB609:
## Description

This PR is a bit of a collection of several bug fixes and improvements. I'm aware that it's grown a bit larger, but most of the blocks are connected in a way that makes partial commits and tests pretty difficult.
I've at least tried to organize the biggest blocks into dedicated commits.

## The big chunks/blocks are:

1. ~~There is a method `installer.enqueue_game_install` which is used to feed installation tasks to a separate worker queue and thread. Gotcha here: The worker is only started when there are items in the queue and it dies once it's done. Using this solves the issue of parallel DLC installations, now that i fixed the deletion logic in my previous PR. It also helps to keep the ui more snappy because the install threads in `LibraryEntry` will now be more short-lived. So, there's less threads overall.~~
Introduced with #692 

3. To use the install queue for game installations, i had to change the control flow. It's not really possible to use direct method return values when queueing tasks (unless i want to spawn and block new threads left and right). So i changed the flow in `LibraryEntry` to use callbacks. Install tasks require one `result_callback`. This callback will be invoked by the installation worker when the corresponding installation completes, either with success or failure. There's a dedicated result object, `InstallResult` with an enum-based type `InstallResultType` which allows fine-grained error reporting.  
Most of the common handling is done in `LibraryEntry.__install_finished_callback`.

4. ~~I've encapsulated the `finish_func_wrapper` used in `LibraryEntry._download` into a dedicated class which makes handling of meta data for multi-file downloads easier. This is not yet the full MultiFileDownload implementation. Just a better abstraction of what is already there, plus the extra to better support several parallel DLC downloads.~~
Done with #693

5. ~~I shifted around a lot of code related to DLC. To be precise, download and install methods for DLC went from `LibraryEntry.__download_dlc`, `LibraryEntry.__install_dlc` etc. into `DlcListEntry` which made them much shorter. Arguments for `_install` and `_download` were adjusted to make them more generic, that means `_install_game`, `_install_update`, `_download_game`, `_download_update` could shrink in size.~~
Done with #695

6. ~~State handling in `LibraryEntry` is now a bit more generic. `_download` sets a `predownload_state` When all installations are canceled, state will shift back to `predownload_state`. Then there's a method to change state via argument, only after all downloads are done. Reason: Parallel DLC downloads and the fact that dlc can be downloaded in states `INSTALLED` as well as `UPDATEABLE`.~~
Done with #695

7. `installer.install_game` gets an alternative flow where it won't return an error_message on failure but raise an`InstallException` instead. Used for async installs to transport different types of errors in a type-safe way. This one is tricky. To be honest, i think every install step should utilize this pattern, because it eliminates the repeating pattern of
    ```python
    if not error_message:
        error_message = do_next_step(...)
    ```

    which is repeated 8 times. But changing this would basically mean rewriting a lot of tests. Not adjust or change, rewriting. Because the control flow of all install steps would change. That would definitely have been overkill for now. So i used a little trick and introduced a wrapper function `installer.fail_on_error` and put all install steps into the wrapper separately. When something fails, it raises an exception, which is caught by `install_game`, effectively aborting the install at that point. Depending on the new optional argument `raise_error` it will either return just the message or re-raise the exception. Whenever we rewrite a step to raise exceptions instead, we only need to remove the wrapper, but the basic structure of `install_game` does not have to be touched anymore.

8. A few helper/conveniences in `Game` and `Api`.

9. Test for the new stuff. I've actually already found (and fixed) errors in the new code thanks to tests. I should probably add some more.

I have not fully tested everything yet and will likely do some more tests in the next few days. But since it's quite a lot to go through i've decided to publish the code early.

## Issues:
- fixes #689
- fixes #685
- fixes #675 - we already have the ui in a grid, but this PR solves the issue of potentially parallel dlc installs
- fixes #630 - remove_installers is now the last step of regular install. that means that, even with keep_installers=False, files will only be deleted on success. Exception: on checksum errors. Even then, only corrupted files will be deleted. I think this matches the issue title as far as currently possible. We'll have to see if there are still frequent corruptions and what we can do then.
- fixes #625 
- fixes #662 

## Roadmap
After this, there would be 1-3 things i would like to consider for the next release.
But, if we're in a rush, we can also go without them i guess.

1. UI: Now that parallel DLC downloads and installs are possible, maybe introduce a `Download all` button at least. We could also enhance the DLC grid with check boxes (or something else suitable) to just install selected. We can also leave this out. I'm not sure if there really are that many games like `Potion Permit` with like hundreds of DLCs.

2. ~~A first fix of progress calculation in `LibraryEntry`. Only the part that adds up percentages up.~~ - Included in this PR.
For full support of MultiFileDownloads some more stuff is necessary, like handling in the download ui, `DownloadManager`, proper handling of cancel per download group and placing one of the 'new' button panels (as used in the download ui) on `GameTile` for the MultiFileDownload while there is something going on.

3. Maybe another small go at the wine installer, to give an option between `/VERYSILENT` and `/SILENT` and try to see if there's a way to get rid of the focus/input grabbing (which seems to be a wine issue). See #681 

## Checklist
 

GB609:
Issues i found now:
- State doesn't change correctly everywhere. Caused by wrong order of actions in FinishFuncWrapper

- This is an old one: checksums retrieved from the api are not persisted in game info.
That means that offline reinstalls (after restart) can never check them.
This also means that there's no real way to detect that all files have been downloaded once minigalaxy has restarted (when `keep_installer=True` is used). So a game consisting of a (small) `.exe` and several `*.bin` files might be considered 'complete' after restart even if the download was aborted before as long as the `.exe` is found by `LibraryEntry.get_keep_executable_path`. None of the `.bin` files would have to be completed for this, leading the the installer asking for additional install media/files.
I think we should change that. Checksum should be persisted with `game.set_info`. I'd suggest the following structure:
    ```python
    checksums = {
        "base_game_id" : {
            "filename": FileInfo
        },
        "dlc_id" : {
            "filename": FileInfo
        }
    }
    ```
    Downloading something will set or update the values. And they can be used for offline installation and to make the behavior of `LibraryEntry.get_keep_executable_path` more accurate.

sharkwouter:
Damn, this is a massive improvement. I just tested it and it seems to work great and fix most of the issues I was seeing before.

One thing I noticed is that the delete files button is not working, though. I mean this one:
image

Another one is that if a download is still going, it does seem closing the game does not stop the process despite the changes.

sharkwouter:
I think the process did not exit because I still had an installation running. Let me retry.

sharkwouter:
I did see this when closing just now:
```
warning [/home/wouter/.cache/minigalaxy/download/Ascendant/gog_ascendant_2.2.0.7.sh]:  689879 extra bytes at beginning or within zipfile
  (attempting to process anyway)
Exception in thread Thread-5 (<lambda>):
  File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.11/threading.py", line 975, in run
  File "/home/wouter/Sources/minigalaxy/minigalaxy/download_manager.py", line 156, in <lambda>
    download_thread = threading.Thread(target=lambda: self.__download_thread(queue))
                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wouter/Sources/minigalaxy/minigalaxy/download_manager.py", line 442, in __download_thread
    self.__download_file(download, download_queue)
  File "/home/wouter/Sources/minigalaxy/minigalaxy/download_manager.py", line 493, in __download_file
    self.__notify_listeners(result, download, additional_params=additional_info)
  File "/home/wouter/Sources/minigalaxy/minigalaxy/download_manager.py", line 196, in __notify_listeners
    self.listener_thread.submit(self.__notify_listeners,
  File "/usr/lib/python3.11/concurrent/futures/thread.py", line 167, in submit
    raise RuntimeError('cannot schedule new futures after shutdown')
RuntimeError: cannot schedule new futures after shutdown
```
It did close, though.

sharkwouter:
I did notice the state issue. While installing, I saw a game still marked as `Downloading...`.

GB609:
I was thinking about a print statement on shutdown when there's still an installation running. Thats why i mentioned the logging concept.
And it seems i forgot about the download notification threadpool on shutdown.

But we'll find all those issue and iron them out, i'm sure.

GB609:
> I did see this when closing just now:
> 
> ```
> warning [/home/wouter/.cache/minigalaxy/download/Ascendant/gog_ascendant_2.2.0.7.sh]:  689879 extra bytes at beginning or within zipfile
>   (attempting to process anyway)
> Exception in thread Thread-5 (<lambda>):
> Traceback (most recent call last):
>   File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
>     self.run()
>   File "/usr/lib/python3.11/threading.py", line 975, in run
>     self._target(*self._args, **self._kwargs)
>   File "/home/wouter/Sources/minigalaxy/minigalaxy/download_manager.py", line 156, in <lambda>
>     download_thread = threading.Thread(target=lambda: self.__download_thread(queue))
>                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/wouter/Sources/minigalaxy/minigalaxy/download_manager.py", line 442, in __download_thread
>     self.__download_file(download, download_queue)
>   File "/home/wouter/Sources/minigalaxy/minigalaxy/download_manager.py", line 493, in __download_file
>     self.__notify_listeners(result, download, additional_params=additional_info)
>   File "/home/wouter/Sources/minigalaxy/minigalaxy/download_manager.py", line 196, in __notify_listeners
>     self.listener_thread.submit(self.__notify_listeners,
>   File "/usr/lib/python3.11/concurrent/futures/thread.py", line 167, in submit
>     raise RuntimeError('cannot schedule new futures after shutdown')
> RuntimeError: cannot schedule new futures after shutdown
> ```
> 
> It did close, though.

This seems like there is another notification from download manager triggered by install.
Or some racing condition on shutdown while there are still ongoing downloads. `DownloadManager` needs a proper shutdown method that also handles the notification thread. Just stopping the downloads isnt enough anymore.

GB609:
> Damn, this is a massive improvement. I just tested it and it seems to work great and fix most of the issues I was seeing before.
> 
> One thing I noticed is that the delete files button is not working, though. I mean this one: image
> 
> Another one is that if a download is still going, it does seem closing the game does not stop the process despite the changes.

About the delete action: i've noticed something similar before.
It might be related to the bug in the download ui i found before, or download manager cancel with state `CANCELED` (which does the deletion) doesnt find/accept stopped downloads anymore (which might also be a problem for `LibraryEntry.confirm_and_cancel`)

sharkwouter:
Maybe we should make more use of signals. That is a way to communicate between threads. I think we could use one in the download manager threads.

sharkwouter:
It might be worth merging this as is and making smaller PRs to fix the issues that are still there. That would speed up reviews from my side. What do you think?

GB609:
> Maybe we should make more use of signals. That is a way to communicate between threads. I think we could use one in the download manager threads.

If you are referring to [this](https://docs.python.org/3/library/signal.html#signals-and-threads), it might not be the most suitable tool. The docs say signals cant be used for inter-thread communication because handlers will always be executed in the main thread.
I would have added some flag or condition in download manager that is checked before something is taken from a queue. If it is set, break out of the poll loop.
In total:
1. Remove daemon flag
2. Set flag, eg. `is_running` to false
3. Stop all downloads so that the threads start the next iteration where...
4. ...they check the flag and die in an orderly fashion.

GB609:
> It might be worth merging this as is and making smaller PRs to fix the issues that are still there. That would speed up reviews from my side. What do you think?

If you want to go on with this in that way we can do this. It's just that i normally try to keep the master working stable.

I can also break out the additions to installer.py first. They're mostly done and more or less independent

GB609:
I've updated this PR, but it's not ready. I can't test with this machine yet.
My changes regarding graceful shutdown keep the tests from exiting correctly.
I'll fix it this evening

minigalaxy/installer.py
```diff
@@ -786,14 +783,26 @@ def __locate_game_in_args(*args, **kwargs):
                 return a
         raise ValueError("No instance of Game in InstallTask constructor arguments")
 
+    @staticmethod
+    def __get_title_for_id(game: Game, item_id):
```
sharkwouter, line 787: I would remove the underscores here. A static function shouldn't have that.

sharkwouter:
Let me know when I should take another look at this one.

GB609:
I will. 
I still got an issue that the tests run by github don't terminate. Even though they do locally.

GB609:
Update: I know the reason now.
I've used stuff in the code that only exists from python version 3.13 onwards.
This works locally without a problem, but it throws exceptions on github which uses python 3.8.

Before i can fix this and continue with that, i need to know which python version we should target because i actually need to change code related to graceful shutdown depending on the version.

3.13 is relatively new. Using that might mean dropping support for some distributions which haven't or won't update.

sharkwouter:
Canonical just dropped support for Ubuntu 20.04 like 2 days ago, that was still using python 3.8.

Right now distros have the following versions:
- Debian 12: 3.11
- Ubuntu 22.04: 3.10

There are some CentOS based distros that still over python 3.9, though.

I would say we can update to 3.9 for sure. Maybe 3.10.

GB609:
Then lets try updating the github runner to 3.10. Still won't be enough for that one method parameter, but at least i've got a definitive number to work with.
It'll work without that option, too.

GB609:
Alright. I think i've got it. I've also added a quick fix for the percentage issue in games with multiple downloads.
As mentioned in the PR description, this is not the full MultifileDownload support including better handling in the download ui. But it's enough that the progress bar won't be jumping around anymore.

- Fixes #625 
- Fixes #662

minigalaxy/download.py
```diff
@@ -86,8 +86,75 @@ def on_cancel(self, callback):
...
+        self.__progress_func = callback
+
     def __str__(self):
         return self.filename()
 
     def __repr__(self):
         return self.filename()
+
+
+class CombinedProgressWatcher():
```
sharkwouter, line 100: I think this class should have its own file. It also looks like something that should be unit test-able. What do you think? Would it be worthwhile adding 1 or 2 tests?

GB609: Sure. I'll change and add that once i got the time.

sharkwouter: Cool, thanks for all the work you put into this :) I'm amazed by the result, it's so much better.

GB609: Thanks.
There are a few areas that could use more improvements though. The entire usage of a PriorityQueue and polling threads for downloads seems like overkill to me. Or rather, it's not the right utility because canceling items from such a queue is more complicated than it has to be and the priorities aren't really used. I even thought about using something similar to the new InstallerQueue instead. But that can wait until after the release. It works as it is right now. The code is just much more complicated than it has to be.

GB609:
Ok, i'm done with my changes. Sorry for the delay, i was busy the last few days.

I re-organized the files a bit when splitting up download.py, because i wanted to still keep them close in a way that indicates they belong together.
So i made a sub-folder and used the same pattern that is used in the `ui/__init__.py` to keep the current imports used everywhere as they are.

I've also added a batch of tests for `CombinedProgressWatcher`.

sharkwouter:
Looks good, thanks for making these changes

sharkwouter:
That is fail, lets keep it at 1.4.0. I was thinking 2.0.0 could work, since we've come a long way from version 1.0.0, but we also don't want to communicate that this release specifically causes breakage.

PR #692, opened 2025-05-22
Feature/async install

GB609:
## Description

This is a part of #690, created in an attempt to reduce review efforts.

I've split out the new capabilities of `installer.py` for async installations into the new PR.
It only contains the additions in `installer.py`, plus changes in tests to make sure it is working. The async install is NOT used by `LibraryEntry` yet.

Some minor changes in this PR might have an impact on installations (and the UI), even now:
1. I've re-ordered the installation steps to something that made more sense to me. Most notable: Deleting installer files is NOT a guaranteed quasi-finally step after the installation. It is part of it. As such, an installation that might fail because of disk size issues will NOT delete downloaded installers. Exception to this are failures because of checksums. Even then i've tried to only delete the files which are actually invalid.
2. I've increased the chunk size used when reading files for checksum calculations from 4KB to 8KB. My PC is pretty old and that gave a noticeable speed increase (around 0.7 less time taken for the installer of Coromon, which is one file of about 900MB, think about games of 10+ files with 4GB in size). I plan to do a bit of experimentation and time measurement because i think we can increase it some more, to 16k or 32k. The larger the buffer, the less often python needs to switch context/threads between io and the python process (which also has a cost). It also means longer slices of time where other threads like the UI/Main thread can run. [Python hashlib](https://docs.python.org/3/library/hashlib.html) docs state that it releases the gil when data > 2047 bytes is passed. As such, increasing the buffer mostly means less reads, less context changes and opens up the GIL much more. We're generally reading files with hundreds of megabytes or even gigabytes in size. Reading them in baby steps of 4K on modern computers seems like overkill.

## Open points
1. To properly handle incomplete/invalid downloads and partial re-download on errors, we need a bit more code and persistent data from `LibraryEntry._download` to correctly detect that all parts of a game have been downloaded. This mostly concerns situations with resuming downloads after restart when `keep_installers` is enabled.
I haven't touched that yet.

2. Graceful shutdown: For this, some additions to the `InstallerQueue` class are required. I'll introduce this the moment we update `LibraryEntry` to use async installations. For now, it wouldn't make a difference anyway and just increase the code to review.

## Checklist
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - The DLC list will now show a table-like multi-column view where necessary. This fixes issues with games having a high number of DLCs. (thanks to GB609)
 - Some bug fixes in DownloadManager related to canceling active and queued downloads (thanks to GB609)
 - Fix deletion logic of downloaded installation files. Base installers and DLC will not interfere with each other anymore. (thanks to GB609)
+- Failed installs will try to keep as many downloaded parts as possible, even without the keep_installers option enabled. (thanks to GB609)
 
 **1.3.2**
 - Completely reworked windows wine installation. This should solve a lot of problems with failing game installs. Innoextract (if installed) is only used to detect and configure the installation language. (thanks to GB609)
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -1,12 +1,17 @@
-import sys
+import hashlib
 import os
-import shutil
+import re
 import shlex
 import subprocess
-import hashlib
+import shutil
+import sys
 import textwrap
 import time
-import re
+
+from collections import deque
+from enum import Enum
+from queue import Empty
+from threading import Thread, RLock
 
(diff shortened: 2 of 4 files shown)


GB609:
@sharkwouter 
I'll leave it up to you to decide if you want to use the step-by-step or all-at-once approach.
If you decide to merge this one, i'll rebase and update the other PR accordingly.

sharkwouter:
This one looks good and works well. Thanks for the work on this!

sharkwouter:
I'm going to leave it up to you how you want to handle this. I think reviewing in smaller chunks is a bit easier, but I would also be okay with you just merging the PR that is still open and making small tweaks in smaller PRs to fix some of the leftover issues. The main goal is to have a release that works well in the end and both ways can get us there.

GB609:
Alright. It'll probably be a mix of both.

PR #693, opened 2025-05-23
Refactors to prepare multi-file and parallel dlc downloads

### LINKED ISSUE #630: "BIN file was corrupted, please download it again" should smartly resume and perhaps not reload entire game

GB609:
## Description

This PR is based on a part of the commit https://github.com/sharkwouter/minigalaxy/pull/690/commits/09259019bbbf53c40e925efefb27fa6e00a8f78f from #690.

I've initially only split out the refactoring part where i introduced the FinishFuncWrapper.
But this didn't work on its own with clean-up and other stuff, so i decided to add another new fix here related to checksums, `keep_installer` and `LibraryEntry.get_keep_executable_path` because i was frequently running into situations where just having the exe downloaded was recognized as the game being downloaded and ready for install.

To fix this once and for all times, i introduced another class and data file: `InstallerInventory`.
It's basically a helper class which is backed by a json file containing a dict file names, their sizes and checksums as we have up to now, but now it's persisted in the same directory where the installer files go. Up to now, checksums were set as property of game with `game.md5sums` which was just a runtime property, nothing persisted.
InstallerInventory files include themselves in their deletion logic, so canceling a download will also remove the inventory file and, if empty, its parent directory.

At first, this might appear to be pretty big PR, but the new class i've introduced alone has around 150 lines.

This has the following effects:
- I have a way to check if all files of a game have been downloaded when minigalaxy is restarted.
- I can just delete one file with an invalid checksum during install and keep all others, throw an error and `get_keep_executable_path` will not find a path => Download button is active and the user won't have to manually delete partially downloaded files anymore.
- I can encapsulate deletion logic for games with multiple files there
- Checksums can be verified when the game is installed a second time, long after it's been downloaded. 
- This also means that we could introduce an option to not automatically immediately start installations after a download, even for files in cache.
- It also means we could in theory split up the resume after restart feature, even for installs from cache. There's no need to re-queue all downloads and tax the api just to realize they're complete and then go to install after download manager returns.
- possibly more i can't see yet

This PR fixes #630 as i'm trying to do all thats possible to keep as many files as possible. The only thing still left to do would be search for other causes of corruption during download. I think the fact that download threads are daemons which are just killed on shutdown _might_ also lead to corruption sometimes. But fixing that is already on the list.

I've also added a rudimentary first interaction between download ui / states and the classic gametile controls: When a file  assumes state `CANCELED` by clicking on 'delete file' after stopping, `LibraryEntry confirm_and_cancel` will be triggered (once for multi-file downloads). The way the state is checked is a little hacky for now, but this will change in the future.

I think i need to do all the rest at once after this.
Allowing multiple DLC downloads at once only makes sense when they can be queued for install. For this i need to change method signatures and add parameters which are easier to handle when i also change the DLC specific methods to be located in DlcListEntry.

## Checklist
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 - Some bug fixes in DownloadManager related to canceling active and queued downloads (thanks to GB609)
 - Fix deletion logic of downloaded installation files. Base installers and DLC will not interfere with each other anymore. (thanks to GB609)
 - Failed installs will try to keep as many downloaded parts as possible, even without the keep_installers option enabled. (thanks to GB609)
+- Deleting a file in download ui now also provides the option to fully cancel the game download as if cancel on the thumbnail was clicked. (thanks to GB609)
+- With keep_installers enabled, Minigalaxy can now detect if a game has been fully downloaded. If not, the UI will still show the 'Download' button. (thanks to GB609)
 
 **1.3.2**
 - Completely reworked windows wine installation. This should solve a lot of problems with failing game installs. Innoextract (if installed) is only used to detect and configure the installation language. (thanks to GB609)


sharkwouter:
Thanks! The weekend was a bit busier than expected. Looking forward to the next one.

PR #699, opened 2025-06-14
Bugfix/cancel downloads

GB609:
## Description

2 small bugfixes:
- i refactored away too much error handling before. Getting a bad result for `api.get_download_info` would have run into `LibraryEntry.__download` with `download_info=False` (where a dict is expected), leading to faulty behaviour
- Resuming multiple DLCs at once after restart didn't work before because the first restart changed the game state to Downloading, which prevented the other resumes. I've improved the check.

The primary change:
- `LibraryEntry.__download`, `CallbackFuncWrapper` and `LibraryEntry.confirm_and_cancel` now use an intermediate data class, `InstallableItem` to bundle id and name to use in the process
- `LibraryEntry.prevent_resume_on_startup` gets an additional parameter `item_id`. If not given, all ids of the game and its DLCs will be removed from active downloads in config (if contained). If given, only the respective download is stopped.

To test those, i locally patched the `DownloadManager` to not queue items with type `DownloadType.GAME_DLC` (and let cancel_download abort them unconditionally to get listener callbacks)

## Checklist
 
 - [x] _CHANGELOG.md_ was updated


sharkwouter:
Thanks! This works very well.

PR #700, opened 2025-06-18
Disable webkit compositing mode by default

sharkwouter:
Otherwise OpenSUSE, Fedora and Arch will not display the login screen.

## Description

I'm still not 100% sure if this is the way to go, but I do think it would solve the issue. Now I just need to test it to make sure it does. Any opinions on this solution?

## Checklist
 

sharkwouter:
I did some testing with it on Fedora 42 and it does solve the problem there.

PR #702, opened 2025-06-30
Fix #698

### LINKED ISSUE #698: It never start
When i try to open the program it start for a second than crash and never reopen.

sharkwouter:
## Description

On some older Nvidia cards, the dmbuf renderer does not work, so if we want Minigalaxy to launch on these systems we should disable it.

## Checklist
 

GB609 (approved):
looks good to me

PR #714, opened 2025-08-11
bugfix: fix config reading issue for certain characters in path

kyle-zhang-42:
Added `encoding="utf-8"` to both read and write methods to ensure that characters such as CJK characters in game library path can be read correctly.

## Checklist
 

GB609:
Looks good to me, @sharkwouter shall we go ahead with it?

sharkwouter:
Thanks! That is a good catch and a nice change!

Issue #715, opened 2025-08-11
Translators using Weblate are missing from the credits and README


sharkwouter:
Maybe this could be automated. We do create commits in the release pipeline.

GB609:
With a bit of git magic. But i'd keep the list in a separate file to avoid merging hell with other files and influences.
Something like going through history and generating a simple text file from all authors (but we'd have to prevent duplicates caused by several commits).
Maybe only take the commits down to the previous tag and add append new section to the translators file.

sharkwouter:
Thanks for the work on this :D

PR #716, opened 2025-08-21
Credits for weblate contributions

GB609:
## Description
This PR is related to #715.

Up to now, i've only prepared a very basic script/logic that just searches git log for all commits commited by Weblate, then extracts the author names. With a bit of logic to prevent duplication.

This input is used to generate a simple name listing file - `weblate-translators.md`, overwriting the previous one, if any.
It is not yet included into any workflow.

I'd like to discuss on how to proceed.
The simplest solution would be just add a link to the end of `README.md` that points to this file and update it with a bit of scripting during releases.

For now, that file makes no difference about when a contribution happened.
I could also change the code to only look at all commits between the currently running release and the previous one (i'd go by git tag for that) and instead just append the new constributions as a new section.
I could also enhance the script to take arguments, like commit ranges, file text prefixes etc. Whatever we might need.

## Checklist
 
### DIFF
--- a/scripts/credit-weblate-translators.sh
+++ b/scripts/credit-weblate-translators.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+cd "$(dirname "$0")"/..
+
+TRANSLATORS_FILE="weblate-translators.md"
+
+function fileHeader {
+  echo "# Weblate translations"
+  echo -e "Special thanks to Weblate and the following users:\n"
+}
+
+function createTranslationCredits {
+  declare -A translators
+
+  while read; do
+    author="${REPLY#Author: }"
+    author="${author%% <*>}"
+
+    read language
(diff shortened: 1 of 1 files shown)


sharkwouter:
This is a good start. I would like to automate this more as part of the release process, but I think this makes a lot of sense to have for now.

sharkwouter:
Thanks for the work on this!

GB609:
Anytime. The script was meant to be included into automation as well. I just didn't touch that yet as i wasn't sure of how you'd like the final result to look like.

sharkwouter:
I'm thinking we should probably parse the about.ui and the README.md and add lines for Weblate contributors that are not in there. We can do that upon release.

sharkwouter:
Cool, have a great vacation :)

PR #718, opened 2025-09-07
improve weblate contribution script

GB609:
This is a follow up on #716.
I've reworked the script and also included it into the release workflow.

* Contributions are now patched into `about.ui` and `README.md` instead of generating a dedicated file.
* The script `scripts/credit-weblate-translators.sh` will be executed in `release.yml` workflow just before `scripts/create-release.sh` is used.

In `README.md`, I've kept the format of all lines as they are. For the about ui, translations coming from Weblate are listed differently. There's no real, reliable way to get a link to the contributors, because they are effectively coming from Weblate, they theoretically don't even need to have a github account.
So, weblate contributions are appended in the following format:

```
Author name (Weblate)
```

Small gotcha: There's no filter for names (yet), so all weblate contributions are listed, including `Wouter Wissman`.

In `about.ui` i'm using `(Weblate)` as a filter to detect pre-added lines. With that, all entries are recreated here.
In `README.md` i'm parsing the current content first to find lines that have been added before. Afterwards, the new lines are appended to the end of the file. So i'm assuming that the current format of the file is kept for now. If that ever changes, we could for example use marker comments. But we can think about this when it actually happens.

## Checklist
 
### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,6 +17,7 @@ jobs:
     - name: Prepare release files
       id: tag
       run: |
+        ./scripts/credit-weblate-translators.sh
         ./scripts/create-release.sh
       env:
         DEBFULLNAME: ${{ secrets.DEBFULLNAME }}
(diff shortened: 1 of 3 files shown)


sharkwouter:
Awesome! Thanks a lot!

sharkwouter:
I guess we could exclude me from it. I'll push a commit for that.

PR #731, opened 2026-02-19
fix: resolve test errors and deprecation warnings

slowsage:
## Description
Installing minigalaxy-git on arch fails (`paru -S minigalaxy-git`) with errors and warnings:
```
__________________________________________ ERROR collecting tests/test_ui_library.py __________________________________________
ImportError while importing test module '/home/$USER/.cache/paru/clone/minigalaxy-git/src/minigalaxy/tests/test_ui_library.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.14/importlib/__init__.py:88: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/test_ui_library.py:9: in <module>
    from MockGiRepository import MockGiRepository
E   ModuleNotFoundError: No module named 'MockGiRepository'
__________________________________________ ERROR collecting tests/test_ui_window.py ___________________________________________
ImportError while importing test module '/home/$USER/.cache/paru/clone/minigalaxy-git/src/minigalaxy/tests/test_ui_window.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.14/importlib/__init__.py:88: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/test_ui_window.py:6: in <module>
    from MockGiRepository import MockGiRepository
E   ModuleNotFoundError: No module named 'MockGiRepository'
====================================================== warnings summary =======================================================
minigalaxy/api.py:235
  /home/$USER/.cache/paru/clone/minigalaxy-git/src/minigalaxy/minigalaxy/api.py:235: SyntaxWarning: 'return' in a 'finally' block
    return result

minigalaxy/translation.py:32
  /home/$USER/.cache/paru/clone/minigalaxy-git/src/minigalaxy/minigalaxy/translation.py:32: DeprecationWarning: 'locale.getdefaultlocale' is deprecated and slated for removal in Python 3.15. Use setlocale(), getencoding() and getlocale() instead.
    default_locale = locale.getdefaultlocale()[0]
```

- ERROR: Add pytest pythonpath config for test discovery
- WARN: Replace deprecated locale.getdefaultlocale() with getlocale()
- WARN: Use explicit 'is not None' for XML Element check
- WARN: Move return statement out of finally block

Ran `python3 -m pytest -v -W error` to ensure no warnings remain.
Note: Not super confident of locale related testing since I only have English.

## Checklist
 
### DIFF
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -224,15 +224,14 @@ class Api:
             response = self.session.get(url)
             if response.status_code == http.HTTPStatus.OK and len(response.text) > 0:
                 response_object = ET.fromstring(response.text)
-                if response_object and response_object.attrib:
+                if response_object is not None and response_object.attrib:
                     result = response_object.attrib
             else:
                 logger.error("Couldn't read xml data. Response with code %s received with the following content: %s",
                              response.status_code, response.text, exc_info=1)
         except requests.exceptions.RequestException:
             logger.error("Couldn't read xml data. Received RequestException", exc_info=1)
-        finally:
-            return result
+        return result
 
     def get_user_info(self) -> str:
         username = self.config.username
--- a/minigalaxy/translation.py
+++ b/minigalaxy/translation.py
@@ -29,7 +29,7 @@ os.unsetenv("LANGUAGE")
 os.unsetenv("LANG")
 
 current_locale = Config().locale
-default_locale = locale.getdefaultlocale()[0]
+default_locale = locale.getlocale()[0]
 if current_locale == '':
     if default_locale is None:
         lang = gettext.translation(TRANSLATION_DOMAIN, LOCALE_DIR, languages=['en'], fallback=True)
--- a/minigalaxy/ui/library.py
+++ b/minigalaxy/ui/library.py
@@ -33,7 +33,8 @@ class Library(Gtk.Viewport):
         self.config = config
 
         current_locale = self.config.locale
-        default_locale = locale.getdefaultlocale()[0]
+        locale.setlocale(locale.LC_ALL, '')
+        default_locale = locale.getlocale()[0]
         if current_locale == '':
             locale.setlocale(locale.LC_ALL, (default_locale, 'UTF-8'))
         else:
--- a/minigalaxy/ui/preferences.py
+++ b/minigalaxy/ui/preferences.py
@@ -62,7 +62,7 @@ class Preferences(Gtk.Dialog):
 
         # Set the active option
         current_locale = self.config.locale
-        default_locale = locale.getdefaultlocale()
+        default_locale = locale.getlocale()
         if current_locale is None:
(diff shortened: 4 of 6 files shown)


GB609:
Hey. Thanks for the PR and sorry for the delay. 
I'm also using ArchLinux, so i would like to test and verify your changes first. I'll come back to you as soon as i can.

slowsage:
closing - the PKGBUILD is broken as well (since the package webkit2gtk was removed from arch AUR last week). Closing. Will post another PR once that is resolved.

GB609:
First let me apologize again for leaving this PR unattended for so long. I had very little spare time recently and - honestly - kind of forgot about it over that.

There are 2 entries in the AUR for minigalaxy:
- [minigalaxy](https://aur.archlinux.org/packages/minigalaxy): Last updated on 2026-01-22 17:15.  
  This is the current release-based variant and it works, i just right now tested it with a freshly upgraded arch installation
- [minigalaxy-git](https://aur.archlinux.org/packages/minigalaxy-git): Last updated on 2022-04-13 15:35.  
  I haven't tested it, but it is very likely not to work. The rolling-release nature of arch linux and its frequent package renames mean that dependencies have to be kept up-to-date which isn't the case here. `-git` package variants might make sense for mostly self-sufficient small utility-packages, not for integrators with larger lists of (os-native) dependencies. (Unless the `-git`-package build provides some clever way of package resolution by itself.)

If the current build and dependencies is all you want to test, you can try to install the latest package once to get all dependencies downloaded. Minigalaxy itself can also be started directly from its repository with `bin/minigalaxy` (which is what i'm doing all the time).

That being said, i've ran the tests in several variants from within eclipse and the command line with `python3 -m unittest discover -v ./tests`.

I do get the `SyntaxError` and `DeprecationWarnings` you mention above (but only once, because they are compile warnings and once the tests run from `__pycache__`, the warnings disappear). But i don't get the path or import errors.

However, minigalaxy 1.4.1 targets python 3.10, the current master updated to 3.11. Both do not mark `getdefaultlocale` as deprecated.

If we can replicate the same behaviour with your changes and that API already exists in 3.11, we can change the code related to locales, although i'm not sure about that. According to the docs, the behaviour of the new methods is different - they read different variables in a different order. I think it's kind of a weird move from python that they drop `getdefaultlocale` which was by far the most flexible one. Not like i can change anything about that decision, sadly.

I'm with you in that we have to find a different solution. Otherwise we'll run into compatibility issues between slow-updating distributions and fast-moving ones.

Anyway, i'll come back later with more details. I'm running out of time for now.

PR #732, opened 2026-05-04
Feature: Improve handling of application configuration

GB609:
## Description
I'm currently working at a larger feature that is related to the handling of multiple platforms where i also plan to introduce a couple additional settings.

This PR is a small preparation for that. I'm more or less doing a bit of code maintainance while i'm adding stuff in `preferences.py` and `config.py` anyway.

With this, I'm introducing a reusable private method, `Config.__set_and_write(key, new_value)` that takes care of the ever-repeating lines 

```python
self.__config[prop_name] = new_value
self.__write()
```

The new method does 2 additional things:
1. Check if the new value really is different (or do nothing otherwise)
2. Only write to file directly when `Config` is not edited in 'batch mode'. This is primarily intended for usage from the `Preferences` dialog's `save_pressed`.

Both additions are made to reduce unnecessary file writes when modifying preferences.

## Checklist
 
### DIFF
--- a/minigalaxy/config.py
+++ b/minigalaxy/config.py
@@ -207,38 +207,38 @@ class Config:
 
     @current_downloads.setter
     def current_downloads(self, new_value: List[int] = []) -> None:
-        self.__set_and_write("current_downloads", [*new_value])
+        self.__set_and_write("current_downloads", new_value)
 
     def add_ongoing_download(self, download_id):
         '''Adds the given id to the list of active downloads if not contained already. Does nothing otherwise.'''
-        current = self.current_downloads
+        current = [*self.current_downloads]
         if download_id not in current:
             current.append(download_id)
             self.current_downloads = current
 
     def remove_ongoing_download(self, download_id):
         '''Removes the given id from the list of active downloads, if contained. Does nothing otherwise.'''
-        current = self.current_downloads
+        current = [*self.current_downloads]
         if download_id in current:
             current.remove(download_id)
             self.current_downloads = current
 
     @property
-    def paused_downloads(self) -> dict[str, int]:
+    def paused_downloads(self) -> dict:
         return self.__config.get("paused_downloads", {})
 
     @paused_downloads.setter
     def paused_downloads(self, new_value={}) -> None:
         """Ongoing downloads are a dictionary of filename:progress_percentage_int. See Download.current_progress."""
-        self.__set_and_write("paused_downloads", {**new_value})
+        self.__set_and_write("paused_downloads", new_value)
 
     def add_paused_download(self, save_location, current_progress):
-        paused = self.paused_downloads
+        paused = {**self.paused_downloads}
         paused[save_location] = current_progress
         self.paused_downloads = paused
 
     def remove_paused_download(self, save_location):
-        paused = self.paused_downloads
+        paused = {**self.paused_downloads}
         if save_location in paused:
             del paused[save_location]
             self.paused_downloads = paused
--- a/minigalaxy/config.py
+++ b/minigalaxy/config.py
@@ -18,6 +18,7 @@ class Config:
     __config: dict
 
     def __init__(self, config_file: str = CONFIG_FILE_PATH):
+        self.__is_batch_edit = False
         self.__config_file = config_file
         self.__config = {}
         self.__load()
@@ -40,6 +41,24 @@ class Config:
             file.write(json.dumps(self.__config, ensure_ascii=False))
         os.rename(temp_file, self.__config_file)
 
+    def start_batch_edit(self):
+        """
+        Flags config for multiple incoming changes. Normally, every change is written to file immediately.
+        This can lead to a lot redundant writes when several settings are changed in quick succession.
+        This method provides a uniform way to flag Config in a rudimentary type of 'transaction'.
+        Any setter which internally uses '__set_and_write' will honor this flag.
+        Use 'Config.save()' to finalize (remove flag and write to the file to disc).
+        """
+        self.__is_batch_edit = True
+
+    def cancel_batch_edit(self):
+        """
+        The counterpart to 'Config.start_batch_edit()'.
+        Resets the flag and reverts all changes since the last write/save.
+        """
+        self.__is_batch_edit = False
+        self.__load()
+
     def save(self) -> None:
         """
         Config will normally immediately save all changes applied to it to the configuration file automatically.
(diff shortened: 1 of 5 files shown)


minigalaxy/ui/preferences.py
```diff
@@ -178,31 +179,40 @@ def __save_install_dir_choice(self) -> bool:
...
-            if self.switch_show_windows_games.get_active() and not shutil.which("wine"):
-                self.parent.show_error(_("Wine wasn't found. Showing Windows games cannot be enabled."))
-                self.config.show_windows_games = False
-            else:
-                self.config.show_windows_games = self.switch_show_windows_games.get_active()
-                self.parent.reset_library()
+        try:
+            self.config.start_batch_edit()
+
+            self.__save_locale_choice()
```
sharkwouter: Maybe we should rename these `__save_xxxxx` functions to `__apply_xxxxx` instead. I was confused by this. Not directly related to this PR, but maybe we could include it.

minigalaxy/config.py
```diff
@@ -48,153 +67,147 @@ def save(self) -> None:
...
         return other none-simple types in the future. Changes made to these sub-objects would not be detected
         and thus also not be persisted automatically.
-        In these situations `Config.save`  can be used to avoid having to re-assign the same object reference.
+        In these situations `Config.save()`  can be used to avoid having to re-assign the same object reference.
+        This method is also needed when editing Config in 'batch mode' to finalize the state.
         """
         self.__write()
+        self.__is_batch_edit = False
+
+    def __set_and_write(self, property_name, new_value):
```
sharkwouter, line 75: Maybe a return and log upon setting the property_name to None would be good to have. I don't think that is valid.

minigalaxy/config.py
```diff
@@ -211,23 +224,21 @@ def remove_ongoing_download(self, download_id):
...
     @property
-    def paused_downloads(self) -> List[int]:
+    def paused_downloads(self) -> dict[str, int]:
         return self.__config.get("paused_downloads", {})
 
     @paused_downloads.setter
-    def paused_downloads(self, new_value: {}) -> None:
-        self.__config["paused_downloads"] = new_value
-        self.__write()
+    def paused_downloads(self, new_value={}) -> None:
```
sharkwouter: I think this is correct?
```suggestion
    def paused_downloads(self, new_value: list) -> None:
```

minigalaxy/ui/preferences.py
```diff
@@ -178,31 +179,40 @@ def __save_install_dir_choice(self) -> bool:
...
+                if self.__save_install_dir_choice():
+                    self.download_manager.cancel_all_downloads()
+                    self.parent.reset_library()
+                else:
+                    self.parent.show_error(_("{} isn't a usable path").format(self.button_file_chooser.get_filename()))
+        except Exception:
+            logger.error("Could not save preferences", exc_info=1)
+            self.config.cancel_batch_edit()
+        finally:
+            self.config.save()
```
sharkwouter, line 219:
```suggestion
            self.config.cancel_batch_edit()
            return
        
        self.config.save()
```
I don't think we have to save if there was an exception.

sharkwouter, line 219:
```suggestion
            self.config.cancel_batch_edit()
            return
        
        self.config.save()
```
I don't think we have to save if there was an exception.

sharkwouter (changes requested):
Hey @GB609, thanks for the PR! I think this is a really nice change to have and it would lower the chance of corruption, nice work. I had a couple of nitpicks, but I do really like this change.

minigalaxy/ui/preferences.py
GB609: Sure, i can also include those.

minigalaxy/config.py
GB609: `paused_downloads` initially started out as list. But then i added the feature that it keeps the current progress just for the display, so it was converted into a dict.

This is more clearly visible with the function `add_paused_download`, where it is used as `dict[str:int]`.

sharkwouter: I think dict is probably specific enough. It was just weird it had {} as type indicator

minigalaxy/config.py
GB609: Now that you mention it. The idea here was to get the current value and see if `new_value` is different before blindly applying it.

I overlooked the possibility of new_value being `None` and still different from the current value.

I'm actually in favor of adding a special handling of new_value=None in that the configuration should be deleted from the dict. This can be useful to enforce the getter of the property to return the default value again the next time it is used.

minigalaxy/config.py
GB609: okay, sure

GB609: I changed from a type indicator to a default initialisation here.

But it seems i have to rework the `add` and `remove` functions for ongoing downloads and paused downloads either way. The change comparison won't work as intended here. I need to make a working copy of the current state first.

minigalaxy/ui/preferences.py
GB609: You are right. I just figured it wouldn't matter too much as the cancel reverts changes. And returning from a finally block produces a python warning.

Maybe i can use some kind of success/failure flag

GB609: This seems to be a duplicate of the comment just above this one

GB609:
Ok, i think i've adressed all of your comments with the new commits.
I noticed that github produces quite a few warnings about deprecated actions. 
We might need to update the actions we use soon, but that's something for another PR and day.

minigalaxy/ui/preferences.py
GB609: I've added a flag and removed the finally block.
The return would have been problematic either way because it would have cut the method short of calling `self.destroy`

minigalaxy/config.py
GB609: I added a warning log for setting `None` to a property and also some code to delete the corresponding entry instead of saving `None`

minigalaxy/config.py
GB609: I simplified the type hints to `list` (`current_downloads`) and `dict` (`paused_downloads`) respectively. 

I also set up the getters to return copies instead of the internal structures. That required me to also change another piece of code that relied on the side-effects to change config (plus the corresponding test).

sharkwouter (approved):
Looks good now, thanks :D

PR #735, opened 2026-05-06
introduce dedicated key enum for Game.[get|set]_info

GB609:
## Description

This is another one of my on-to-go refactorings. I want to introduce a proper Enum for the keys of the data pieces that get read/written from/to `Game.get_info` and `Game.set_info`.
That enum only exists on python level. Any access to json will be mapped to the corresponding string key first.

Reason:
These values often get read and written across multiple files, it's pretty easy to lose track of what kind of info there is and how exactly the key was spelled when only hard-coded strings are used.
Introducing an enum helps mitigate this as type hints can be used in the IDE for a direct documentation and there's a list of options to chose from.

I need to introduce a few more keys and i'm starting to get issues with keeping track of which ones and where, especially since i need to use them in more than just 2 or 3 places.

## Checklist
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 **1.4.3**
+- Some code refactoring for more clarity and to avoid potential coding errors. (thanks to GB609)
 - Reduce unnecessary writes to the config file. (thanks to GB609)
 
 **1.4.2**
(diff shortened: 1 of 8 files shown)


sharkwouter (changes requested):
This is a great change, cool :D

I do think we should probably not include it in the changelog, since it doesn't affect the users.

GB609:
Sure, i'll take it out

sharkwouter (approved):

sharkwouter:
Thanks!

Issue #404, opened 2021-11-08
Make Wine and Innoextract dependencies


sharkwouter:
Please leave feedback about this idea if you have any, but I am starting to think we should require wine and Innnoextract to be installed. It would make it easier to rely on them and most GOG users have some Windows exclusive games in their library. I'd like to no longer see bugs stating that some games are not showing.

Changes:

- [ ] Make Innoextract and wine dependencies
- [ ] See if some code can be cleaned up
- [ ] Change the "Show Windows games" to "Hide Windows games"
- [ ] Make sure Windows games are shown by default

f35f0ef9d0e827dae86552d3899f78fc:
While it would work fine and dandy for me, I have to assume there are Linux purists that will not install Wine. My two cents.

sharkwouter:
That is possible. We do have a lot of wine related code and not showing Windows games confuses people I've found. Alternatively, it would be possible to show windows games and give an error message, but that's also a bit weird.

GB609:
I wouldn't add innoextract and wine as dependencies.
Instead, as default, i would show all games, but with greyed out/disabled controls when there is no wine installation. Then add a tooltip to these games saying something like "requires wine". Plus a system settings that lets change this to full hiding.

sharkwouter:
@GB609 that makes sense. Would improve usability over what we have now. I would close this issue with that change.

sharkwouter:
Cool, take your time. I'm looking forward to your contribution :)

PR #736, opened 2026-05-06
Feature: Prepare new preferences for platform-selection

GB609:
## Description

This is a very small PR which only adds a couple of new preferences in `Config` which i'll be needing later.
There is no UI for them yet, it will follow after the implementation of the free platform selection feature.

### Look-ahead about preferences and how i plan to use them:

I plan to replace the current `show_windows_games` with a more dynamic variant consisting of 2 other properties:
  1. Users can specify one `preferred_platform` (which effectively is what we have hard-coded as 'linux' everywhere)
  2. The combination of `preferred_platform` and `game_listing_mode` will control which games are shown and which are hidden. The `preferred_platform` generally is the one that is always shown. `game_listing_mode` defines if and how other platforms are shown.
- The combination of `preferred_platform=linux` and `game_listing_mode=all` would correspond to the current behavior when `show_windows_games=True`
- `preferred_platform=linux` + `game_listing_mode=preferred` => The current default where only linux games are visible
- `game_list_mode=mixed` would be new and correspond to a somewhat older suggestion from https://github.com/sharkwouter/minigalaxy/issues/404#issuecomment-2498998199 but in a more generalised way

`preferred_platform` has 2 more purposes, it's not just used as a library filter:
1. Games that support multiple platforms generally allow a per-game selection, but this can quickly become tedious to click and change if a user generally prefers another platform as what we would deem 'the standard'. `preferred_platform` is used as a default fallback instead of forcing the user to pick every time. What platform is preferred will be placed very prominently in the UI later, so that it's always visible.
2. To declutter icons, but still behave in a way that's similar to how the wine icon is shown currently. Installed games will not show the platform icon, if the installed variant corresponds to the current `preferred_platform`. E.g. one might set `preferred_platform=windows`, then installed windows games will appear without the wine icon and native linux games instead will get a little penguin. A user that decides to primarily install windows games should have the wine icon obstruct almost every game.  
Evaluation of this setting will be dynamic, changing the preferred platfrom and restarting will impact the icons shown on installed games.  
Not installed games supporting several platforms will always show one icon per platform. They will be clickable to allow quick platform selection.

Changelog wasn't touched in this PR and will be updated with the UI changes to prevent unnecessary merge conflicts with the other open PR.

@sharkwouter I'm open to any feedback or ideas you might have about my plans. I can also post a few screenshots

## Checklist
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 **1.4.3**
+- Some code refactoring for more clarity and to avoid potential coding errors. (thanks to GB609)
 - Reduce unnecessary writes to the config file. (thanks to GB609)
 
 **1.4.2**
--- a/THIRD-PARTY-LICENSES.md
+++ b/THIRD-PARTY-LICENSES.md
@@ -1,4 +1,4 @@
-## Logo image (data/minigalaxy.png)
+## Logo image (minigalaxy/ui/data/io.github.sharkwouter.Minigalaxy.png)
 
 **[Copyright 2014 Epic Runes](https://opengameart.org/users/epic-runes)**
 
--- a/minigalaxy/config.py
+++ b/minigalaxy/config.py
@@ -142,6 +142,14 @@ class Config:
         self.__set_and_write("keep_installers", new_value)
 
     @property
+    def keep_installers_per_platform(self) -> bool:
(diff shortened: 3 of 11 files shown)


GB609:
there's something wrong with the commits - the PR is larger than it should be one moment

PR #737, opened 2026-05-06
Feature: Prepare new preferences for platform choice (reopened)

GB609:
(I had to recreate the PR because github didn't let my reopen the first PR (#736 ) after my rebase)

## Description

This is a very small PR which only adds a couple of new preferences in `Config` which i'll be needing later.
There is no UI for them yet, it will follow after the implementation of the free platform selection feature.

### Look-ahead about preferences and how i plan to use them:

I plan to replace the current `show_windows_games` with a more dynamic variant consisting of 2 other properties:
  1. Users can specify one `preferred_platform` (which effectively is what we have hard-coded as 'linux' everywhere)
  2. The combination of `preferred_platform` and `game_listing_mode` will control which games are shown and which are hidden. The `preferred_platform` generally is the one that is always shown. `game_listing_mode` defines if and how other platforms are shown.
- The combination of `preferred_platform=linux` and `game_listing_mode=all` would correspond to the current behavior when `show_windows_games=True`
- `preferred_platform=linux` + `game_listing_mode=preferred` => The current default where only linux games are visible
- `game_list_mode=mixed` would be new and correspond to a somewhat older suggestion from https://github.com/sharkwouter/minigalaxy/issues/404#issuecomment-2498998199 but in a more generalised way

`preferred_platform` has 2 more purposes, it's not just used as a library filter:
1. Games that support multiple platforms generally allow a per-game selection, but this can quickly become tedious to click and change if a user generally prefers another platform as what we would deem 'the standard'. `preferred_platform` is used as a default fallback instead of forcing the user to pick every time. What platform is preferred will be placed very prominently in the UI later, so that it's always visible.
2. To declutter icons, but still behave in a way that's similar to how the wine icon is shown currently. Installed games will not show the platform icon, if the installed variant corresponds to the current `preferred_platform`. E.g. one might set `preferred_platform=windows`, then installed windows games will appear without the wine icon and native linux games instead will get a little penguin. A user that decides to primarily install windows games shouldn't have the wine icon obstruct almost every game's thumbnail.  
Evaluation of this setting will be dynamic, changing the preferred platfrom and restarting will impact the icons shown on installed games.  
Not installed games supporting several platforms will always show one icon per platform. They will be clickable to allow quick platform selection.

`keep_installers_per_platform` is for folks that like to switch and try both variants. Initially i ran into problems with `keep_installers` alone when i switched platforms after downloading the other one. E.g. I had windows downloaded and switched to linux. The install failed miserably because it took the windows files. I've got some download manangement in place now so that this won't happen again - and ``keep_installers_per_platform` controls part of it.

When `True`, the downloads of the previous platform are kept by renaming the game installer directory (which is restored when switching back). When `False`, switching platforms deletes all previously downloaded files to make sure the correct ones for the new current are downloaded.
That means to keep all installers for ALL platforms, both `keep_installers` and `keep_installers_per_platform` need to be `True`. `keep_installers=True` alone will only persist installers until the next effective platform switch for a game. `keep_installers_per_platform` will effectively be ignore when `keep_installers=False`

Changelog wasn't touched in this PR and will be updated with the UI changes.

@sharkwouter I'm open to any feedback or ideas you might have about my plans. I can also post a few screenshots

## Checklist
 
### DIFF
--- a/minigalaxy/config.py
+++ b/minigalaxy/config.py
@@ -142,6 +142,14 @@ class Config:
         self.__set_and_write("keep_installers", new_value)
 
     @property
+    def keep_installers_per_platform(self) -> bool:
+        return self.__config.get("keep_installers_per_platform", True)
+
+    @keep_installers_per_platform.setter
+    def keep_installers_per_platform(self, new_value: bool) -> None:
+        self.__set_and_write("keep_installers_per_platform", new_value)
+
+    @property
     def stay_logged_in(self) -> bool:
         return self.__config.get("stay_logged_in", True)
 
@@ -172,6 +180,22 @@ class Config:
     @show_windows_games.setter
     def show_windows_games(self, new_value: bool) -> None:
         self.__set_and_write("show_windows_games", new_value)
+
+    @property
+    def preferred_platform(self) -> str:
+        return self.__config.get("preferred_platform", "linux")
+
+    @preferred_platform.setter
+    def preferred_platform(self, new_value: str) -> None:
+        self.__set_and_write("preferred_platform", new_value)
+
+    @property
+    def game_listing_mode(self) -> str:
+        return self.__config.get("game_listing_mode", "linux")
+
+    @game_listing_mode.setter
+    def game_listing_mode(self, new_value: str) -> None:
+        self.__set_and_write("game_listing_mode", new_value)
 
     @property
     def keep_window_maximized(self) -> bool:
--- a/minigalaxy/config.py
+++ b/minigalaxy/config.py
@@ -20,16 +20,22 @@ class Config:
         self.__is_batch_edit = False
         self.__config_file = config_file
         self.__config = {}
+        # this property is cached, because its config value representation differs from the structure in code
+        self.__platform_mode = None
         self.__load()
 
     def __load(self) -> None:
-        if os.path.isfile(self.__config_file):
-            with open(self.__config_file, "r", encoding="utf-8") as file:
-                try:
-                    self.__config = json.loads(file.read())
-                except (json.decoder.JSONDecodeError, UnicodeDecodeError):
-                    logger.warning("Reading config.json failed, creating new config file.")
-                    os.remove(self.__config_file)
+        if not os.path.isfile(self.__config_file):
+            return
+
+        with open(self.__config_file, "r", encoding="utf-8") as file:
+            try:
+                self.__config = json.loads(file.read())
+            except (json.decoder.JSONDecodeError, UnicodeDecodeError):
+                logger.warning("Reading config.json failed, creating new config file.")
+                os.remove(self.__config_file)
+        # reset cached transformed properties
+        self.__platform_mode = None
 
     def __write(self) -> None:
         if not os.path.isfile(self.__config_file):
@@ -39,6 +45,13 @@ class Config:
         with open(temp_file, "w", encoding="utf-8") as file:
             file.write(json.dumps(self.__config, ensure_ascii=False))
         os.rename(temp_file, self.__config_file)
+
+    def __as_list(self, value: str):
+        return value.split(",") if str else []
+
+    def __as_csv(self, value):
+        """When value is a list, join it with comma and return, else expect it to be string and just return as is"""
+        return ",".join(value) if isinstance(value, list) else value
 
     def start_batch_edit(self):
         """
@@ -142,14 +155,6 @@ class Config:
         self.__set_and_write("keep_installers", new_value)
 
     @property
-    def keep_installers_per_platform(self) -> bool:
-        return self.__config.get("keep_installers_per_platform", True)
-
-    @keep_installers_per_platform.setter
-    def keep_installers_per_platform(self, new_value: bool) -> None:
-        self.__set_and_write("keep_installers_per_platform", new_value)
-
-    @property
     def stay_logged_in(self) -> bool:
         return self.__config.get("stay_logged_in", True)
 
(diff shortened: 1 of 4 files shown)


GB609:
This is how `preferred_platform=windows` would look like (with `game_listing_mode=all`, because the others are not implemented yet). What's missing is the indicator of the preferred platform. I plan to place the icon of the preferred platform in the header, right next to the 'installed' label.
prefer_windows

This is with `preferred_platform=linux`:
select_windows

The main difference is which icon is shown on installed games (and also which platform's installers would be used by default, but that is not directly visible)

sharkwouter:
I do think how to do the usability of this is really hard. I don't know if this is the way to go either. I don't think there will be many users that will never want to install a Linux version of a game. I would say the user will want to be able to take the following actions:
- Install a Windows version of a game of which they know the Linux version is worse/doesn't work.
- Replace the Linux version of a game which doesn't work with the Windows version.

I think hiding Windows games has been causing confusion for a while now and our wine support is much better now than it used to be. I do think we should stop doing this.

A couple of ways I think of which could make installing Windows versions of games that have Linux versions make sense to the user could be:
- Ask the user if they want to install the Windows version instead when we detect the game a game is not launching. Not sure if this would be on first launch or second launch.
- Allow individual games to be easily installed as Windows version instead. Next to the download button we could have the first option in the menu behind the arrow button "Download Windows version".
- Having a Windows installation mode can make sense as well, though. Would have to be very visible and I don't know if we should keep it on after the user has started a download, we could automatically revert to the default mode. This is somewhat similar to what you're building here and would allow us to use all of the code in the PR.

These are just some ideas, though. What do you think?

GB609:
Let me go through your ideas and input one-by-one...

> I do think how to do the usability of this is really hard. I don't know if this is the way to go either. I don't think there will be many users that will never want to install a Linux version of a game. I would say the user will want to be able to take the following actions:
> 
>     * Install a Windows version of a game of which they know the Linux version is worse/doesn't work.
> 
>     * Replace the Linux version of a game which doesn't work with the Windows version.

My implementation does not disable linux games. In its core, all i'm doing is giving the additional **option** to pick any of the supported platforms **when downloading**.
As an **additional convenience**, i thought of introducing an option to change the "default" instead of expecting "linux" for everyone. 
That does not make linux games go away. It only means that windows versions would be preferred (unless manually overridden per game). It's not an issue of "never", but a matter of which platform is installed "more often". If a user tends to prefer windows in general (linux ports are actually often worse), it would mean a lot of additional clicks (compared to a user prefering linux) for that user because (s)he would have to override the platform choice for every game anew.
Maybe the way it works right now would become clearer if you tested my current state. See the current branch at https://github.com/GB609/minigalaxy/tree/feature/platform-selection (the preferences dialog isn't done yet)

For usability, i'm always trying to keep the current default behavior as much as possible, so that users can just update and go on as if nothing ever happened.

> 
> I think hiding Windows games has been causing confusion for a while now and our wine support is much better now than it used to be. I do think we should stop doing this.

I wouldn't completely drop the hiding feature... there are also tickets from users who just don't want to see windows games or use wine. And since that logic is already there, it's not too difficult to just keep it.
What we _could_ do, is reverse the default: Show all by default (if wine is installed), hide them if enabled otherwise.
I'm open to dropping the idea of "mixed" visibility, with games ONLY supporting the secondary platform being shown greyed out. That would mean that there is no need for `game_listing_mode` as an enum, but we'd have something like `hide_secondary_platform=False|True` as a generic 'inverse' of `show_windows_games=False|True` instead.

> 
> A couple of ways I think of which could make installing Windows versions of games that have Linux versions make sense to the user could be:
> 
>     * Ask the user if they want to install the Windows version instead when we detect the game a game is not launching. Not sure if this would be on first launch or second launch.

This is a very hard thing to do in my opinion, because it is a complex action that requires 
1. Detection of failed launches. Also, it's not always an issue with the games not launching. Some linux ports are simply badly optimized or not maintained, missing patches etc. And what could we do if the user choses to launch the game outside of MG through the desktop file that MG generates?
2. uninstalling
3. downloading (which might take a LOT of time)
4. installing.

We _can_ do this kind of "guided" platform swap as an addition, but i don't think it should be the only way, as it forces users to jump through hoops and download the game a second time before allowing them to install what they really wanted in the first place.

I think it isn't too difficult to let the user decide before actually downloading. The user can always uninstall the current version and then chose to select the other one (this is what was requested in the past - they want to chose manually)

> 
>     * Allow individual games to be easily installed as Windows version instead. Next to the download button we could have the first option in the menu behind the arrow button "Download Windows version".

This is basically what the little icon bar on the game's thumbnail is for in my current implementation. I just didn't want to bloat up the menu any further by adding more options.
Thus, clicking the icon selects the platform to download and install (which is only needed if the desired target differs from the "default"), the download button does the rest.
One more click for the occasional override instead of pulling down the menu and then selecting the alternative install.
 I you prefer to have the options in the game menu instead, i can also change it like this. But that would mean that i have to fiddle with the button labels when the preferred platform changes.

> 
>     * Having a Windows installation mode can make sense as well, though. Would have to be very visible and I don't know if we should keep it on after the user has started a download, we could automatically revert to the default mode. This is somewhat similar to what you're building here and would allow us to use all of the code in the PR.

I'd personally prefer to at least keep the option of defining/changing the "default" platform (=`preferred_platform`) (which you call "windows mode") and at least adjust the "hide" option correspondingly to work with it. Whether we use buttons or the icon bar for platform choice wouldn't make a difference for this PR with the settings, that would be a presentation detail.
The basic idea is that MG allows users to set up a "default" and align the UI along the currently selected default to reduce ui clutter and to allow any user the same flow of clicks and navigation, even if they prefer windows versions in general.

We could revert the "windows install mode" after one install. But then it wouldn't be a "mode", but a complicated way of manually picking another version for one game. The way i've done the "mode" right now is that it defines the default that is picked when the user does NOT manually chose the game version (which is ALWAYS possible when a game supports multiple platforms). What's missing right now is a very obvious mention of the current "mode". There are 2 ideas how to do that:
1. in each game's icon bar (where the update icon is), pre-select the icon of the currently preferred mode
2. show the current "mode" in a very central, always visible place like the top bar.

I actually prefer variant 2, because showing the selection for every game individually will draw a lot of borders around icons where the user has never clicked anything. It will then be hard to distinguish if the user has manually picked, or the default is in place.

minigalaxy/config.py
```diff
@@ -141,6 +141,14 @@ def keep_installers(self) -> bool:
     def keep_installers(self, new_value: bool) -> None:
         self.__set_and_write("keep_installers", new_value)
 
+    @property
+    def keep_installers_per_platform(self) -> bool:
```
sharkwouter: We might want to always do this

sharkwouter (approved):
Hey @GB609 , thanks for the PR! I had one comment, but I'm giving you the okay to merge anyway since config setting ui changes will still allow you to make changes in the config settings if you think those are meaningful.

minigalaxy/config.py
sharkwouter: Feel free to resolve this one and merge. See it more as a note.

GB609: Shouldnt i do the property name change first, as we discussed on discord? The one for changing `preferred_platform to` `platform_support`?

GB609: About this property you mentioned: i made it an option because keeping everything by default can quickly rack up a lot of installers if people try changing a lot of games back and forth.

GB609:
Alright, i'v incorporated the ideas and feedback from our discussion on discord, which has changed the content of this PR a bit more.
I also thought about your remark with the `keep_installers_per_platform` and you are right, we should not need it. MG already provides a button to open the `installer` directory, users are free to view and delete whatever. It's less confusing that way.

I'll merge for now. If there are any more adjustments needed, i'll do them when doing the UI. But i think, converting platform_mode into a priorised list of values will work nicely for visibility and as a platform filter / default

PR #739, opened 2026-05-10
Feature: Introduce `platform_mode` in Preferences and use it to select installers

GB609:
## Description

This PR introduces a partial implementation of the final platform selection feature:
It is now possible to set up globally if linux or windows installers shall be used (applies for games supporting both).
Additionally, pure windows games will be shown or hidden depending on the value of `platform_mode` that was picked.
The code in this PR is complete enough to have a working state in master - the new preference is used. So a basic way to chose exists now. Users going from git directly could use/test the current state without having to wait for a release.

This new setting replaces the previous option `show_windows_games`. Code to migrate the property value was added and will apply when Minigalaxy loads its config.json:
1. when `show_windows_games` exists as a key, take the value and then remove the key
2. if it was `True` before, apply `platform_mode=linux,windows`
3. If it was `False`, apply `platform_mode=linux`

NEW modes:
- "Linux only":`platform_mode=linux`: Show all games supporting a native linux version. Any additional support for windows is ignored and won't even be given as a choice to pick from. This is the _current_ behaviour when `show_windows_games=False`
- "Prefer Linux":`platform_mode=linux,windows`: Show all games supporting at least one of the given platforms.  
   Use/prefer the linux installers where available.
- "Prefer Windows":`platform_mode=windows,linux`: Show all games supporting at least one of the given platforms.  
   Use/prefer the windows installers where available.
- "Windows only":`platform_mode=windows`: The inverse of `platform_mode=linux`. In theory, **pure** linux games would be hidden (if such games exist, although rare). Aside from that, for games that support both, it will later have the effect that the linux version won't be offered for download at all (icon to change is hidden).

The difference between the two "Prefer" variants gets more pronounced and important when selection per game is added as well since it will then impact the default and the icon visibility. For now (without the per-game icons), "Prefer Windows" is effectively the same as "Windows only" because there are literally close to zero **pure** linux games on GOG (if at all).

This PR also includes a few code refactorings to reduce nesting levels and complexity.

**Open bugs**:
1. State
Downloading a game and then changing the preferred platform BEFORE it has finished downloading or before it is installed, will likely utterly confuse MG and have it try to e.g. resume windows-started downloads with linux files (effectively a restart), or try to use wine for shell scripts.
I haven't added extra code to accomodate this situation, as the current state is intermediate.
It works as in the master branch starts ups and is capable of downloading and installing according to the selected preferences. But changing the pref should best be done only when MG is idle. The rest will be dealt with in the next 1-2 PRs.

2. Wine
I have temporarily removed part of the `shutil.which("wine")` check in preferences which could result in an error dialog and the reverting of `show_windows_games`. I plan to instead just make the game's main button disabled when there's no wine (plus add a suitable tooltip)

## Checklist
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 **1.4.3**
+- Introduce a global preference to pick if linux or windows game installers are preferred. (thanks to GB609)
 - Reduce unnecessary writes to the config file. (thanks to GB609)
 
 **1.4.2**
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -78,46 +78,55 @@ class Api:
             refresh_token = ""
         return refresh_token
 
-    # Get all Linux games in the library of the user. Ignore other platforms and movies
     def get_library(self):
+        """Get all Linux games in the library of the user. Ignore movies."""
         err_msg = ""
-        games = []
-        if self.active_token:
-            current_page = 1
-            all_pages_processed = False
-            url = "https://embed.gog.com/account/getFilteredProducts"
-
-            while not all_pages_processed:
-                params = {
-                    'mediaType': 1,  # 1 means game
-                    'page': current_page,
-                }
-                response = self.__request(url, params=params)
-                if "totalPages" not in response:
-                    err_msg = "Couldn't load game library"
-                    return games, err_msg
-                total_pages = response["totalPages"]
-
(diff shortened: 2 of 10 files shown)


GB609:
I'm missing a few chunks of code for a working integration. I'll put the PR to draft for now and come back to it once i've added the missing parts.

PR #741, opened 2026-05-12
Feature: Migrate preferences to `platform_mode` (reworked)

GB609:
## Description

After the issues we found yesterday with the other PR (#739), i decided to take a little step back and don't immediately implement a 'global' way to change the default platform.

Functionally, this PR now ONLY contains the move from `show_windows_games` to using `platform_mode` for game visibility instead. NO other selection capabilities _yet_. And a game's platform is set up as it has been before - if linux is supported, use linux. Try windows otherwise. Resuming downloads and installing is no problem, because any game can only either be linux or windows, no selection. If a Game has platform "windows" it will also be resumed as such, regardless of whether "Linux only" or "Prefer Linux" is picked.
So it's not **more** to fix the selection, it's **less** first. Just enough to make the new combo work instead of the `show_windows_games` switch. The rest WILL still come. Just takes a bit more time.

I added 2 other refactorings in `api.py` and `library.py` where i split up complicated methods into smaller ones for which i also added tests.

Before any more work on platform selection can be done, i might need to take another little detour and work on how downloads are resumed, or at least think some more about how to best integrate platform selection into that.
The way it is done now makes switching (default) platforms very hard because the entire download list for a game is refetched and rebuild from API after every restart (this might also cause download failures when GOG has published a new version before a download is resumed at a later date).

## Checklist
 
### DIFF
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -114,19 +114,19 @@ class Api:
 
             worksOn = product.get("worksOn", {})
             platform = None
-            if worksOn["Linux"]:
+            if worksOn.get("Linux", False):
                 platform = "linux"
-            elif worksOn["Windows"]:
+            elif worksOn.get("Windows", False):
                 platform = "windows"
 
             if not platform:
                 logger.warn("%s has no platform information - skip", product["title"])
                 continue
 
-            if not product["url"]:
+            if not product.get("url", None):
                 logger.warning("%s (%s) has no store page url", product["title"], product['id'])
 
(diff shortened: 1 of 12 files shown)


sharkwouter (commented):
This seems fine, but what I am noticing is that there is no default for the new combobox. When I first opened the settings, it was empty.

sharkwouter:
It could be that I had it on a Windows setting, though. Then it makes sense. Let me do a quick test.

sharkwouter:
Okay, yes, my theory was correct. This works fine when the platform is not set.

PR #746, opened 2026-05-25
Fix exception handling in installation callback

sharkwouter:
## Description

Regular exceptions don't have a message attribute, which was causing exceptions in some of my tests with a change I'm making. With this change printing it to the logs should work again.

## Checklist
 
### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -815,8 +815,8 @@ class InstallTask:
         '''Small proxy method to be passed to install_game for intermediate progress report'''
         try:
             self.callback(InstallResult(self.installer_id, result_type, reason, details))
-        except Exception as e:
-            logger.error("Installation callback handler threw an error: %s", e.message, exc_info=1)
+        except Exception:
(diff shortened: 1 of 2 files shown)


minigalaxy/installer.py
```diff
@@ -815,8 +815,8 @@ def notifyStep(self, result_type: InstallResultType, reason='', details=None):
         '''Small proxy method to be passed to install_game for intermediate progress report'''
         try:
             self.callback(InstallResult(self.installer_id, result_type, reason, details))
-        except Exception as e:
-            logger.error("Installation callback handler threw an error: %s", e.message, exc_info=1)
+        except Exception:
+            logger.error("Installation callback handler threw an error: %s", exc_info=1)
```
GB609: Doesn't this line now miss a parameter for the string arg `%s`? We could either remove it as well or use `str(e)`

sharkwouter: Fair, my bad. Using `str(e)` won't be necessary, because the full trace and message are already included in the log because of `exc_info=1`.

PR #747, opened 2026-05-25
Improve logging

sharkwouter:
## Description

This makes us use the logging.funtion instead of having to import our custom logger, while still having control over it. This makes development a bit easier. I also replaced almost all print statements and made the logger write to a file (`~/.cache/minigalaxy/minigalaxy.log`). Please squash when merging.

## Checklist
 

minigalaxy/minigalaxy.py
```diff
@@ -53,12 +54,42 @@ def show_installer_notification(installer_item):
     popup.show()
 
 
+def setup_logging():
+    from minigalaxy.paths import LOG_FILE_PATH
+
+    log_directory = os.path.dirname(LOG_FILE_PATH)
```
GB609: I'd suggest to do the path initialisation/directory creation in the paths.py file, like it's done for cache too to keep things consistent.

GB609 (approved):
I've just got one small suggestion. Looks good otherwise.

minigalaxy/minigalaxy.py
sharkwouter: Fair, will do.

sharkwouter: Done

PR #749, opened 2026-05-25
Add selection launch option

### LINKED ISSUE #532: Regression: Wine launcher from current master no longer finds the correct executable for some games
After 571eeb6 and 44510f4 were applied, executables for some of my currently installed games are no longer detected correctly (though I'm pretty sure the changes from 44510f4 didn't really affect anything as it only affects games with no primary tasks… which probably don't actually exist). I believe most of them are some of the older GOG installers, as their tasks don't have a `category` field; due to which the app no longer can find the main executable, and tries to run whatever (which has no chance of working if the file it _should_ run is in a subfolder… which is always the case for DOSBox/ScummVM games running via Wine). Other games have a launcher as the only primary task, which is rejected by this check and thus fallback logic is employed.

Affected games (in my current install list):
* **Aliens versus Predator Classic 2000** (runs _language setup_ instead of the game)
* **Betrayal at Krondor** (technically not tested but the fallback logic should fail as it's a _DOSBox game_)
* **Betrayal in Antara** (fallback logic finds nothing because _game executable is in uppercase_)
* **Deus** (technically not tested but the fallback logic should fail as it's a _DOSBox game_)
* **Diablo** (has _no_ primary `game` task due to being a 2-game pack; since `launcher` type is ignored, it tries to use fallback logic but _all correct executables are in subfolders_ – so instead it finds the _old game executable_ which demands a CD)
* **Die By the Sword** (same here – the `game` task is not primary due to a launcher existing, and the fallback logic picks _the video settings tool_ instead)
* **Disciples I** (fallback logic finds nothing because _all executables are in subfolders_… also the game task is a shortcut file, `.lnk`)
* **Evil Islands** (runs _the video settings tool_)
* **Hob** (appears to try running _the crash reporter_… which isn't meant to be run this way so it fails silently instead)
* **Pilot Brothers II**/**III** (fallback logic finds nothing because _the executable is in a subfolder_)
* **Quake I: The Offering** (runs _the video settings tool_)
* **Rage of Mages I**/**II** (tries to run _map editor_… but fails due to its 16-bit color profile)
* **Stonekeep** (technically not tested but the fallback logic should fail as it's a _DOSBox game_)
* **The Elder Scroll Adventures: Redguard** (fallback logic finds nothing because _all executables are in subfolders_ …also it's a _DOSBox game_ in the first place)
* **Wizardry VI** (technically not tested but the fallback logic should fail as it's a _DOSBox game_)

These are also affected but still work due to fallback logic stumbling upon the correct executable (either because it's the only one in game folder or purely by chance):
* **Amid Evil** (finds launcher)
* **Avernum** (finds the game)
* **Blades of Avernum** (finds the game)
* **Driftland** (finds launcher)
* **Might and Magic VI** (finds the game)
* **Neighbours Back from Hell** (finds launcher)
* **Prince of Persia II: Warrior Within** (finds launcher)
* **The Elder Scrolls III: Morrowind** (finds launcher)
* **The Elder Scrolls IV: Oblivion** (finds the game… instead of launcher)
* **Treasure Adventure World** (finds launcher)
* **Warcraft II** (finds the game… instead of launcher)
* **Wizardry VII** (finds the game… instead of launcher)

***
<details><summary><strong><code>.info</code>-file for AvP (no <code>category</code>)</strong></summary>

```json
{
    "clientId": "47282734206214863", 
    "gameId": "1207665883", 
    "language": "English", 
    "name": "Aliens vs Predator Classic 2000", 
    "playTasks": [
        {
            "isPrimary": true, 
            "name": "Aliens vs Predator Classic 2000", 
            "path": "AvP_Classic.exe", 
            "type": "FileTask"
        }, 
        {
            "name": "Language Setup", 
            "path": "language_setup.exe", 
            "type": "FileTask"
        }, 
        {
            "link": "http://www.gog.com/support/aliens_versus_predator_classic_2000", 
            "name": "Support", 
            "type": "URLTask"
        }, 
        {
            "link": "http://www.gog.com/forum/aliens_versus_predator_classic_2000", 
            "name": "Send multiplayer or launcher feedback", 
            "type": "URLTask"
        }
    ], 
    "rootGameId": "1207665883", 
    "version": 1
}
```
</details>

<details><summary><strong><code>.info</code>-file for Hob (<code>launcher</code> as the primary task)</strong></summary>

Note: in many cases the game entry(-ies) included `"isHidden": true`.
```json
{
    "buildId": "50707556912620595", 
    "clientId": "49842418730375899", 
    "gameId": "1300281766", 
    "language": "English", 
    "languages": [
        "en"
    ], 
    "name": "Hob", 
    "playTasks": [
        {
            "arguments": "skip_file_check", 
            "category": "launcher", 
            "isPrimary": true, 
            "languages": [
                "*"
            ], 
            "name": "Play Hob", 
            "path": "hoblauncher.exe", 
            "type": "FileTask"
        }, 
        {
            "arguments": "skip_file_check", 
            "category": "game", 
            "languages": [
                "*"
            ], 
            "name": "Play Hob (no launcher)", 
            "path": "hob.exe", 
            "type": "FileTask"
        }, 
        {
            "languages": [
                "*"
            ], 
            "link": "https://support.runicgames.com", 
            "name": "Support", 
            "type": "URLTask"
        }
    ], 
    "rootGameId": "1300281766", 
    "version": 1
}
```
</details>

***
Suggested fix (for games without `category`): either use `task.get('category', 'game')` in the loop check instead of skipping primary tasks without a category, or do a second run if the first one doesn't find anything.

In case of games that have a `launcher` as their primary task, you can opt to either detect it as well, or search for games among non-primary tasks (which is different from what was removed in 44510f4 as there _are_ primary tasks in the list but none of them are `game`s).

Also (or instead), there should probably be a task/executable selector, because for instance Diablo is a 2-game pack (with a launcher)… and I'd say being able to pick/set temporarily secondary executables like external video settings tool would be nice as well.

<details><summary><em>P.S. Incidentally…</em></summary>

This:
```python
                if "playTasks" in info:
                    for task in info["playTasks"]:
                        if "isPrimary" not in task or not task["isPrimary"]:
                            continue
                        …
```
can be simplified like this:
```python
                primary_tasks = [task for task in info.get('playTasks', [])  if task.get('isPrimary')]
                for task in primary_tasks:
                    …
```
(and the logic removed in 44510f4 would be `for task in primary_tasks or info.get('playTasks', []):`)
</details>

sharkwouter:
## Description

This adds a dialog that will pop up if multiple binaries are found for a Windows game that does not have a launch shortcut or gameinfo file instead of just launching the first binary it finds.

In the future this could still be extended, for now it is very limited, but it does solve a problem.

## Checklist
 

minigalaxy/installer.py
```diff
@@ -335,9 +335,9 @@ def get_exec_line(game):
     See https://specifications.freedesktop.org/desktop-entry-spec/latest/exec-variables.html for details"""
     chars_to_quote = re.compile(r'(["$`\\])', re.ASCII)  # double-qoute, dollar and backtick
     replace_pattern = r'\\\1'
-    exe_cmd_list = get_execute_command(game)
-    for i in range(len(exe_cmd_list)):
-        entry = exe_cmd_list[i]
+    launch_command_list = get_execute_commands(game)
```
GB609: I'd suggest to assign `launch_command_list = get_execute_commands(game)[0]` directly instead of dragging the `[0]` along (you can also make it a 2-step assignment of you don't want to do both at once). Same effect but easier to read.

Maybe also rename `launch_command_list` a bit because it is not a list of launch commands that is relevant here, but the __actual__ command to use, represented as a list of parameters. So something like `command_words` or `launch_command_arguments` might be a better fit.

There's a todo here now as you introduced a slight situational regression:
You have removed the game-specific filters for undesirable executables in `constants.py`.
That means, that installing Diablo 2 or FreeSpace 2 with the option "Create desktop shortcut" enabled might now end up with the wrong executable in the `*.desktop` file (again).

Right now, it's a specific situation for very few users. But we shouldn't forget about the feature in general.
I personally tend to use the desktop shortcuts **a lot**. They are great for integration with other front-ends like EmulationStation etc.

In the long run, i see a new game info property here. Something like `default_launch_command`. And everytime that is changed, the `*.desktop` shortcut would have to be regenerated (game properties windows already triggers a recreate under some conditions).

minigalaxy/launcher.py
```diff
@@ -64,30 +67,35 @@ def start_game(game):
...
         raise FileNotFoundError()
-    if game.get_info(InfoKey.GAMEMODE) is True:
-        exe_cmd.insert(0, "gamemoderun")
-    if game.get_info(InfoKey.MANGOHUD) is True:
-        exe_cmd.insert(0, "mangohud")
-        exe_cmd.insert(1, "--dlsym")
-    exe_cmd = get_exe_cmd_with_var_command(game, exe_cmd)
-    logging.info("Launch command for %s: %s", game.name, " ".join(exe_cmd))
-    return exe_cmd
+    for launch_command in launch_commands:
```
GB609, line 88: Introducing a class for launch commands is a great idea.
I think we can leverage that further and simplify code here.

Gamemode, Mangohud and the variables + arguments from `append_user_defined_variables_to_launch_commands` and `set_fps_display` all feed from game info and manipulate the final launch command.

I think we could do that in a member function of `LaunchCommand` instead.
This would also make testing of launch configuration a bit easier as it allows to test more of the command composition outside of the complex launch call chain.

```python
class LaunchCommand:
    # either manipulate self.command, or return a new [] with all modifications
    # whatever makes more sense to you
    def apply_game_launch_config(game: Game):
        if game.get_info(InfoKey.GAMEMODE) is True:
            self.command.insert(0, "gamemoderun")
        if game.get_info(InfoKey.MANGOHUD) is True:
            self.command.insert(0, "mangohud")
            self.command.insert(1, "--dlsym")

        var_list = shlex.split(game.get_info(InfoKey.VARIABLES))
        command_list = shlex.split(game.get_info(InfoKey.COMMAND))

        # to replace the `set_fps_display` function and stop it from polluting os.environ
        # this also makes it compatible with generating desktop files
        if game.get_info(InfoKey.SHOW_FPS):
            var_list.extend(
                # For Nvidia users + OpenGL/Vulkan games
                "__GL_SHOW_GRAPHICS_OSD=1",
                # For AMDGPU users + OpenGL games
                "GALLIUM_HUD=simple,fps",
                # For AMDGPU users + Vulkan games
                "VK_INSTANCE_LAYERS=VK_LAYER_MESA_overlay"
            )

        if var_list:
            if var_list[0] not in ["env"]:
                var_list.insert(0, "env")
            if 'env' == launch_command.command[0]:
                self.command = launch_command.command[1:]

    # either modify own data or return adjusted/effective value
    self.command = var_list + launch_command.command + command_list
```

I also think there's no real need to do this for ALL found commands, it should be enough to run it once for the final choice in `run_game_subprocess`.

minigalaxy/launcher.py
```diff
@@ -47,13 +48,15 @@ def winetricks_game(game):
...
 
 
-def start_game(game):
+def start_game(game, execute_command: LaunchCommand) -> str:
     error_message = ""
     process = None
+    if not execute_command:
+        error_message = "Cannot launch game, because no command to execute was specified"
     if not error_message:
         error_message = set_fps_display(game)
```
GB609, line 57: I think we can drop this function call and move/merge its logical content to a method in `LaunchCommand` that deals with game-specific launch configs. See the detailed comment i've placed in `get_execute_commands`

minigalaxy/launcher.py
```diff
@@ -145,42 +154,59 @@ def get_windows_exe_cmd_from_goggame_info(game, file: str) -> List[str]:
...
         for file in files:
             if os.path.splitext(file.upper())[-1] not in [".EXE", ".LNK"]:
                 continue
             if file in BINARY_NAMES_TO_IGNORE:
                 continue
-            executable = file
-            break
-        exe_cmd = [get_wine_path(game), os.path.join(game.install_dir, executable)]
+            executables_found.append(file)
+        for executable in executables_found:
```
GB609: Why are you splitting this into a second iteration?
Since you run the second loop at the same nesting level, there is no real benefit to it.

minigalaxy/ui/library_entry.py
```diff
@@ -126,6 +127,25 @@ def run_primary_action(self, widget) -> None:
...
 
+    def launch_game(self) -> str:
+        error_message = ""
+        launch_commands = get_execute_commands(game=self.game)
+        if not launch_commands:
+            error_message = f"Failed to get execute command for game {self.game.name}"
+        elif len(launch_commands) == 1:
+            error_message = start_game(game=self.game, execute_command=launch_commands[0])
+        else:
+            dialog = ChooseLaunchOption(parent=self.parent_window, launch_command_list=launch_commands)
```
GB609: I suggest to make a `@staticmethod` in `ChooseLaunchOption` that encapsulates the steps you are doing here and have it return `launch_command` or `None`. Eg.

```python
ChooseLaunchOption
    @staticmethod
    def ask_for_launch_command(parent=self.parent_window, options: list[LaunchCommand]):
        dialog = ChooseLaunchOption(parent, options)
        dialog.run()
        launch_command = dialog.selection
        dialog.destroy()

        return launch_command
```

Then use that here:
```python
def launch_game():
   # ...
   launch_command = ChooseLaunchOption.ask_for_launch_command(...)
   if not launch_command:
      return "No launch option selected"
   return start_game(game=self.game, execute_command=launch_command)
```

minigalaxy/ui/chooselauchoption.py
```diff
@@ -0,0 +1,47 @@
...
+@Gtk.Template(string=load_ui("chooselaunchoption.ui"))
+class ChooseLaunchOption(Gtk.Dialog):
+    __gtype_name__ = "ChooseLaunchOption"
+
+    choices_combobox = Gtk.Template.Child()
+
+    def __init__(self, parent, launch_command_list: list[LaunchCommand]):
+        Gtk.Dialog.__init__(self, title=_("Choose Launch Option"), parent=parent, modal=True)
+        self.launch_command_list = launch_command_list
+        self.cancelled = True
```
GB609: Is there really a need to differentiate cancel from a `None` result?
They way it is set up, it isn't really possible to click ok without having a selection set:
 - the list of choices fed into the combo is the one searched for the selected LaunchCommand.
 - The lists are identical and there is no dedicated "empty" choice.
 - In combination with the preselection done by `populate_combobox`, the ok button handler is guaranteed to locate a valid choice.

minigalaxy/ui/library_entry.py
```diff
@@ -126,6 +127,25 @@ def run_primary_action(self, widget) -> None:
...
+            dialog.run()
+            launch_command = dialog.selection
+            dialog_cancelled = dialog.cancelled
+            dialog.destroy()
+            if launch_command:
```
GB609, line 140: I think you have set up the dialog with a pre-selected default. So in theory, it shouldn't be possible to click OK and NOT have a launch_command selected.

As such, getting `None` already means cancel (is only logically possible when cancel is clicked, because ok is sure to set a not `None` value).

minigalaxy/launcher.py
```diff
@@ -234,10 +278,10 @@ def set_fps_display(game):
     return error_message
 
 
-def run_game_subprocess(game):
+def run_game_subprocess(game, launch_command: LaunchCommand) -> tuple[str, subprocess.Popen]:
     try:
         process = subprocess.Popen(
-            get_execute_command(game),
+            launch_command.command,
```
GB609, line 264: If you follow my suggestions about `LaunchCommand.apply_game_launch_config`, it would have to be used here instead (instead of applying it in a loop for all LaunchCommand instances).

GB609 (commented):
I think we are off to a good start with this change.

Still, there are a few areas where i think we can refine it a bit more before merging. It's generally related to the fact that you have introduced a new class `LaunchCommand` (which is good!) but don't use it effectively yet (imho).

I've also got 2 small coding / style nitpicks.

minigalaxy/ui/library_entry.py
sharkwouter: I think it makes sense to handle the case where the combobox contains no values here as well, even if it is really unlikely.

minigalaxy/ui/library_entry.py
sharkwouter: This is a good idea, I'll look into this.

minigalaxy/launcher.py
sharkwouter: That is a good idea, I like it.

minigalaxy/ui/library_entry.py
GB609: I mean its not possible from the code itself. No value means there was not a single launch command in the first place. Wouldnt it be an error to pass an empty list?

sharkwouter: Ah, I see what you mean. We already know for sure that there are launch options which we passed to the dialog.

GB609: Yes. The dialog is only useful when there are at least 2 options. And one of them is guaranteed to be selected by populate_combobox. Getting a none result can only happen for the cancel button.

As such, the value of `cancellation` is currently always identical to `launch_command is None`

minigalaxy/launcher.py
GB609: Just a quick heads-up: i had a typo in my code example for SHOW_FPS, it was missing a piece at `GALLIUM_HUD=simple`, the `,fps` somehow was lost when copy-pasting. Fixed that now.

minigalaxy/launcher.py
sharkwouter: Fair, I think I was a bit tired when I coded this. I fixed this now.

minigalaxy/ui/library_entry.py
sharkwouter: I did this now, but slightly differently. Not selecting a launch option means the user closed the window, which isn't really a scenario where an error should be shown.

minigalaxy/ui/chooselauchoption.py
sharkwouter: You are right, I changed this now.

minigalaxy/launcher.py
sharkwouter: Good idea

minigalaxy/installer.py
sharkwouter: I did the fix for this line, but I don't have an answer for the regression yet.

sharkwouter:
I addressed some of the comments now. I'll get back to the other ones later.

LeXofLeviafan:
…Alright, so my first piece of feedback is: that dialog is _really tiny_. I believe it'd look much better with some extra spacing added & with radiobuttons rather than a dropdown (maybe within some optional scrollbox if you're worried about it becoming too large in some cases) :thinking:
selection dialog

It'd also be nice to have it _memorize the last selected option as default_ (always or with an added "launch & set as default" button?), instead of needing to operate said dialog every time (+ to have an extra command/button for _invoking it forcibly_; perhaps within that drop-down menu next to Play button?). …Incidentally, Steam client has similar behaviour implemented for games with multiple launch options.

> for a Windows game that does not have a launch shortcut or gameinfo file

Honestly I'd say there should be an option to circumvent this restriction. In fact, I'd prefer that _when using the "forced invocation" I've suggested above_, the behaviour for game shortcut detection would be different: the gameinfo file in such a case should produce _all_ entries instead of just the primary one, and the `Launch *.lnk` file should _not_ prevent the other EXE/LNK files from being listed (…just don't forget to skip the already added option when iterating over the executables :sweat_smile:)

GB609:
@LeXofLeviafan 
Thanks for the feedback. We are aware of the current restrictions but going at it iteratively.

First is enabling the selection possibility itself. Next will be refining it.

The game info file is not a file made or managed by Minigalaxy. It comes from GOG and either exists or doesn't... and when it doesn't, it's a problem because that file contains the launch commands.
Everything we do - for now - is trying to solve the problem that there is NO info from GOG. There is little that Minigalaxy can do then aside from full-width search and asking the user. This is what we introduce for now. Adding a selection for the launch commands from GOG is a second step which can then use the framework we're establishing here.

minigalaxy/launcher.py
sharkwouter: I've implemented this. I did not add the fps showing code to this part, because it modifies the env, not the launch command.

minigalaxy/launcher.py
sharkwouter: I think I'd prefer to have this check here for now even if it never triggers. That could prevent a hard to debug bug in the future after code changes.

minigalaxy/launcher.py
GB609: > I did not add the fps showing code to this part, because it modifies the env, not the launch command.

This is exactly why i was suggesting to add it there and convert it to be a part of the game command. I think setting any game-specific config as env in the parent process itself is error-prone and a malpractice in general.

- The way it's done here via setting in the env of MG prevents the user from setting a default for it in the executing shell. MG just overrides indiscriminately either with a predefined value (would be ok) or with an empty string (when the config is disabled, when is not OK, imho. When MG is not set to enable FPS, the variable shouldnt be touched at all in the case it exists).

- There is aways the potential danger of cross-contamination when one game gets launched, closed and then another one is started instead. It's likely that setting the variable to "" in the else case was made to prevent that cross-contamination.

- Also, when you add it to the command, the settings apply to the desktop file too - you get the same behavior. When only setting into env when MG runs, the game launched via desktop file (created by MG) will behave differently.

minigalaxy/launcher.py
GB609: fine by me

minigalaxy/launcher.py
LeXofLeviafan: > There is aways the potential danger of cross-contamination when one game gets launched, closed and then another one is started instead

If you mean invoking separate subprocesses, both the `env={}` param of [`subprocess`](https://docs.python.org/3/library/subprocess.html) & the [`env`](https://www.man7.org/linux/man-pages/man1/env.1.html) shell command prevent that entirely (and if it's the same subprocess, inheriting the same env should be the expected behaviour) :thinking:

> Also, when you add it to the command, the settings apply to the desktop file too - you get the same behavior. When only setting into env when MG runs, the game launched via desktop file (created by MG) will behave differently.

AFAIK desktop files do not support any way to have dynamic envs (without updating the file itself) nor to prevent shadowing existing vars; unless you invoke the game via a wrapper script implementing that, anyway. (…Technically you could 'bake' such functionality into the desktop file by wrapping the invocation in a Bash call, but that approach sounds both awkward and brittle :sweat_smile:)

minigalaxy/installer.py
GB609: About the regression. We discussed this on discord already. Just leave it be for now, the impact should be very small. And it'll be gone once we allow to set a default executable via game properties.

minigalaxy/launcher.py
GB609: @LeXofLeviafan 
Minigalaxy does not use a dedicated env dictionary. It sets some of the variables to `os.environ`, which is a **global** object. 

Even env does inherit anything it doesn't overwrite, unless it is invoked with the option `-i | --ignore-environment` (which MG doesn't use, it's also rarily useful)

Generally, with a few exceptional variables, MG uses `env` and prefixes the game command with it. It does not use an env dictionary when spawning processes.

About desktop files: 
- MG generates them with the exec key in the form `env VAR=value ... real command ...`. There is no need to bake or hack anything.
- The command is collected and build basically the same way it would be for launching from within MG (except the problematic env vars above). It's just stringified and written to file instead of being passed to subprocess.
- They already contain envs defined that way today. This is most notable with desktop files for windows games which contain the `wineprefix`.
- Desktop files are re-written (updated) whenever a property in a game's property dialog is changed. That is all the dynamic they need.

GB609 (approved):
I think we can have a go with that for now and optimize it some more with follow-ups

PR #751, opened 2026-05-28
Platform selection: Prepare UI elements

GB609:
## Description

This is the next part for the greater feature that'll enable users to select which game variant/platform they want to install.

I've focussed on the UI in this PR.

1. New icon for linux. While i added it to the `THIRD-PARTY-LICENSES.MD` i saw that another icon file mentioned there has been moved and/or renamed. I've corrected that as well.
2. The icon display over a Game's thumbnail has been streamlined. There's now a new custom widget, `GameIconBar`. Rendering and visibility of icons is decided by this class.
3. The `GameIconBar` also contains the 'update available' icon now. Reasons:  
    - Visibility: All icons share a common, semi-transparent background to make them more visible independently of the thumbnail's colors.
    - Simplification of code. Before the change, `LibraryEntry` did some positioning math when a game had an update available AND had the wine icon visible to shift the wine-icon to the right by a hard-coded amount of pixels. That is not necessary anymore. All icons are in one box which is set up to be left-oriented. We only need to hide or show icons as needed, the rest of the layout is then done by GTK.
4. There's now a possibility to cache icons we load. I've found this necessary considering that we load the same to icons for wine and/or linux potentially hundreds of times. Why load them from the file system for every new gametile as a new asset, when they can just be shared and reused?  
  This addition is accompanied by a new test file.
5. The PR already contains some (inactive) code (and css) to use the icon bar as ui for platform selection for installs. It's about reaction to mouse events for hover and click to highlight and save choices.

Small changes in behaviour:
- I have dropped the '(Wine)' attachment to the game thumbnails quickinfo and game title. The platform is symbolised by a wine-icon already, why manipulate texts as well?
- At least temporary, ALL games will show their (current) platform. There is no selection yet, the preselection as done by `library` is still in place.

Here's a little snapshot of the current UI state:
vokoscreenNG-2026-05-28_14-39-18 406

## Checklist
 
Note: Final changelog will follow when i'm done.


sharkwouter (approved):
Works well and looks good. Even seems to improve some of the jumping around.

PR #754, opened 2026-06-13
chore: resolve warnings and test output noise

slowsage:
## Description

Resolves deprecation/syntax warnings and quiets test output noise - easier to track if anything is actually wrong. No behavioral change; 177 tests pass.
Follow up to #731.
Arch PKGBUILD issue tracked [here](https://github.com/FabioLolix/PKGBUILD-AUR_fix/pull/91).

Warnings:
- `api.py`: use `logging.warning` instead of deprecated `logging.warn`.
- `api.py`: compare XML Element with `is not None` rather than its deprecated boolean truth-value.
- `api.py`: move `return result` out of the `finally` block (SyntaxWarning) and catch `ET.ParseError` alongside `RequestException`, so malformed XML yields an empty result instead of propagating.
- `library`/`window`/`preferences`/`translation`: replace deprecated `locale.getdefaultlocale()` with `locale.getlocale()`; the process locale is already initialized by `translation.py` at import time.

Test output noise:
- `test_installer` mocked Popen `stdout.readlines` (plural), but `_exe_cmd` reads `readline` (singular), so `readline` returned a live MagicMock that was truthy yet had `__len__ == 0` — the read loop only terminated by accident, and `extract_by_wine` (`print_output=True`) printed the mock to stdout. Stub `readline` on both streams to return `""` for deterministic termination and clean output.
- `test_api` 404 paths set `status_code` but not `text`, so `__get_xml_checksum` logged `response.text` as a MagicMock repr. Stub `text` to `""`.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Logs are now also written to `~/.cache/minigalaxy/minigalaxy.log`.
 - Ask the user which binary to launch if Minigalaxy is unsure.
 - The `wine` executable can now also be changed before a game is installed (thanks to GB609).
+- Resolve deprecation/syntax warnings and quiet test output noise. (thanks to slowsage)
 
 **1.4.2**
 - Switch to using PEP 302 compliant resources, excluding translations. (thanks to EmperorArthur)
--- a/minigalaxy/api.py
+++ b/minigalaxy/api.py
@@ -118,7 +118,7 @@ class Api:
                 platform = "windows"
 
             if not platform:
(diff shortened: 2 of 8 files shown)


sharkwouter:
Hey @slowsage, thanks for the PR! I  just want to let you know that I have seen this and I will review it, hopefully today, otherwise within this week.

sharkwouter (approved):

sharkwouter:
Thanks for the contribution!

PR #755, opened 2026-06-22
Remember last chosen launch option

sharkwouter:
## Description

This makes the launch option dialog remember which option was chosen last and automatically select that option in the dropdown.

## Checklist
 

GB609 (approved):
looks good to me

PR #761, opened 2026-07-06
Target Ubuntu 26.04

sharkwouter:
## Description

This seems necessary, since Ubuntu 24.04 requires changes which break the ability to build for any modern system at the moment. This is new and I don't know why, but it means we have to switch.

## Checklist
 
### DIFF
--- a/.github/workflows/pythontests.yml
+++ b/.github/workflows/pythontests.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   test:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-26.04
 
     steps:
     - uses: actions/checkout@v6
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,7 +4,7 @@ on: workflow_dispatch
 
 jobs:
   release:
-    runs-on: ubuntu-latest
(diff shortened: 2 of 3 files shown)


GB609:
Will this change or break the package for older versions?

sharkwouter:
This will completely remove support for Ubuntu 24.04. It fixes the build for Ubuntu 26.04 and makes it our default target. It also fixes building on Debian 13, but I'm not 100% sure if the package we ship will work on Debian. The situation we are in right now is that we cannot support both anymore.

sharkwouter:
Even worse, currently the pip package cannot build for any system with a somewhat modern python version. This PR would fix that too.

GB609:
Ok. I'm experimenting with building an AppImage... I've got something locally that works but no way to upload it for a little test run with a larger user base because it's too large for discord (max 10mb < ~25mb) and i didnt want to attach an inofficial build to the release.
If we get this to work we can provide the AppImage for older distributions instead.

GB609:
I think we can go with the change. Not that there's much of a choice anyway.

sharkwouter:
Using an AppImage may or may not work for older distros. We'll have to test that. Otherwise we could look into building multiple packages for multiple distributions.

PR #763, opened 2026-07-06
Feature: Provide AppImage with release

GB609:
## Description

This PR integrates an experimental script to build an AppImage of Minigalaxy into the `release.yml` workflow.

This introduces a new script, `scripts/make-appimage.sh [version]`, which does:
1. Set up a temporary directory (`repo-root/tmp`) and switch into it
2. Build MG wheel
3. Download a pre-packaged python runtime AppImage (Manylinux variant to allow broad distribition support)
4. Generate a temporary recipe folder for the AppImage
5. Patch some files into the AppImage folder (`*.desktop` and `requirements.txt`) to include whats needed for the AppImage to build correctly
6. Run `python-appimage build app [...]`

The new script can basically be started from anywhere. It will save the current directory, then switch to its working dir and switch back at the end.

The script has a couple of requirements, however:
- wget
- python-appimage
- python build

The script is called from the release workflow after the debian package is built. And the AppImage is also added to the `Release` step to publish it.

**Note**: I'm not entirely sure about the path i've added in the `Release` step. Normally, actions run from the root of the repo, but the `../Minigalaxy*deb` line confuses me as that would mean that...
1. debian packages are build one folder level ABOVE the repository.
2. OR the release action performs some kind of folder change?

## Checklist
 
 - [X] _CHANGELOG.md_ was updated (**format**: - Change made (thanks to github_username))

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,7 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install -y debhelper dh-python python3-all help2man devscripts gettext lsb-release xmlstarlet git build-essential pybuild-plugin-pyproject
-        pip install setuptools
+        pip install setuptools python-appimage build
     - uses: actions/checkout@v6
       with:
         ref: master
@@ -26,6 +26,13 @@ jobs:
     - name: Build deb package
       run: |
         dpkg-buildpackage -us -uc
+    - name: Build AppImage
+      uses: kohlerdominik/docker-run-action@v2.1.0
+      with:
+        image: quay.io/pypa/manylinux_2_28_x86_64
+        volumes: ${{ github.workspace }}:/mnt/minigalaxy
+        run: |
+          /mnt/minigalaxy/scripts/make-appimage.sh
     - name: Commit changes
       run: |
         git config --global user.name 'Wouter Wijsman'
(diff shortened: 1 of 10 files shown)


sharkwouter (approved):

GB609:
Please don't merge yet. I'd like to try a few more things to make the image more compatible.

GB609:
I had to rework the AppImage build script almost completely. But it is working now.
See https://github.com/GB609/minigalaxy/releases/tag/2.0-AppImage-Test where i created a test-release in my fork with a slightly modified `release.yml` (mostly commented out a lot of steps related to debian or code updates)

For now, the build script hardcodes a few versions and path, but i think it's a good base to work with.

The appimage itself is build with and against a manylinux docker container. To be precise, against version 2_28. This impacts the compile dependencies of certain system libraries, or rather, the required min versions.

I do plan to generate some kind of version/library info file with `pkg-config` for the most important dependencies, like gtk, girepository etc., but that's not included yet.

Working with (and building) `PythonGObject` is an especially obnoxious part of the build as a lot of system library versions (and the PyGObject version too) have to be carefully curated. 
I tried a large range of versions to find one that works with ALMA 8, as well my local, more recent machine. The biggest issue here was the dependency to `girepository`. Starting with PyGObject > 3.50 they require `girepository-2.0` which just doesnt exist for ALMA 8 (probably the same for other distributions of similar age), and anything above 3.44.2 had another gtk-related dependency the docker container couldn't satisfy.
And as a lower border, it can't go below 3.30 specified in minigalaxy's `pyproject.toml`.

Once i've upgraded my dev PC, i can also try to give it a go with the manylinux target 2_34, which is based on ALMA 9. That should allow AppImage builds with more recent versions of `PyGObject`.

sharkwouter:
Thanks!

Issue #767, opened 2026-07-27
`--dlsym` is the default for mangohud


RoGreat:
I've updated minigalaxy here https://github.com/NixOS/nixpkgs/pull/543574 and included a patch to remove `--dlsym` from it a short while ago. I don't think it hurts having it, just that I don't think it does anything since it's now the default as stated here: https://github.com/flightlessmango/MangoHud#opengl

> OpenGL games may also need dlsym hooking, which is now enabled by default. Set the MANGOHUD_DLSYM env to 0 to disable like MANGOHUD_DLSYM=0 %command% for Steam.`

Mainly did it to not have to worry about command formation with the patched in steam sniper runtime (using `umu-run` in this case) to run linux scripts without issue on NixOS. Might help for this project to if you end up going that route.

GB609:
Good to know, thank you for the input.
I am planning to implement proper umu support in the long run.

About the `dlsym` config: If you want, you can also provide your patch as a pull request instead.
The change to the default behaviour of mangohud is in since version 0.8.0 from Feb. 2025. I think this should be 'old enough' to have broad support across distributions.

PR #768, opened 2026-07-28
Mangohud dlsym enabled by default

RoGreat:
## Description

Removes `--dlysm` since mangohud enables it by default since last year https://github.com/sharkwouter/minigalaxy/issues/767.

## Checklist
 

sharkwouter:
Thanks!

Issue #639, opened 2025-02-08
(Feature request) Proton support


hardBSDk:
The [Kron4ek's fork](https://github.com/Kron4ek/Wine-Builds/releases/tag/proton-9.0-4) allow Proton to work without Steam and it up-to-date.

[Wine GE](https://github.com/GloriousEggroll/wine-ge-custom) was [archived](https://github.com/GloriousEggroll/wine-ge-custom/releases/tag/GE-Proton8-26) to focus on [UMU](https://github.com/Open-Wine-Components/umu-launcher) and Proton-GE.

It will increase the compatibility and improve performance.

GB609:
This goes into the same direction as #288 and #622.
I've started implementing the support for multiple wine versions/variants last year, but got held up by IRL and other projects.

Issue #288, opened 2021-02-12
Adding support for crossover


CTRL-ALT-DEL-92:
Dear minigalaxy developer,
 
This is an amazing project, the application works like a charm and using it was a very good and simple experience. For windows games vanilla wine is a good choise,  but not always the best, sometimes alternatives like proton or Crossover (the latter especially for mac users) could really improve the situation. My question is, could be possible to select which wrapper to use to run a game (wine, proton and especially crossover for mac users)?

sharkwouter:
Thanks for the report, great to hear you're enjoying Minigalaxy. There has been some discussion about something like this in #64 and #38. It would be possible to do this, but it would require work from our side. This probably won't make it into the next release, but I will keep it up for later..

Supporting Mac OS is not currently on our list of things to work on. We only support Linux at the Moment..

CTRL-ALT-DEL-92:
> Thanks for the report, great to hear you're enjoying Minigalaxy. There has been some discussion about something like this in #64 and #38. It would be possible to do this, but it would require work from our side. This probably won't make it into the next release, but I will keep it up for later.
> 
> Supporting Mac OS is not currently on our list of things to work on. We only support Linux at the Moment..

Thank you for your quick response and your kindness, I'm glad you gave a response to my feedback and I trust you'll do your best to improve your amazing work. Changes take time and I'm in no rush, I just want to see this project  grow as it deserves. Have a nice day, good work and good luck!!!

ThracianKnight1907:
Integration with [Wine-GE](https://github.com/GloriousEggroll/wine-ge-custom) could also be considered, as Proton can be (as far as i know) problematic when used outside of Steam.

mechamegane:
To take full advantage of Proton related work to get games to run when vanilla Wine isn't doing it I might suggest using https://github.com/Open-Wine-Components/umu-launcher

GB609:
@mechamegane I'm currently in the process of implementing this in my fork. 
Once done and tested, PR will follow.
When the basis is established, adding crossover might be possible. But i myself won't be able to test it since i'm not using mac, nor do i own a license for crossover.

[...]

PR #769, opened 2026-08-16
Add per-game Steam Proton support through UMU

adamjvr:
## Description

Adds per-game Steam Proton support for Windows GOG games through UMU while preserving the existing Wine behavior as the default.

### What this adds

- Discovers Valve Proton installations from Steam libraries, including additional `libraryfolders.vdf` locations.
- Discovers custom/GE Proton compatibility tools.
- Adds a per-game compatibility selector for System Wine, Custom Wine, and discovered Proton tools.
- Uses the selected Proton build through UMU for both the GOG installer and later game launches.
- Creates Proton prefixes through UMU from the first installer process instead of installing with Wine first or converting/moving prefixes afterward.
- Keeps MiniGalaxy in control of the existing GOG game-library location.
- Explicitly prevents a configured Proton backend from silently falling back to host Wine when Proton or UMU is unavailable.
- Requires UMU Launcher 1.4.4+ for Proton mode and accepts newer compatible installations.
- Detects system/user UMU installations and can provision a checksum-verified upstream UMU 1.4.4 standalone zipapp into MiniGalaxy's XDG data directory without root privileges or distro-specific package-manager logic.
- Shows UMU runtime/version status in game Properties when Proton is selected.
- Leaves System Wine and Custom Wine behavior independent of UMU.

### Validation

- Full regression suite: **221 tests passed**.
- `flake8`: **0 findings**.
- `git diff --check`: clean.
- Python compile check: passed.
- Properties XML parsing: passed.
- GTK template import: passed.
- Live UMU dependency detection tested with `/usr/bin/umu-run` 1.4.4.
- Real isolated managed-UMU provisioning test passed.
- End-to-end Windows game test: **Croc Legend of the Gobbos Classic** installed successfully using Steam Proton Experimental through UMU/Steam Linux Runtime.

### Compatibility model

- System Wine → system Wine, no UMU requirement.
- Custom Wine → selected Wine executable, no UMU requirement.
- Proton → selected Proton + UMU + Steam Linux Runtime, no host-Wine dependency and no silent Wine fallback.

This does not add games to Steam as non-Steam shortcuts. Steam is used only as a source of installed Proton compatibility tools.

### Related upstream work

- Addresses #639.
- Implements the Proton portion of #288.
- Related to draft PR #622; this PR takes a narrower per-game approach while preserving the existing Wine path.

## Checklist

- [x] CHANGELOG.md was updated.

### DIFF
--- a/minigalaxy/compatibility.py
+++ b/minigalaxy/compatibility.py
@@ -234,3 +234,148 @@ def find_steam_proton_tools(
             str(tool.path),
         ),
     )
+
+
+# ---------------------------------------------------------------------------
+# User-facing compatibility choices
+# ---------------------------------------------------------------------------
+
+SYSTEM_WINE_CHOICE_ID = "wine:system"
+CUSTOM_WINE_CHOICE_ID = "wine:custom"
+PROTON_CHOICE_PREFIX = "proton:"
+
+
+@dataclass(frozen=True)
+class CompatibilityChoice:
+    """
+    One compatibility option presented by the per-game Properties dialog.
+
+    ``kind`` is one of:
+        wine-system
+        wine-custom
+        proton
+
+    Proton choices preserve the full canonical compatibility-tool path so the
+    exact selected Steam/GE Proton installation can be persisted per game.
+    """
+
+    choice_id: str
+    kind: str
+    name: str
+    path: str = ""
+    source: str = ""
+    available: bool = True
+
+
+def _make_proton_choice(
+    path: Path,
+    name: str,
+    source: str,
+    available: bool = True,
+) -> CompatibilityChoice:
+    canonical_path = str(_canonical_path(Path(path)))
+
+    return CompatibilityChoice(
+        choice_id=f"{PROTON_CHOICE_PREFIX}{canonical_path}",
+        kind="proton",
+        name=name,
+        path=canonical_path,
+        source=source,
+        available=available,
+    )
+
+
+def build_compatibility_choices(
+    proton_tools: Iterable[ProtonTool] | None = None,
+    selected_runner: str = "",
+    selected_proton_path: str = "",
+) -> list[CompatibilityChoice]:
+    """
+    Build the choices displayed in the per-game compatibility selector.
+
+    System Wine and Custom Wine always remain available.
+
+    All discovered Proton compatibility tools follow those entries.
+
+    If a game already references a Proton installation which has since been
+    removed, retain a non-available entry. This prevents the GUI from silently
+    replacing the user's configured Proton version with Wine.
+    """
+    if proton_tools is None:
+        proton_tools = find_steam_proton_tools()
+
+    choices = [
+        CompatibilityChoice(
+            choice_id=SYSTEM_WINE_CHOICE_ID,
+            kind="wine-system",
+            name="System Wine",
+        ),
+        CompatibilityChoice(
+            choice_id=CUSTOM_WINE_CHOICE_ID,
+            kind="wine-custom",
+            name="Custom Wine",
+        ),
+    ]
+
+    seen_proton_paths = set()
+
+    for tool in proton_tools:
+        choice = _make_proton_choice(
+            path=tool.path,
+            name=tool.name,
+            source=tool.source,
+        )
+
+        if choice.path in seen_proton_paths:
+            continue
+
+        seen_proton_paths.add(choice.path)
+        choices.append(choice)
+
+    if selected_runner == "proton" and selected_proton_path:
+        selected_path = str(
+            _canonical_path(Path(selected_proton_path))
+        )
+
+        if selected_path not in seen_proton_paths:
+            selected_name = Path(selected_path).name or selected_path
+
+            choices.append(
+                _make_proton_choice(
+                    path=Path(selected_path),
+                    name=selected_name,
+                    source="missing",
+                    available=False,
+                )
+            )
+
+    return choices
+
+
+def selected_compatibility_choice_id(
+    selected_runner: str,
+    selected_proton_path: str,
+    custom_wine_path: str,
+    system_wine_path: str,
+) -> str:
+    """
+    Convert existing per-game metadata into a compatibility selector ID.
+
+    Legacy games which have no runner metadata continue to resolve to Wine.
+    An old custom Wine executable remains Custom Wine unless it is simply the
+    current system Wine executable.
+    """
+    if selected_runner == "proton" and selected_proton_path:
+        selected_path = str(
+            _canonical_path(Path(selected_proton_path))
+        )
+
+        return f"{PROTON_CHOICE_PREFIX}{selected_path}"
+
+    if (
+        custom_wine_path
+        and custom_wine_path != system_wine_path
+    ):
+        return CUSTOM_WINE_CHOICE_ID
+
+    return SYSTEM_WINE_CHOICE_ID
--- a/minigalaxy/compatibility.py
+++ b/minigalaxy/compatibility.py
@@ -1,0 +1,236 @@
+"""Discovery helpers for Windows compatibility tools installed by Steam."""
+
+from __future__ import annotations
+
+import re
+
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Iterable
+
+
+_LIBRARY_PATH_RE = re.compile(r'^\s*"path"\s*"((?:\\.|[^"])*)"\s*$')
+
+
+@dataclass(frozen=True)
+class ProtonTool:
+    """A Proton compatibility tool discovered in a Steam installation."""
+
+    name: str
+    path: Path
+    source: str
+
+
+def _canonical_path(path: Path) -> Path:
+    """
+    Return a normalized absolute path.
+
+    Steam commonly exposes the same installation through several symlinks,
+    for example ~/.local/share/Steam, ~/.steam/root and ~/.steam/steam.
+    Resolving those aliases prevents duplicate compatibility-tool entries.
+    """
+    path = path.expanduser()
+
+    try:
+        return path.resolve()
+    except (OSError, RuntimeError):
+        return path.absolute()
+
+
+def _unique_existing_directories(paths: Iterable[Path]) -> list[Path]:
+    """Return existing directories with equivalent paths deduplicated."""
+    unique_paths = []
+    seen_paths = set()
+
+    for path in paths:
+        if not path.is_dir():
+            continue
+
+        canonical_path = _canonical_path(path)
+
+        if canonical_path in seen_paths:
+            continue
+
+        seen_paths.add(canonical_path)
+        unique_paths.append(canonical_path)
+
+    return unique_paths
+
+
+def find_steam_roots(home: Path | None = None) -> list[Path]:
+    """
+    Find supported Steam installation roots.
+
+    Native Steam normally uses ~/.local/share/Steam and exposes compatibility
+    symlinks at ~/.steam/root and ~/.steam/steam. The Flatpak Steam location is
+    also considered so discovery can be reused by environments where that path
+    is accessible.
+    """
+    home = Path.home() if home is None else Path(home)
+
+    candidates = [
+        home / ".local" / "share" / "Steam",
+        home / ".steam" / "root",
+        home / ".steam" / "steam",
+        home / ".var" / "app" / "com.valvesoftware.Steam" / "data" / "Steam",
+    ]
+
+    return _unique_existing_directories(candidates)
+
+
+def _unescape_vdf_string(value: str) -> str:
+    """
+    Decode the small subset of VDF escaping relevant to filesystem paths.
+
+    We intentionally avoid adding a VDF dependency just to obtain Steam library
+    paths. Valve strings may escape a backslash or a double quote.
+    """
+    result = []
+    index = 0
+
+    while index < len(value):
+        character = value[index]
+
+        if character == "\\" and index + 1 < len(value):
+            next_character = value[index + 1]
+
+            if next_character in ("\\", '"'):
+                result.append(next_character)
+                index += 2
+                continue
+
+        result.append(character)
+        index += 1
+
+    return "".join(result)
+
+
+def parse_libraryfolders(vdf_path: Path) -> list[Path]:
+    """
+    Parse Steam library paths from steamapps/libraryfolders.vdf.
+
+    Only the ``path`` fields are required for compatibility-tool discovery, so
+    a complete KeyValues/VDF parser is unnecessary here.
+    """
+    vdf_path = Path(vdf_path)
+
+    if not vdf_path.is_file():
+        return []
+
+    libraries = []
+
+    try:
+        contents = vdf_path.read_text(encoding="utf-8", errors="replace")
+    except OSError:
+        return []
+
+    for line in contents.splitlines():
+        match = _LIBRARY_PATH_RE.match(line)
+
+        if not match:
+            continue
+
+        libraries.append(Path(_unescape_vdf_string(match.group(1))))
+
+    return libraries
+
+
+def find_steam_libraries(
+    steam_roots: Iterable[Path] | None = None,
+    home: Path | None = None,
+) -> list[Path]:
+    """
+    Return every accessible Steam library, including secondary libraries.
+
+    The Steam installation root itself is a library and additional locations
+    are obtained from steamapps/libraryfolders.vdf.
+    """
+    roots = list(find_steam_roots(home=home) if steam_roots is None else steam_roots)
+
+    candidates = []
+
+    for steam_root in roots:
+        steam_root = Path(steam_root)
+        candidates.append(steam_root)
+
+        library_file = steam_root / "steamapps" / "libraryfolders.vdf"
+        candidates.extend(parse_libraryfolders(library_file))
+
+    return _unique_existing_directories(candidates)
+
+
+def _directory_children(path: Path) -> list[Path]:
+    """Return directory children safely and in a deterministic order."""
+    if not path.is_dir():
+        return []
+
+    try:
+        return sorted(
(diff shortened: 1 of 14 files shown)


GB609:
Hi @adamjvr . 

I do admit that i've benched my work on UMU for quite a bit of time and i appreciate your willingness to contribute. But i also would have appreciated a bit of a message or question in the current draft PR if you were planning to work on the same. It's not exactly a pleasant feeling to have it "pulled" from me without as little as a bit of a heads-up.

Before i do a detailed review (which might take me one or two more days as i'm pretty busy atm), there are a few general things and questions i would like to discuss about your approach.

- You call it UMU support, but at the end, you are placing dependencies to steam and proton directly all over the code. UMU was made exactly to NOT have to manage stuff like that and not to have to depend on steam to use proton. And i personally think it's wrong to implement some vdf parsing in minigalaxy. Why do you need to parse steam library files in the first place? UMU can manage its runtimes on its own. If more configuration is needed, you could instead provide keyword-based options for `PROTONPATH` in the game properties, like `Ge-Proton` or `UMU-Proton`.

- Why the enforced minimal version 1.4.4 of UMU (which is the current latest release)? Which feature available since 1.4.4 are you relying on? When i tested it with older versions in the past it worked just as well. You're potentially locking out old distributions which might not be compatible because of other dependency reasons, the very thing you wanted to avoid by not depending on the distribution's package management.

- I think you are mixing UMU, Steam and proton in the user messages too much with a couple of very long and verbose messages. In the end, how exactly umu works doesn't and shouldn't have to be explained in every message. Minigalaxy could instead add a link to the umu documentation somewhere.

- I don't think we should manage the umu install. Especially not with hardcoded zip download paths. The probability of this getting outdated or breaking is too high. The distributions and their package managers are much better at doing that than we ever could be. If umu is missing a package for a certain distribution where you want to see it, it might be better to try and contribute there to get it into the builds and/or releases. The zipapp might not be compatible with all distributions in the first place.

- Your code is a bit odd in style at places. It looks as if you've set up your IDE with a line length of about 40-60 characters, resulting in a lot of unnecessary wraps. Moreover imports and ifs are using paranthesis in several places. Please have a look at what's there already and try to adjust. MG flake is set up with a max line length of 127. You don't have to fully use it everywhere, but you could use a bit more of that line length in several places.

- The 'per-game' approach will very quickly become very tedious for users preferring UMU in general. They would have to open game properties and select umu for every new game they want to install.

- The biggest charm of UMU is that it also includes a library of game fixes (similar to protonfixes), which would make most winetricks unnecessary. The trick is how to locate which ones to apply. UMU does a bit of that by itself, but ONLY when the name of the game passed to it 100% matches what it contains in its database. That is problematic in parts as the GOG/minigalaxy game names are not 100% the same as in the database (MG applies some name simplification logic to strip problematic characters).  
Aside from the automatic lookup, it can be specified manually by making proper use of the UMU environment variable `GAMEID`. You're always passing `GAMEID=0` which effectively disables the fixes. Minigalaxy does have the gog game ids, we can use those as lookup in `umu-database` (column `codename`) to find the right value for `GAMEID`.

minigalaxy/ui/properties.py
```diff
@@ -39,6 +60,25 @@ def __init__(self, parent_library, game, config: Config, api):
         self.game = game
         self.config = config
         self.api = api
+        self._compatibility_choices = {}
+
+        self.combo_properties_compatibility.connect(
```
GB609, line 65: I'm going to leave the question of whether all these buttons and the entire management of UMU is needed aside for now and will only comment on how the listeners are set up.

We use `@Gtk.Template.Callback` in combination with corresponding event names/configurations in the `ui` files. Please have a look at how it's done other places.
So there's no need to manually use `connect`.

minigalaxy/ui/properties.py
```diff
@@ -72,13 +115,351 @@ def __init__(self, parent_library, game, config: Config, api):
         # Center properties window
         self.set_position(Gtk.WindowPosition.CENTER_ALWAYS)
 
+    def _compatibility_choice_label(self, choice):
```
GB609, line 118: This is completely unnecessary, as well as most of the next function and most others related to it.
We have a common/generic utility function to populate combo boxes with a fixed data list following a certain pattern.

Please have a look at how it's done for several combos in `preferences.py`, like `__set_language_list`
- usage: [`preferences.py:__set_language_list`](https://github.com/sharkwouter/minigalaxy/blob/e60ee9ad9279391a6c48aacc7daee22a6400d922/minigalaxy/ui/preferences.py#L64)
- [Data definition](https://github.com/sharkwouter/minigalaxy/blob/e60ee9ad9279391a6c48aacc7daee22a6400d922/minigalaxy/constants.py#L3)

minigalaxy/ui/properties.py
```diff
@@ -30,6 +43,14 @@ class Properties(Gtk.Dialog):
     button_properties_cancel = Gtk.Template.Child()
     button_properties_ok = Gtk.Template.Child()
     label_wine_custom = Gtk.Template.Child()
+    label_properties_compatibility_tool = Gtk.Template.Child()
```
GB609, line 46: Aside from a few very rare exceptions, there is normally no need to pull in labels from the UI files.
Static labels can just be defined in the UI file and marked as translatable.

minigalaxy/umu.py
GB609, line 1: As I've said it in my first response to your PR, minigalaxy should NOT manage UMU by itself. No hardcoding of versions, checksums or download URLs. 

Let the user do a proper system installation so that it's available via PATH and, for unsupported distributions, add a preference that allows to supply a custom path to a local install (which the user has to do himself).

tests/test_launcher.py
GB609, line 1: You claim in the original, very verbose, but structured description of the PR (aside from a lot of pointless 'validations' which should be self-evident without having to explicitely mention them): 
> Full regression suite: 221 tests passed.

While, in fact, it's not true. And i haven't seen any more commits trying to fix it since you pushed the code. You also haven't enabled the test workflow in your fork and didn't try to run any test before filing your PR. It might have run locally for you, but you've got UMU installed. The Github runners (and other users likely too) don't.

This hints, among other things, strongly in the direction that you've just used AI and gave it minigalaxy, a few tickets and open PRs as input, then threw it at us without any preliminary verification or refinement.

I'm not going to look at your changes to the tests in detail. There's no point. There's so much that has to change in the code that the tests will require a rewrite afterwards.

minigalaxy/ui/data/properties.ui
```diff
@@ -286,44 +286,156 @@
                         <property name="top-attach">8</property>
                       </packing>
                     </child>
+                    <child>
```
GB609, line 289: Let's assume for minute that we really need all these new items.
They are all grouped and belong together and basically share the same condition for their visibility.

This is a situation where it would make sense to encapsule this in another, dedicated template ui file (coupled with a matching class) and instead define just a placeholder composite here as a second child of the `GtkBox` starting in line 39, which would then be initialised in the constructor of `properties.py` instead.

For the sake of this review, let me call it `UmuConfigPanel`.

When the labels and buttons are placed in `UmuConfigPanel`, it would be enough to just hide the panel itself from `properties.py` for linux games. No need to go into every single child element.

This pattern is used in several places. An easy example of that is in [`window.py` lines 55 - 58](https://github.com/sharkwouter/minigalaxy/blob/e60ee9ad9279391a6c48aacc7daee22a6400d922/minigalaxy/ui/window.py#L55)

minigalaxy/launcher.py
```diff
@@ -33,14 +131,48 @@ def wine_restore_game_link(game):
         os.symlink(relative, game_dir)
 
 
+def get_windows_helper_environment(game):
+    """Return the compatibility environment for prefix maintenance tools."""
+    environment = get_windows_environment(game)
+
+    if uses_proton(game):
```
GB609, line 138: `get_windows_environment` already patches the env with proton-specific variables. So why would you need another function calling `get_windows_environment` and do the very same type of patching a second time?

~~Just include the line `PROTON_VERB=runinprefix` in the first function.~~

Besides, `runinprefix` is the wrong verb. It should be `waitforexitandrun`.
`runinprefix` only makes sense for secondary starts when proton has pulled up the wineserver before.
Minigalaxy just launches the games and waits until they return. It leads to all kind of trouble when the game just forks out and its launching shell then dies immediately again. In that situation, minigalaxy will lose track of the game immediately and think it crashed. I'm speaking from experience here, this is one of the issues i encountered when i worked at that topic.

`runinprefix` would only make sense if it is done conditionally:
- IF the game is running at that moment
- with umu
- AND the user then opens wine config etc.
- (or the same in reverted order: opening config first, then launching game in parallel)

THEN `runinprefix` makes sense, otherwise proton will reject launching the second command.

But you're doing nothing of the sort here and just assume launching the game doesn't need it and the tools do, which, in fact, is not guaranteed to always be the case.

And all of this would add quite a lot of logic beyond initial umu support.

minigalaxy/launcher.py
```diff
@@ -22,6 +27,99 @@ def get_wine_path(game):
...
+    layout exactly. Proton adds the variables required by UMU after it.
+    """
+    prefix = os.path.join(game.install_dir, "prefix")
+    environment = [f"WINEPREFIX={prefix}"]
+
+    if uses_proton(game):
+        environment.extend([
+            f"PROTONPATH={game.get_info(InfoKey.PROTON_PATH)}",
+            "STORE=gog",
+            "GAMEID=0",
```
GB609, line 117: This line is the point i don't agree with and it's where you should have spent coding effort instead of how to manage umu installs:

Getting the right `GAMEID`. There's an online database from UMU with an API, but minigalaxy might also be in offline mode.
I'd try to find and include a valid value for `GAMEID` if:
- Minigalaxy is online and
- has not set cached ID before

save the lookup for the next time via `game.set_info` so that minigalaxy only has to try the lookup when there is no gameid AND its only. There's no need to waste time and api to do the lookup at EVERY start if there's a valid value as it normally doesn't change.

minigalaxy/compatibility.py
```diff
@@ -0,0 +1,381 @@
...
+
+    choice_id: str
+    kind: str
+    name: str
+    path: str = ""
+    source: str = ""
+    available: bool = True
+
+
+def _make_proton_choice(
```
GB609, line 270: This function doesn't really do much more than pass a few values to a class constructor while doing a few mutations.

You could move that code into the constructor itself and drop this function entirely, it doesn't really bring anything worthwhile to the table.

minigalaxy/installer.py
```diff
@@ -230,35 +237,80 @@ def extract_windows(game: Game, installer: str, language: str):
     return extract_by_wine(game, installer, game_lang), False
 
 
-def extract_by_wine(game, installer, game_lang, config=Config()):
-    # Set the prefix for Windows games
+def extract_by_wine(game, installer, game_lang, config=Config()):  # noqa: C901
+    # Validate the selected backend before touching the game or prefix.
+    compatibility_error = validate_windows_compatibility(game)
```
GB609, line 242: Your compatibility error check is unnecessary and it pollutes every function needing a windows runner.

But **all** of these functions effective rely on `get_windows_runner`. So, assuming a compatibility check really was necessary, it might be much better to put that check into `get_windows_runner` and have it `raise` instead of returning on error.

That way, all users of `get_windows_runner` will implicitely get to use it, without any need at all for duplicating that check everywhere.

Aside from that, i think checking for the umu version everywhere is prone to failures. What will we do when they decide to change their versioning strategy or semantics? Will it be possible as a developer to test unversioned snapshots? There's a lot to account for here, which adds unnecessary complexity.

Why not just let them try with other versions? If it runs, cool. If not, there will be an appropriate error message when trying to start either way.

minigalaxy/ui/properties.py
```diff
@@ -72,13 +115,351 @@ def __init__(self, parent_library, game, config: Config, api):
...
+
+        self.combo_properties_compatibility.remove_all()
+
+        for choice in choices:
+            self.combo_properties_compatibility.append(
+                choice.choice_id,
+                self._compatibility_choice_label(choice),
+            )
+
+        configured_choice_id = selected_compatibility_choice_id(
```
GB609, line 160: What happens here is a strong indicator that you've set up the arguments of the function wrongly. It feeds entirely from `Game`. So why is that not the only argument? Why are you passing all values individually?

This adds unnecessary boilerplate here and it makes the function called very hard to understand.

minigalaxy/compatibility.py
```diff
@@ -0,0 +1,381 @@
...
+            str(tool.path),
+        ),
+    )
+
+
+# ---------------------------------------------------------------------------
+# User-facing compatibility choices
+# ---------------------------------------------------------------------------
+
+SYSTEM_WINE_CHOICE_ID = "wine:system"
```
GB609, line 243: I generally can agree to giving a list of compatibility 'choices' to pick from.
But i don't think we should build it dynamically and add several proton versions.
It adds a lot of complexity to how the UI is set up.

That doesn't mean that there should be no way to pick the proton version.
However, it would be much easier if the proton version selection is a simple, dedicated combo or file picker which only gets enabled with `compatibility_choice==proton`.

That way you can drop most of your roundabout handling and parsing back-and-forth of CompatibilityChoice.

But picking a proton path or version must remain optional. Umu can handle this fine on its own.

minigalaxy/compatibility.py
```diff
@@ -0,0 +1,381 @@
...
+    A candidate must contain a ``proton`` launcher file. Canonical paths are
+    used to prevent Steam symlink aliases from producing duplicate entries.
+    """
+    roots = list(find_steam_roots(home=home) if steam_roots is None else steam_roots)
+    libraries = find_steam_libraries(steam_roots=roots)
+
+    tools = []
+    seen_paths = set()
+
+    def add_tool(path: Path, source: str) -> None:
```
GB609, line 196: There's very few places and reasons where it makes sense to define an inner function.
That's mostly useful when you need it to take some private scope values with it while it is being passed around as a callback function itself.

This is not the case here. Please, just move it to the top-level of the file. If you need encapsulation, wrap it (and all others) into a Class instead, where you can declare them as static if need be.

minigalaxy/compatibility.py
```diff
@@ -0,0 +1,381 @@
...
+    Discover Proton builds installed for Steam.
+
+    Valve Proton builds are normally installed below
+    ``steamapps/common/Proton*``. User-installed compatibility tools such as
+    GE-Proton are normally installed below ``compatibilitytools.d``.
+
+    A candidate must contain a ``proton`` launcher file. Canonical paths are
+    used to prevent Steam symlink aliases from producing duplicate entries.
+    """
+    roots = list(find_steam_roots(home=home) if steam_roots is None else steam_roots)
```
GB609, line 190: `find_steam_roots` already returns a list.
Whatever condition leads that function to return `None` might as well just return an empty list, so the entire `if ... else ...` fallback handling becomes unnecessary.

However, that part (and anything relying on it) is code that should be dropped completely.

minigalaxy/compatibility.py
GB609, line 1: One general note about locating proton versions for umu:
See [How do I use a Proton that is installed in a system path?](https://github.com/Open-Wine-Components/umu-launcher/wiki/Frequently-asked-questions-(FAQ)#how-do-i-use-a-proton-that-is-installed-in-a-system-path)

> umu-launcher only searches $HOME/.local/share/Steam/compatibilitytools.d for UMU-Proton or GE-Proton builds when configured to use the latest Proton build. See the [documentation](https://github.com/Open-Wine-Components/umu-launcher/blob/main/docs/umu.1.scd#examples) for an example on how to do this.

> Proton is compiled and is designed to execute in a runtime environment, and the reason umu-launcher does not search system paths is because compatibility tools there may not all be compiled against the [Steam Linux Runtime](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/main/docs/container-runtime.md?ref_type=heads). Since the goal of this project is for users to run their games through Proton just as it were launched from Steam, using Proton builds that were compiled outside the SLR would be against the project's goals.

What you are doing here smells like trouble all the way.
If we need to give a proton choice, make it a version name from a pre-defined list + latest and/or search the path umu looks at **only**.

Looking at the steam flatpak data path definitely **IS** the kind of possible compilation target issue UMU mentions in its documentation.

Aside from that, i don't think we should link Minigalaxy/UMU/proton to an existing steam in any way, not even indirectly. Nothing of this belongs here.

Consider the case that steam might change the `*.vdf` file's content or the entire format itself at some time in the future. Will you be there forevermore to continue to maintain that, so that minigalaxy won't start crashing all of a sudden?

Steam is a large company that does its own thing in an entirely different speed than we ever could keep up with. The `*.vdf` files are effectively internal data of steam. They are free to change them without notice to anyone whenever they want which could lead to minigalaxy breaking over night.

minigalaxy/launcher.py
```diff
@@ -22,6 +27,99 @@ def get_wine_path(game):
...
+    """Return a compatible UMU launcher, including MiniGalaxy's managed copy."""
+    installation = find_compatible_umu()
+
+    if installation:
+        return installation.path
+
+    return ""
+
+
+def validate_windows_compatibility(game):
```
GB609, line 45: I find this questionable. For which reason is there a need for a version check and for it to be at least 1.4.4?

GB609 (changes requested):
I've adressed the most important issues for now. There probably is some more, but doing all of that in one iteration is too much and too exhausting all at once.

It smells like AI without sufficient post-refinement. And it is incredibly taxing to deal with that in the form of review comments.
It's not in a mergeable state, neither from a feature point of view nor from a coding quality/style point of view.

Please, go over it with your own - certainly bright - natural intelligence and do a revision of my comments. Don't let the AI do this for you, i don't appreciate being put into an indirect feedback loop with your agent.

PR #770, opened 2026-08-18
fix: show all linux games in library

slowsage:
## Description
`__create_gametiles_iteratively` dispatches `self.games` in fixed-size index chunks via `GLib.idle_add`, computed on the worker thread without accounting for concurrent removals.
When `__create_gametiles` removes a windows-only game from an already-dispatched chunk, every later game shifts down one index into the range that chunk already consumed and won't revisit.
The next chunk starts counting from the new, shifted index, so that game is never picked up by either chunk — it silently gets no tile, which is why some linux games vanish from the library.

- Drop the mid-walk self.games.remove() that desynced chunk indices
- Split the update into a worker-thread fetch and a GLib.idle_add apply, so self.games is only ever written on the GTK thread
- Track a generation counter so a superseded update_library() call stops applying state once a newer call has started
- Apply installed tiles before the API fetch, then add remaining tiles in sorted idle batches
- Prune hidden-platform games from self.games after the merge instead of during the walk
- Add library tile creation tests

Note: Used ai to get this done.

## Checklist
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 **1.4.3**
+- Fix not-installed Linux games missing from the library (thanks to slowsage)
 - Releases will now include a basic AppImage (experimental) (thanks to GB609)
 - Releases now target Ubuntu 26.04
 - Mangohud dlsym enabled by default (thanks to RoGreat)
--- a/minigalaxy/ui/library.py
+++ b/minigalaxy/ui/library.py
@@ -4,7 +4,6 @@ import logging
 import os
 import re
 import threading
-import time
 
 from minigalaxy.api import Api
 from minigalaxy.config import Config
@@ -52,6 +51,7 @@ class Library(Gtk.Viewport):
         self.owned_products_ids = []
         self._queue = []
         self.category_filters = []
+        self._library_generation = 0
         self.configure_library_sort()
 
     def _debounce(self, thunk):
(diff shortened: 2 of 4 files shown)


GB609:
Hey. Thanks for finding that bug. Quite an oversight i had there. Guess stuff like that happens when one adds some last minute improvements.

But i'm wondering if the fix for this bug really has to be that complicated.
Wouldn't it be enough to just make a local copy of `self.games` in `__create_gametiles_iteratively` before looping and then add an if in `__add_gametile` to check if the game to add is contained in self.games (and add it if not)?

GB609:
Thinking about it a second time. Why would the bug happen the way you describe?

`__create_gametiles_iteratiely` runs on the ui thread. Which is a single thread. It pushes other executions BEHIND itself in the work schedule. self.games is only really manipulated with removes after it completes pushing chunks.

I tried to analyse that bug before as well and my traces rather lead me to issues installed linux games not having an id before api calls update stuff

slowsage:
Agree on the unnecessary complexity. Simplified it.

A snapshot in `__create_gametiles_iteratively` does not handle  `__add_games_from_api` in `minigalaxy/ui/library.py`, which walks `self.games` (`in` / `append` / `index`) on the worker while `GTK __create_gametiles` still does `self.games.remove()` on the same list.

The rest of the code handles stale refreshes - also now simplified.

GB609:
I took the master branch and added logs in several places to reproduce the bug and have a protocol of what actually happens.

Python isn't truly multithreaded. Even when `Glib.add_idle` is used, nothing happens until the currently running thread yields its execution by either sleeping or terminating. And my log clearly shows that.

You are right about the bug with the concurrent modification where `__create_gametiles` impacts the chunks produced by `__create_gametiles_iteratively`. The bug happens here because the worker thread and the UI thread are alternating turns due to `__create_gametiles_iteratively` using sleep (on purpose).
But once `__create_gametiles_iteratively` has finished and returned control to `__update_library`, there is no more ui thread running in parallel. `__add_games_from_api` has exclusive access to `self.games` while it is running. The interleaving and concurrency issues only start happening again on the second call to `__create_gametiles_iteratively`.

The flow in the logs roughly is:
| Worker | UI |
| --- | --- |
| update lib | |
| get installed | |
| create_iteratively | |
| `-> chunk self.games[...]` | |
| sleep 0.1s | |
| | `create tiles (possibly remove from self.games)` |
| -> next chunk | ... |
| get api games | |
| create_iteratively | |
| `-> chunk self.games[...]` | |
| sleep 0.1s | |
| | `create tiles (possibly remove from self.games)` |

The fix i did in the second branch is small:

```python
def __create_gametiles_iteratively:
    # ... some code
    # important: it MUST be a NEW list, not just another var pointing to self.games
    games_to_add = [*self.games]  
    # loop over games_to_add instead of self.games
```
That is enough. Whatever modification happens to `self.games` while `__create_gametiles_iteratively` and `__create_gametiles` are taking turns does NOT impact the calculation of the next chunk anymore and all games were displayed correctly.

Ok, granted, i did comment out one little code block at the beginning of `__create_gametiles_iteratively` which is meant to avoid offloading work for very small libraries. But for most users this only means that the list of installed games is displayed a bit quicker and at once, even when slightly larger than the update chunk size. The actual bug only happens for games pulled from the API, as only those can be filtered away.

See code at:
- Broken: https://github.com/GB609/minigalaxy/tree/test/library_bug_notfixed
- Fixed: https://github.com/GB609/minigalaxy/tree/test/library_bug_FIXED

Logs:
- broken: [mg_threading_bug.trace.log](https://github.com/user-attachments/files/31238095/mg_threading_bug.trace.log)
- fixed: [mg_threading_bug_FIXED.trace.log](https://github.com/user-attachments/files/31236945/mg_threading_bug_FIXED.trace.log)

GB609:
Please, don't get me wrong. 
I really appreciate you finding that bug. And i definitely like the tests you wrote.
But i still think the actual fix doesn't need much more than making sure that there's no alternating read/write of `self.games` simply by using another list instance with the same content in `__create_gametiles_iteratively`. And i think my test branch #2 proves that.

But if you want, i can also pull in your tests and have it run them too.

slowsage:
No issues at all - Defer to you to make these choices. Cheers!

PR #771, opened 2026-08-19
Bugfix: Concurreny issue leading to games from API being skipped in library rendering

GB609:
## Description

See the original explanation and PR at #770 

Thanks again to @slowsage for finding that issue.

I've cherry-picked the original commit and adjusted the fix according to my original suggestion before the PR was cancelled, while keeping as much as possible from the original test work.

@slowsage Is the way i've done it ok with you? I've made sure to keep your commit and credits intact as i don't want to steal your code. 

## Checklist
 

slowsage:
This works for me.

PS. I think you should take the credit to make the PR author mapping not confusing. All good.

Thank you!

sharkwouter:
Thanks guys!

PR #772, opened 2026-08-20
Feature: Better build reporting

GB609:
## Description

I recently discovered in the github logs that minigalaxy also records and prints coverage. As funny as it may read, i actually didn't know (or forgot) that we do this.
But it's hidden away in build step logs, which is a shame. We should leverage the reporting we have and make it a bit easier to see and access the most important points.

This is what i'm intending to do here. The following steps will now output log info to the job summary page:
- `flake8` for regular sources
- python tests print the final summary. Detail logs still need to be locked at by navigating into the logs of the step itself. The full details are just too large to show and i haven't found a good way to filter it down to failures only.
- coverage: Regardless of whether the tests fail, the recorded coverage table is added to the job summary. When the tests failed, the build will still fail afterwards the same way it did before

I know, it's not really a feature directed at end users. But i think it's useful for the dev CI/CD setup.

You can see the effects in the job history at my fork:
- PR branch = good case: https://github.com/GB609/minigalaxy/actions/runs/32380168099
- Test branch with flake error: https://github.com/GB609/minigalaxy/actions/runs/32379558357
- Test branch with failing test: https://github.com/GB609/minigalaxy/actions/runs/32379387572

## Checklist
 
### DIFF
--- a/.github/workflows/pythontests.yml
+++ b/.github/workflows/pythontests.yml
@@ -18,12 +18,31 @@ jobs:
         python -m pip install --upgrade pip
         pip install -r requirements-testing.txt
     - name: Lint with flake8
+      shell: bash
       run: |
-        flake8 . --append-config=.flake8.soft --count --show-source --statistics
+        echo -e '# Flake results\n```' >> "${GITHUB_STEP_SUMMARY}"
+        if flake8 . --append-config=.flake8.soft --count --show-source --statistics 2>&1 | tee -a "${GITHUB_STEP_SUMMARY}"; then
+          success=0
+        else
+          success=1
+        fi
+        echo -e '\n```\n' >> "${GITHUB_STEP_SUMMARY}"
+        exit "${success}"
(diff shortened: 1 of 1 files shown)


sharkwouter:
Look good, thanks!

I find it honestly a bit sad that GitHub does not allow tracking code coverage like Gitlab can. They do have support for it, but they only allow it for enterprise customers: https://docs.github.com/en/code-security/how-tos/maintain-quality-code/set-up-code-coverage

sharkwouter (approved):

GB609:
In regards to reporting, Jenkins (java world mostly) is incredibly mature and powerful. Graphical charts of all kinds of reportings directly on a jobs landing page, including ways to show build-to-build changes (history charts).

For github, it's mostly about finding workarounds to somehow do this by yourself. But still, it's better than nothing.

sharkwouter:
There are some applications which you can connect to github which give you a coverage report as a comment. I'm pretty sure codecov is free for open source projects.

PR #775, opened 2026-08-28
convert hard-coded platform name strings to enum

GB609:
## Description

Here's another idea: Let's try to move to using an enum for the magic strings `windows` and `linux` what appear everywhere.

It appears more verbose for now, but when we start adding type-hints to functions requiring a platform, it'll become easier to get the parameters right.

## Checklist
 
Nothing to add here for now.


sharkwouter (approved):

GB609:
I might have to do a few more tests about how the change behaves in preferences before i merge.

GB609:
OK, i made several switches back and forth in preferences and checked the config file. Even though the data list for the combo box uses the new enum for linux, the result looks as it as has, there's no issue.

sharkwouter:
Cool, I also didn't find any issues when I tested this :)

GB609:
I just wasnt too sure if python threated the enum as string correctly where it should. But it worked.

PR #778, opened 2026-09-06
Bugfixes: fixed some small bugs and errors during install

GB609:
## Description

1. Wine install is a bit more clever about the failure reason of an install: Manual user cancels change the error message and won't trigger a fallback retry in case the unattended installer was stopped midway by force-closing the window

2. Excluded the installer inventory json from the list of files for the checksum check. This is purely cosmetic and removes a warning from the log

3. Actual bug: `verify_installer_integrity` builds an array of error messages, except in one situation where a string is assigned. This lead to problems with the generic `\n`.join at the end of the function, which resulted in the entire message getting a linebreak after every character.

4. Developer convenience: Pretty-print installer inventory files

## Checklist
 
### DIFF
--- a/minigalaxy/installer.py
+++ b/minigalaxy/installer.py
@@ -152,14 +152,14 @@ def fail_on_error(message_to_test, fail_type=None, data=None):
 
 
 def verify_installer_integrity(game, installer_inventory, progress_callback=None):
-    error_message = []
+    error_messages = []
     invalid_files = {}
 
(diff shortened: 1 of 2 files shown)


GB609:
I just discovered i added a bit of unnecessary complexity: The `include_json` option in `InstallerInventory.as_keep_files_list` is not needed. This is basically the same as `contained_files()` which i might as well just use directly.

GB609:
fixed

sharkwouter (approved):

PR #780, opened 2026-09-06
Preserve library tiles during refresh

### LINKED ISSUE #740: `Library.reset` triggered by preference change destroys download progressbar update on game tiles
I stumbled over this issue while testing a set of changes related to `platform_mode`.
The problem is that `Library.reset` indiscriminately destroys all game tiles, which means that the DownloadProgress listeners attached to them won't be able to update anymore.

The _newly_ created game tiles are not attached to the download manager events anymore. Because there is no code to actually do that. It wasn't intended, the expectation has always been that the Game tile starts the downloads and listens to them, not that it gets re-created and has to re-attach to a running download.

No change in the Download UI. The progress just continues counting onwards there.

Solution: The best option would probably be to just not destroy all game tiles on reset.

be-student:
## Description

Preserve existing game tiles during ordinary library refreshes so active download listeners keep receiving progress updates. Stale tiles are removed after reconciliation, while changing grid/list view still performs an explicit rebuild.

Fixes #740

## Checklist

### DIFF
--- a/minigalaxy/ui/library.py
+++ b/minigalaxy/ui/library.py
@@ -82,28 +82,17 @@ class Library(Gtk.Viewport):
         GLib.idle_add(self.__load_tile_states)
         self.owned_products_ids = self.api.get_owned_products_ids()
         # Get already installed games first
-        previous_games = self.games
         self.games = self.__get_installed_games()
-        for game in previous_games:
-            if game.id in self.config.current_downloads and game not in self.games:
-                self.games.append(game)
         self.__create_gametiles_iteratively(5)
 
         # Get games from the API
         self.__add_games_from_api()
-        self.__create_gametiles_iteratively(5)
-        GLib.idle_add(self.__remove_stale_gametiles)
+        self.__create_gametiles_iteratively(5, prune_stale=True)
         GLib.idle_add(self.filter_library)
 
-    def __remove_stale_gametiles(self):
-        for child in self.flowbox.get_children():
-            tile = child.get_children()[0]
-            if tile.game not in self.games:
-                self.flowbox.remove(child)
-
-    def __create_gametiles_iteratively(self, step_width=5):
+    def __create_gametiles_iteratively(self, step_width=5, prune_stale=False):
         if len(self.games) < step_width*2:
-            GLib.idle_add(self.__create_gametiles, [*self.games])
+            GLib.idle_add(self.__create_gametiles, [*self.games], prune_stale)
             # wait until the update is done to make sure there's a consistent state before going on
             time.sleep(0.1)
             return
@@ -113,7 +102,7 @@ class Library(Gtk.Viewport):
         index = 0
         while index < len(games_to_add):
             games_chunk = games_to_add[index:index+step_width]
-            GLib.idle_add(self.__create_gametiles, games_chunk)
+            GLib.idle_add(self.__create_gametiles, games_chunk, prune_stale)
             index += step_width
             time.sleep(0.1)
 
@@ -158,7 +147,7 @@ class Library(Gtk.Viewport):
         tile2 = child2.get_children()[0].game
         return tile2 < tile1
 
-    def __create_gametiles(self, games_to_add=None) -> None:
+    def __create_gametiles(self, games_to_add=None, prune_stale=False) -> None:
         """Gets called twice: Once for installed, once for not installed games."""
 
         if not games_to_add:
@@ -168,6 +157,12 @@ class Library(Gtk.Viewport):
 
         for child in self.flowbox.get_children():
             tile = child.get_children()[0]
+            if prune_stale and tile.game.id in self.config.current_downloads and tile.game not in self.games:
+                self.games.append(tile.game)
+            if prune_stale and tile.game not in self.games:
+                self.flowbox.remove(child)
+                tile.game.library_tile = None
+                continue
             if tile.game in games_to_add:
                 logging.debug("Update existing tile for [%s] with new game instance", tile.game.name)
                 new_game = games_to_add[games_to_add.index(tile.game)]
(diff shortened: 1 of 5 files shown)


minigalaxy/ui/library.py
```diff
@@ -81,14 +82,25 @@ def __update_library(self):
...
+                self.games.append(game)
         self.__create_gametiles_iteratively(5)
 
         # Get games from the API
         self.__add_games_from_api()
         self.__create_gametiles_iteratively(5)
+        GLib.idle_add(self.__remove_stale_gametiles)
         GLib.idle_add(self.filter_library)
 
+    def __remove_stale_gametiles(self):
```
GB609, line 98: I wouldn't add a post-processing method like this which has to loop over everything again.
It would be better to change `__create_gametiles` a bit to drop stale tiles for the games passed to it in the last `elif game in self.games` block. 

This does require a bit a rework to the code dealing with `game.library_tile` because pre-existing tiles would then sometimes have to be dropped when none of the other conditions to display the tile holds true. Thus, game.library_tile' must not be checked first, but as the second check when a tile is determined to be visible.

Therefore, the check for `game.library_tile` should be moved to `__add_gametile` instead so that this method can deal with pre-existing tiles when the method is called from within `__create_gametiles`:

```python
   def __add_gametile(self, game):
        if game.library_tile:
            # the game already has a visible entry in the library
            # request to load the thumbnail, if there is a url for it and it hasnt been loaded before
            game.library_tile.load_thumbnail()
            return

        view = self.config.view
        if view == "grid":
            game_tile = GameTile(self, game)
        elif view == "list":
            game_tile = GameTileList(self, game)
        game.library_tile = game_tile
        # Start download if Minigalaxy was closed while downloading this game
        game_tile.resume_download_if_expected()
        self.flowbox.add(game_tile)
        '''
        using flowbox.show_all at this point would overrule any state-based
        hide() statements in game_tile (progress_bar in GameTileList)
        '''
        game_tile.show()
```

minigalaxy/ui/library.py
```diff
@@ -81,14 +82,25 @@ def __update_library(self):
         GLib.idle_add(self.__load_tile_states)
         self.owned_products_ids = self.api.get_owned_products_ids()
         # Get already installed games first
+        previous_games = self.games
         self.games = self.__get_installed_games()
+        for game in previous_games:
```
GB609, line 87: This section is not necessary. 
Games currently being downloaded can be taken care of in the second iteration of `__create_gametiles_iteratively`, as long as their tiles weren't dropped by the initial reset.

And the conditions for dropping would mean that the game instance fulfills none of the conditions from the first 3 checks in `__create_gametiles`

```python
if game.is_installed():
    self.__add_gametile(game)
elif game.id in self.config.current_downloads:
    self.__add_gametile(game)
elif game.platform in self.config.platform_mode:
    self.__add_gametile(game)
elif game in self.games:
    # housekeeping: API.get_library returns all owned games
    # (useful when api-caching is introduced as the same request can be used independent of platform_mode)
    # TODO: tiles should be dropped here
    self.games.remove(game)
```

Let's see what kind of tiles can have downloads:

1. NOT installed games. These will have their ID in `self.config.current_downloads` -> NOT dropped because it fulfills the second condition
2. Installed games for which DLCs are currently being downloaded, which means `game.is_installed() == True` -> NOT dropped because it fulfills the first condition

The one thing that has to change is how `game.library_tile` is checked currently. It has to move. But i'll come to that in another comment.

minigalaxy/ui/preferences.py
```diff
@@ -82,8 +82,8 @@ def __apply_locale_choice(self) -> None:
     def __apply_view_choice(self) -> None:
         view = get_combo_value(self.combobox_view)
         if view != self.config.view:
-            self.parent.reset_library()
-        self.config.view = view
+            self.config.view = view
+            self.parent.reset_library(rebuild=True)
```
GB609, line 86: This was not mentioned in the original ticket you linked in your PR, but changing the view still resets and drops all tiles, even those which are tracking active downloads. Changing the view requires dropping tiles, so the solution suggested in that ticket does not apply.

So the new tiles effectively have to 'reattach' to the currently running downloads.
The problem is that reconnecting download listeners is not really possible or very hard to implement into LibraryEntry because a lot of the related classes use instance- and context-bound callback functions through the entire feedback chain.

But there is an easy fix here: Before dropping all tiles, we can temporarily stop the downloads before resetting the library. 

When the tiles are re-created, they will then automatically restart/resume the downloads.

To achieve this, you have to use `self.download_manager.cancel_all_downloads(cancel_state=DownloadState.STOPPED)` **before** calling `reset_library`

It's a bit of a workaround. And there's a bit of an overhead involved with restarting downloads because some interactions with the api have to be repeated.
But given the rarity of occurence and the amount of work for the alternative (which would involve large rework of instance-bound download callback listeners like the `LibraryEntry.download_progress_manager` and others), this solution is still acceptable.

minigalaxy/ui/library.py
```diff
@@ -81,14 +82,25 @@ def __update_library(self):
...
+            if game.id in self.config.current_downloads and game not in self.games:
+                self.games.append(game)
         self.__create_gametiles_iteratively(5)
 
         # Get games from the API
         self.__add_games_from_api()
         self.__create_gametiles_iteratively(5)
+        GLib.idle_add(self.__remove_stale_gametiles)
```
GB609, line 95: It's better to integrate the tile dropping into `__add_gametiles`.
Maybe that would also a good opportunity to find a new, more suitable name for `__add_gametiles`.

GB609 (changes requested):
Thanks for your input. But there's few things that won't work out the way they are.

I hope you take the time to actually take care of the comments and tests yourself instead of leaving it up to codex.

PR #781, opened 2026-09-09
[Installer rework]: Step 2 - More and better install data classes

GB609:
## Description

This is part 2 of a series of changes to come to improve handling of game installations and make them more maintainable.

The key point is that, to support dynamic target platform selection, the way how installers are saved and downloaded must be harmonized. Moreover, Minigalaxy needs to save more info about the installers themselves to ensure that changing configuration will not corrupt or confuse them.

It might be possible with how the code looks right now, but would likely be much more complicated. My initial approach without the rework above involved a lot of file checking and directory renaming, which doesn't scale all too well and is needlessly complicated.

With a bit of inspiriration from `gogdownloader` i plan to rework the directory structure used to save offline installers to be more like 
- `installer/<game-slug>/<platform>/<installer-files>`
- `installer/<game-slug>/<platform>/<dlc-slug>/<dlc-installer-files>`

Slugs are given by gog in the api and provide reproducible, file-system compatible names while also being human-understandable (contrary to just using the game ids). I think they are better than doing our own string manipulations.

Potential future applications of this rework:
- a dedicated offline installers management feature
- integration with batch downloader tools like `goginstaller` just by following the directory scheme above

## Roadmap

- [x] 1. Changed `install_game` to use `Config` instead of several atomatic parameters
- [ ] **`CURRENT STEP`**: 2. Introduce and enhance data classes for installation.  
         `InstallerInventory` gets a new batch of meta information about the installers.  
         Inventories created with this commit can't be opened with previous versions (incompatible through new key `%META%`).
- [ ] 3. change `install_game` and `LibraryEntry` to use `InstallerInventory` and drop string parameter for installer executable
- [ ] 4. Code re-organization: Split up `installer.py` into a real directory sub-module. Simplifies testing and allow for extended use of type hints which aren't possible right now because of the code structure
- [ ] 5. Generalize `LibraryEntry` and `install_game` by using a new data type `InstallableItem` instead of `Game` and dlc titles/ids
- [ ] 6. Introduce `InstallerRepository` and offload logic from `LibraryEntry` related to locating and retrieving installers to that class
- (not _directly_ part of this series, but depending on step 6.) Finishing feature for platform selection
- (optional) For tests, i've converted a few methods in `LibraryEntry` not strictly related to GTK and UI into static methods to make them testable. The next logical action here would be to extract them into an (unsorted) utility function file.

## Checklist
 (marking this for now, but the current step is not yet worth an entry to changelog)

### DIFF
--- a/minigalaxy/ui/library_entry.py
+++ b/minigalaxy/ui/library_entry.py
@@ -118,7 +118,11 @@ class LibraryEntry:
         elif self.current_state in [State.INSTALLED, State.UPDATABLE]:
             err_msg = self.launch_game()
         elif self.current_state == State.INSTALLABLE:
-            install_thread = threading.Thread(target=self.__install_game, args=(self.get_keep_executable_path(),))
+            exe_path = self.get_keep_executable_path()
+            inventory = InstallerInventory.from_file_system(exe_path)
+            install_thread = threading.Thread(target=self.__install_game,
+                                              args=(self.get_keep_executable_path(),),
+                                              kwargs={"inventory": inventory})
             install_thread.start()
         elif self.current_state == State.DOWNLOADABLE:
             download_thread = threading.Thread(target=self.__download_game)
@@ -286,6 +290,9 @@ class LibraryEntry:
         download_files = []
 
         download_inventory = InstallerInventory()
+        # make sure to init the new meta data
+        download_inventory.item_id = gog_item.id
+        download_inventory.target_platform = download_info.get('os', self.game.platform)
         callback_factory = CallbackFuncWrapper(gog_item,
                                                finish_func,
                                                self.__cancel,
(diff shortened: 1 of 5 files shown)


minigalaxy/ui/library_entry.py
```diff
@@ -118,7 +118,11 @@ def run_primary_action(self, widget) -> None:
         elif self.current_state in [State.INSTALLED, State.UPDATABLE]:
             err_msg = self.launch_game()
         elif self.current_state == State.INSTALLABLE:
-            install_thread = threading.Thread(target=self.__install_game, args=(self.get_keep_executable_path(),))
+            exe_path = self.get_keep_executable_path()
+            inventory = InstallerInventory.from_file_system(exe_path)
+            install_thread = threading.Thread(target=self.__install_game,
+                                              args=(self.get_keep_executable_path(),),
```
sharkwouter, line 124: This should be exe_path I think

sharkwouter:
I've looked at the code and did some testing and it works well. I just had one comment.

GB609:
I'll take care of it in the next 1-2 days

sharkwouter (approved):

